PR #18, opened 2023-10-15
Initial model for ChiGame

borjasotomayor:
Here is the initial model for ChiGame. Models related to users, user profiles, and friending can be found in `[src/chigame/users/models.py](https://github.com/uchicago-cs/chigame/compare/initial-model?expand=1#diff-4f4ed8a381689a0e41cf8c30bf0ccff3763dd5ac92d439db7a9eb98c600f7c5f). Models related to games can be found in [src/chigame/games/models.py](https://github.com/uchicago-cs/chigame/compare/initial-model?expand=1#diff-cc8b2c0384e35671c45bc36bdc11c702f0f9d3f766f9f0aaa0fa1c353488b202).

Bear in mind that this is not meant to be a comprehensive model, but rather a model that can be used as a starting point to develop some of the features in the [Feature Wishlist](https://github.com/uchicago-cs/chigame/wiki/Feature-Wishlist). It is likely that students will expand on these models further.

### DIFF
--- a/src/chigame/games/models.py
+++ b/src/chigame/games/models.py
@@ -1,6 +1,6 @@
 from django.db import models
 
-from chigame.users.models import FriendGroup, User
+from chigame.users.models import Group, User
 
 
 class Game(models.Model):
@@ -56,7 +56,7 @@ class MatchProposal(models.Model):
 
     game = models.ForeignKey(Game, on_delete=models.CASCADE)
     proposer = models.ForeignKey(User, on_delete=models.CASCADE)
-    friend_group = models.ForeignKey(FriendGroup, on_delete=models.CASCADE)
+    group = models.ForeignKey(Group, on_delete=models.CASCADE)
     proposed_time = models.DateTimeField()
     min_players = models.PositiveIntegerField()
     joined = models.ManyToManyField(User, related_name="joined_matches", blank=True)
--- a/src/chigame/games/models.py
+++ b/src/chigame/games/models.py
@@ -1,0 +1,62 @@
+from django.db import models
+
+from chigame.users.models import FriendGroup, User
+
+
+class Game(models.Model):
+    """
+    A game (Chess, Checkers, etc.)
+    """
+
+    name = models.TextField()
+    description = models.TextField()
+    min_players = models.PositiveIntegerField()
+    max_players = models.PositiveIntegerField()
+
+
+class Lobby(models.Model):
+    """
+    A lobby that users can join before starting a match.
+    The match can start as soon as min_players join the lobby.
+    """
+
+    name = models.TextField()
+    game = models.ForeignKey(Game, on_delete=models.CASCADE)
+    created_by = models.ForeignKey(User, on_delete=models.CASCADE)
+    members = models.ManyToManyField(User, related_name="lobbies")
+    min_players = models.PositiveIntegerField()
+    max_players = models.PositiveIntegerField()
+
+
+class Match(models.Model):
+    """
+    A specific match of a game, between a set of players.
+    """
+
+    game = models.ForeignKey(Game, on_delete=models.CASCADE)
+    lobby = models.OneToOneField(Lobby, on_delete=models.CASCADE)
+    winners = models.ManyToManyField(User, related_name="won_matches", blank=True)
+    players = models.ManyToManyField(User, through="Player")
+
+
+class Player(models.Model):
+    """
+    A player in a match.
+    """
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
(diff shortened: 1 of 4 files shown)


src/chigame/games/models.py
```diff
@@ -0,0 +1,62 @@
...
+
+class Game(models.Model):
+    """
+    A game (Chess, Checkers, etc.)
+    """
+
+    name = models.TextField()
+    description = models.TextField()
+    min_players = models.PositiveIntegerField()
+    max_players = models.PositiveIntegerField()
```
majorsylvie, line 14: Do we want to provide options for a game type of any kind? (turn based, draft, real time, etc...)
Or leave that to the games team.

src/chigame/games/models.py
```diff
@@ -0,0 +1,62 @@
...
+    players = models.ManyToManyField(User, through="Player")
+
+
+class Player(models.Model):
+    """
+    A player in a match.
+    """
+
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    match = models.ForeignKey(Match, on_delete=models.CASCADE)
```
majorsylvie, line 48: Why is this included as a through table? Does the definition of the ManytoMany field not implicitly define this table?

Is it to make it easier for the Games/matches team to flesh out the model?

Or it is necessary now

src/chigame/users/models.py
```diff
@@ -33,3 +33,57 @@ def get_absolute_url(self) -> str:
...
+    An invitation from a User to another User, requesting that they become friends.
+    """
+
+    sender = models.ForeignKey(User, related_name="sent_friend_invitations", on_delete=models.CASCADE)
+    receiver = models.ForeignKey(User, related_name="received_friend_invitations", on_delete=models.CASCADE)
+    accepted = models.BooleanField(default=False)
+    timestamp = models.DateTimeField(auto_now_add=True)
+
+
+class FriendGroup(models.Model):
```
majorsylvie: Is a friend group necessarily a group of people that are all mutually Friends (as in have accepted FriendInvitation entries)?

Or is a FriendGroup any collection of users.

If it's the latter, then I feel the name could be misleading, might want to be changed to Group or PlayerGroup

src/chigame/users/models.py
```diff
@@ -14,10 +14,10 @@ class User(AbstractUser):
     """
 
     # First and last name do not cover name patterns around the globe
-    name = CharField(_("Name of User"), blank=True, max_length=255)
+    name = models.CharField(_("Name of User"), blank=True, max_length=255)
```
majorsylvie, line 17: What is the underscore doing here? (for both this and the email field)

Is it something about machine name vs human readable display name? 
If so might want to add a comment to this mentioning that

majorsylvie: actually just saw the `gettext_lazy as _`
read that its for simple translation: https://docs.djangoproject.com/en/4.2/topics/i18n/translation/

src/chigame/users/models.py
```diff
@@ -14,10 +14,10 @@ class User(AbstractUser):
...
     first_name = None  # type: ignore
     last_name = None  # type: ignore
```
majorsylvie, line 19: if we are using a generic `name` attribute to be more accomodating of different global naming practices, what point is there in still having the first and last name as attributes?

Also what does having a `field = None` actually do?

majorsylvie (approved):
There are a few points of clarification that I'd want both for my own sake and for ease of use.

Overall, looks good as a functional set of models for teams to use and build off of!

src/chigame/games/models.py
borjasotomayor: I don't think we want to get ahead of ourselves and start modeling that kind of stuff when we don't yet know what kind of games we're going to end up supporting. The games team could add fields to that effect, but I think it should be driven by the actual games we end up supporting.

src/chigame/games/models.py
borjasotomayor: It is very likely that Players will end up having attributes, so we may as well go ahead and define it explicitly as a through table.

src/chigame/users/models.py
borjasotomayor: Django's built-in User model already includes a first_name and last_name attribute. The above code was generated by cookiecutter, but I believe it ensures that those two fields are going to always be set to None.

src/chigame/users/models.py
borjasotomayor: We had clarified this in the Feature Wishlist, but hadn't changed the corresponding code. I've pushed an update that changes all references of "friend groups" to just "groups"

elizabethli31 (approved):
Looks good!

PR #69, opened 2023-10-28
(game management) Registered all models in games/models.py with admin page

### LINKED ISSUE #42: (game management) Register Game model with Admin page (and add more features?) (Patrick)
The models in games must be registered with the Admin page before the models can be utilized and populated

ptlin25:
Added code to games/admin.py to register all models in games/models.py with the admin page.

### DIFF
--- a/src/chigame/games/admin.py
+++ b/src/chigame/games/admin.py
@@ -1,0 +1,6 @@
+from django.apps import apps
+from django.contrib import admin
+
+app_models = apps.get_app_config("games").get_models()
+for model in app_models:
+    admin.site.register(model)
--- a/src/chigame/games/admin.py
+++ b/src/chigame/games/admin.py
@@ -1,6 +1,30 @@
-from django.apps import apps
 from django.contrib import admin
 
-app_models = apps.get_app_config("games").get_models()
-for model in app_models:
-    admin.site.register(model)
+from .models import Game, Lobby, Match, MatchProposal, Player
+
+
+@admin.register(Game)
+class GameAdmin(admin.ModelAdmin):
+    list_display = ("name", "min_players", "max_players")
+    search_fields = ("name__startswith",)
+
+
+# for future admin page customizations
+@admin.register(Lobby)
+class LobbyAdmin(admin.ModelAdmin):
+    pass
+
+
+@admin.register(Match)
+class MatchAdmin(admin.ModelAdmin):
+    pass
+
+
+@admin.register(Player)
+class PlayerAdmin(admin.ModelAdmin):
+    pass
+
+
+@admin.register(MatchProposal)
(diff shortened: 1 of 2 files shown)


rakrusch (approved):
Everything here looks good to me. Models are successfully registered in admin page. Good job Patrick!

src/chigame/games/admin.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import apps
+from django.contrib import admin
+
+app_models = apps.get_app_config("games").get_models()
+for model in app_models:
+    admin.site.register(model)
```
majorsylvie: I want you do register the models individually, since this way we will have more control over the customization features of each model.

https://docs.djangoproject.com/en/4.2/ref/contrib/admin/

https://realpython.com/customize-django-admin-python/



Then you can check out the above links to see how you can use the admin interface to customize your models

src/chigame/games/admin.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import apps
+from django.contrib import admin
+
```
majorsylvie, line 2: Image

As of now, your code doesn't have the names of models displayed in the admin interface. Please add an appropriate `__str__` method to the `Game` model such that the admin interface is more readable/useable.

majorsylvie (changes requested):
See comments. Your current code does work, there are a few things you can do to make the admin interface more usable and make your registration code more robust for the future.

src/chigame/games/admin.py
ptlin25: Thank you for the feedback! I added a `__str__` method to the `Game` model and customized the game model in the admin interface in the latest commit. Please let me know what you think

ptlin25: In the admin interface, each game is listed with its name as well as min/max number of players. I also added a search bar, so games are searchable by their name

majorsylvie: FYI this search features is somewhat limited, only matching the very beginning of titles and seemingly not working for spaces:

Not that this is a huge problem that needs to be adjusted, but FYI. this could be a task to make the admin search bar better.

image
image
image
image
image

majorsylvie (approved):
Admin interface is much cleaner and you've refactored. 
You could consider making the admin search more robust, but you do not need to as of now. 
I'll put a task on the spring backlog for it.

ptlin25:
Admin interface now shows the name, minimum number of players, and maximum number of players for each game when viewing the list. In the future, we can think about improving the search feature

majorsylvie:
Issue Score: Excellent

Comments:
Great job! Thank you for taking the feedback given and providing a simple and clear closing comment.

Keep moving forward :)

PR #73, opened 2023-10-29
added general url, view, form, and html for match-init

### LINKED ISSUE #27: (matches) create view for initializing a match (Chloe and Nat)
**Goal:** We currently lack a way for users to start matches once inside a lobby. We want a "Create New Match" view that allows users to initialize a match in a lobby with a minimum number of players or more. This view should have a field for the host user to choose which players in the lobby are going to participate in the match. This means the "create" view should automatically fill in the game attribute of the match (this will already be decided by the game associated with the lobby).

**Motivation:** Users should be able to decide when to start a match once inside a lobby.

**Open questions:**
1. Should only specific types of lobbies allow hosts to choose specific players for a match (while general lobbies include all users in the lobby as players by default)?
2. Should users be allowed to choose unregistered users as players?

jiyeongchloe:
We've implemented code for a general (not yet advanced) Create View for match initialization. (Issue #27) 
Files Updated:
- `games/urls.py`
- `games/views.py`

Files Added:
- `games/forms.py`
- `templates/games/match_form.html`

### DIFF
--- a/src/chigame/games/forms.py
+++ b/src/chigame/games/forms.py
@@ -1,0 +1,8 @@
+from django import forms
+from .models import *
+
+class MatchInitForm(forms.ModelForm):
+    class Meta:
+        model = Match
+        fields = ['game', 'lobby', 'winners', 'players']
+
--- a/src/chigame/games/urls.py
+++ b/src/chigame/games/urls.py
@@ -4,4 +4,5 @@ from . import views
 
 urlpatterns = [
     path("", views.GameListView.as_view(), name="games-list"),
+    path("lobby/<int:pk>/match/create/", views.match_init, name="match-init"),
 ]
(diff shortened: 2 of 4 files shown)


nugerhard (approved):
I think this code is good, and I think this accomplishes what was described in the issue. However, I think this will not eventually align with our design of matches: in particular, as discussed in #27, users will probably not ever create matches, and as such this form is unnecessary. In any case, if we end up needing a match form, this is a good place to start from.

@elizabethli31 any thoughts about using class-based vs function-based views? My only thought here is that we might use a class-based view on the Match model instead. But if there isn't a standard/expectation, I'm ok with this.

jiyeongchloe:
We decided we no longer need a view of match instantiation for the users to interact with. We've closed the issue, and the PR is no longer needed.

PR #78, opened 2023-10-30
(user profiles) Create user profile page that displays relevant information

### LINKED ISSUE #28: (user profiles) Create user profile page that displays relevant information
Create a user profile pages that displays relevant personal information: display name, picture(if added). Profile page should only be accessible by the authenticated, active user and should be able to direct to a page to edit personal info if a user is viewing their own profile page. It should also include an Add Friend button if a user is viewing another user's profile.

omarelkashef:
Create a user profile pages that displays relevant personal information: display name, picture(if added). Profile page should only be accessible by the authenticated, active user and should be able to direct to a page to edit personal info if a user is viewing their own profile page. It should also include an Add Friend button if a user is viewing another user's profile.

### DIFF
--- a/src/templates/users/userprofile_detail.html
+++ b/src/templates/users/userprofile_detail.html
@@ -17,7 +17,7 @@
               <div class="media-body">
                 <p>Bio: {{ object.bio }}</p>
                 {% if object.user == request.user %}
-                  <a href="{% url users:edit-profile %}"" class="btn btn-primary">Edit Profile</a>
+                  <a href="{% url 'users:update' %}"" class="btn btn-primary">Edit Profile</a>
                 {% endif %}
               </div>
             </div>
--- a/src/templates/users/userprofile_detail.html
+++ b/src/templates/users/userprofile_detail.html
@@ -1,0 +1,29 @@
+{% extends "base.html" %}
+
+{% load static %}
+
+{% block title %}
+  User:
+  {{ object.dispaly_name }}
+{% endblock title %}
+{% block content %}
+  <div class="container mt-5">
+    <div class="row">
+      <div class="col-md-6 offset-md-3">
+        <div class="card">
+          <div class="card-header">{{ object.display_name }}</div>
+          <div class="card-body">
+            <div class="media">
+              <div class="media-body">
+                <p>Bio: {{ object.bio }}</p>
+                {% if object.user == request.user %}
+                  <a href="{% url users:edit-profile %}"" class="btn btn-primary">Edit Profile</a>
+                {% endif %}
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+{% endblock content %}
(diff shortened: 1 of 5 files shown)


src/templates/users/userprofile_detail.html
```diff
@@ -0,0 +1,31 @@
+{% extends "base.html" %}
+
+{% load static %}
+
+{% block title %}
+  User:
+  {{ object.dispaly_name }}
```
AndrewM131: I'm assuming you meant "object.display_name", not "object.dispaly_name"

src/templates/users/userprofile_detail.html
```diff
@@ -0,0 +1,31 @@
...
+    <div class="row">
+      <div class="col-md-6 offset-md-3">
+        <div class="card">
+          <div class="card-header">{{ object.display_name }}</div>
+          <div class="card-body">
+            <div class="media">
+              <div class="media-body">
+                <p>Bio: {{ object.bio }}</p>
+                {% if object.user == request.user %}
+                  <a href="{% url 'users:update' %}" class="btn btn-primary">Edit Profile</a>
```
AndrewM131, line 20: Just a quick question. How is this view different from templates/users/user_detail.html file as it also has a similar format to this HTML file?

AndrewM131 (changes requested):

src/templates/users/userprofile_detail.html
AndrewM131: Specifically lines 16-28

omarelkashef:
@24raniwalar and I realized that a function-based implementation of this view is better since we are dealing with two models(UserProfile and User) at the same time.

src/templates/users/userprofile_detail.html
omarelkashef: Good catch! Fixed the typo

src/templates/users/userprofile_detail.html
omarelkashef: This is a different view to display a profile page not the user itself. Later, it will include friends, achievement, games, etc. user_detail.html is accessible through the User model unlike userprofile_detail.html which is accessible through UserProfile mode

AndrewM131: Oh I see, thank you for clarifying, that makes more sense, and it's a great base to build off of.

AndrewM131 (approved):
Hey, thanks for addressing my comments and questions. Looked it over and it looks good to me now!

elizabethli31 (approved):
Looks good and great job!

omarelkashef:
Closing statement: We created a profile pages that has display name and bio which is only accessible by the authenticated. My Profile now directs to this view instead of user detail page. 

@omarelkashef worked on displaying the information on the profile page
@24raniwalar worked on adding the logic for the Add Friend button and redirection

elizabethli31:
Issue Score: Excellent

Comments:
Great job! Simple PR that works well and good error handling.

PR #103, opened 2023-10-31
Added Lobby List and Lobby detail pages

### LINKED ISSUE #65: (matches) create view for allowing users to join a lobby (Lucien and Moses)
Goal: implement a lobby list page that display all join-able lobbies to any user. When clicking on a lobby title, user should be directed to the detail page of that lobby.

### LINKED ISSUE #66: (matches) create a view for lobby details
Goal: Create a view page for users to be able to view lobby details based on the view to join a lobby.

xylucien:
We have implemented basic html pages for the lobby list and detail pages for each lobby. They are under the URL "lobby/". For the lobby list, since I want to display it as a table element, I did not use the default ListView from django and instead wrote a customized view method for it.

### DIFF
--- a/src/chigame/games/views.py
+++ b/src/chigame/games/views.py
@@ -1,9 +1,14 @@
 from django.views.generic import ListView
 
-from .models import Game
+from .models import Game, Lobby
 
 
 class GameListView(ListView):
     model = Game
     queryset = Game.objects.all()
     template_name = "games/game_list.html"
+
+class ViewLobbyDetails(ListView):
+    model = Lobby
+    queryset = Lobby.objects.all()
+    template_name = "games/lobby_details.html"
--- a/src/chigame/games/views.py
+++ b/src/chigame/games/views.py
@@ -1,9 +1,15 @@
 from django.views.generic import ListView
+from django.shortcuts import render, redirect, reverse
 
-from .models import Game
+from .models import *
 
 
 class GameListView(ListView):
     model = Game
     queryset = Game.objects.all()
     template_name = "games/game_list.html"
+
+def lobby_list(request):
+    lobbies = Lobby.objects.all()
+    context = {"object_list": lobbies}
+    return render(request, "games/lobby_list.html", context)
--- a/src/chigame/games/views.py
+++ b/src/chigame/games/views.py
@@ -13,10 +13,9 @@ def lobby_list(request):
     context = {"object_list": lobbies}
     return render(request, "games/lobby_list.html", context)
 
-class ViewLobbyDetails(ListView):
-    model = Lobby
-    queryset = Lobby.objects.all()
-    template_name = "games/lobby_details.html"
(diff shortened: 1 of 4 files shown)


src/chigame/games/views.py
```diff
@@ -1,15 +1,24 @@
...
 
 class GameListView(ListView):
     model = Game
     queryset = Game.objects.all()
     template_name = "games/game_list.html"
 
+def lobby_list(request):
+    lobbies = Lobby.objects.all()
+    context = {"object_list": lobbies}
+    return render(request, "games/lobby_list.html", context)
```
nugerhard, line 15: Maybe worth writing as a class for consistency, maybe not. I'm good with this though.

src/chigame/games/views.py
```diff
@@ -1,15 +1,24 @@
...
+
+class ViewLobbyDetails(ListView):
```
nugerhard: I think this should inherit from DetailView vs. ListView. That way, if we go to a lobby/pk/ where a lobby with this pk doesn't exist, we get a 404.

nugerhard (changes requested):
Code looks good, displays fine when I run it locally. There's only one change I suggest, and that's to have ViewLobbyDetails inherit from DetailView vs. ListView. See my note in the code. After that, gtg.

src/chigame/games/views.py
elizabethli31: Either way (function or class) works.

elizabethli31 (changes requested):
List view looks fine, but I'm not seeing anything for lobby_details. I believe you need to use DetailView and add a context_object_name.

agustinmoses:
Updated files so that they reflect suggested changes. Namely, changed the class ViewLobbyDetails to a function instead in views.py, and updated the URL in urls.py so that they reflect the changes in views.py

nugerhard (changes requested):
This code isn't functional. Please test before doing a PR. See guidelines on how to fix in slack channel.

nugerhard (approved):
Approved in error

nugerhard (changes requested):
You set `context_object_name = "lobby_detail"` in the LobbyDetail class but then refer to `lobby` in the actual html template.

nugerhard (approved):
Works for me

elizabethli31 (approved):
LGTM! Good job guys. Can you update this branch with dev and commit? I'll merge once you've done that

xylucien:
Lobby list view is now available. For each lobby, a detailed list is also implemented. In the future we might be more careful before each commit and should test around with a fixture.

elizabethli31:
Issue Score: Excellent

Comments:
Good work on a simple detail page! For future PRs, make sure you code is 100% tested and ready to go instead of iterating multiple rounds of reviews. Also, make sure you detail exactly who did what in your closing comment.

PR #110, opened 2023-11-01
Created users-friends fixture file

### LINKED ISSUE #31: (user profiles) create sample user profile page

henrietta-k:
Addressing issue #31 

Created a fixture file with sample Users and userProfiles.


24raniwalar:
I think it may be a good idea to add a userprofile for the admin user as well, but other than that LGTM!

24raniwalar (changes requested):
I think it may be a good idea to add a userprofile for the admin user as well, but other than that LGTM!

ishiii-s:
> I think it may be a good idea to add a userprofile for the admin user as well, but other than that LGTM!

Thanks for the review Rishabh! I believe we already have an admin user in the file, which is user 1 (is_staff is set to true). Let us know if that checks out or if we still be to make changes for an admin.

elizabethli31 (changes requested):
Agreed with Rishabh. I believe what he's saying is to add a userprofile for the admin user. Also, please make a new directory for user fixtures "src/chigame/users/fixtures" and move this file there.

henrietta-k:
Turns out the userProfile instance for the admin was there but it had a primary key of 5 for some reason. I've updated all the primary keys for the userProfile to match their corresponding users. I also moved the file into a new folder.

henrietta-k:
Turns out the userProfile instance for the admin was there but it had a primary key of 5 for some reason. I've updated all the primary keys for the userProfile to match their corresponding users. I also moved the file into a new folder.

24raniwalar (approved):
LGTM! Nice revision

elizabethli31 (changes requested):
Did you guys create the users with the create_user method? I tested the fixture data, and I'm not able to sign in with the users and their passwords. The user profiles are showing up correctly though.

elizabethli31 (approved):
LGTM! Pull from the most updated dev and I'll merge. Great work!

ishiii-s:
Closing Statment: 
We were able to successfully complete this PR. In this PR, @henrietta-k and @ishiii-s both worked together to figure out the general approach (e.g. how many users, what type of users), and both added code for sample users to the file. @henrietta-k then addressed the requested changes.

elizabethli31:
Issue Score: Excellent

Comments:
Like mentioned in the issue, this PR was super helpful for testing. Nice closing comment and thanks for addressing my changes. The description summary is a bit too terse, and I have taken a point off for that.

PR #117, opened 2023-11-02
Made the Game edit view and Game create view only accessible to staff members

### LINKED ISSUE #111: Make non GET requests to Game CRUD pages available to staff users only
We need to restrict non-GET requests to only be allowed from staff users. This means that the edit and create views should not be accessible to non-staff users. The links to the edit and create views should only be shown if the user is a staff member, and if a non-staff user goes to those URLs, it should return a HTTP 403 or 404 error

ptlin25:
Addressing issue #111. The edit link on the game detail page and the create link on the game list page are now only displayed if the user is a staff member. Additionally, if a non-staff user tries to access those URLs directly, a HTTP 403 Forbidden error is raised, telling them that they do not have permission for the page.


rakrusch (approved):
Good work @ptlin25  ! Edit and create view links appear to be successfully hidden from non-staff members.

src/chigame/games/views.py
```diff
@@ -16,17 +17,25 @@ class GameDetailView(DetailView):
...
+    def test_func(self):
+        return self.request.user.is_staff
 
-class GameEditView(UpdateView):
+
+class GameEditView(UserPassesTestMixin, UpdateView):
     model = Game
     fields = ["name", "description", "min_players", "max_players"]
     template_name = "games/game_form.html"
+    raise_exception = True
```
majorsylvie: Would be good to add a comment and/or reference to documentation which states what the purpose of this line is

src/chigame/games/views.py
```diff
@@ -16,17 +17,25 @@ class GameDetailView(DetailView):
...
 
     def get_success_url(self):
         return reverse_lazy("game-detail", kwargs={"pk": self.kwargs["pk"]})
+
+    def test_func(self):
```
majorsylvie, line 56: Similarly would be good to add a comment and/or documentation about what this function is doing. Since it doesn't appear directly in this file anywhere.

majorsylvie (changes requested):
Tested and this code works exactly as expected. 
Would be good to add some explanatory comments as mentioned in review. 

After adding those comments this code is good to go!

ptlin25:
Added comments and resolved merge conflicts as requested by the review. Ready for re-review

majorsylvie (approved):
Thank you for the comments and resolving the merge conflict! 
With CI passing this should be good to go.

ptlin25:
The Game edit view and Game create view are now only accessible to staff members. Comments have been added to clarify that an exception will be raised if a non-staff member tries to access the edit and create views.

majorsylvie:
Issue Score: Excellent

Comments:
Great work! 
I will always appreciate well documented code.

PR #120, opened 2023-11-02
New page showing list of all users on the platform 

### LINKED ISSUE #112: List view for all users

henrietta-k:
Addresses Issue #112 

Added a new template page that shows a list of all users on the platform. It only displays a user's username. Login is required for this page so it is accessible to regular users and admin once logged in. The goal of this page is so that in the future, this page can hold tournament information related to a user, such as which user is ranked the highest on the platform, etc. A new template page was added, a new view function was added, and a new URL path was added, and a new tables.py file was created with a UserTable class.


24raniwalar (commented):
Had a question: Is there a reason we are allowing all users to see the user table, as opposed to just staff or just admin users?

henrietta-k:
> Had a question: Is there a reason we are allowing all users to see the user table, as opposed to just staff or just admin users?

Comments have been added to the files for future fixes (eg. adding a feature to see the top ranking users, users who have played the most matches, etc).

24raniwalar (approved):
Sounds good! Let's make sure we revisit this with the other teams in Sprint 3.

requirements/base.txt
```diff
@@ -7,3 +7,4 @@ django-environ==0.10.0  # https://github.com/joke2k/django-environ
 django-allauth==0.57.0  # https://github.com/pennersr/django-allauth
 django-crispy-forms==2.0  # https://github.com/django-crispy-forms/django-crispy-forms
 crispy-bootstrap5==0.7  # https://github.com/django-crispy-forms/crispy-bootstrap5
+django-tables2==2.5.0  # https://github.com/jieter/django-tables2
```
elizabethli31: This needs to be in local.txt. But I just merged a PR that covers this, so you can just remove it.

src/chigame/users/tables.py
```diff
@@ -0,0 +1,12 @@
+import django_tables2 as tables
+from .models import User
+
+class UserTable(tables.Table):
+    name = tables.Column(verbose_name="Name")
```
elizabethli31, line 7: Is it intentional for the table to only have name? I see that this is intended to display users for everyone and not just admin, so it makes sense security wise. However, when signing up, you only need to use an email and password so it could just be empty for all users.

elizabethli31 (changes requested):
Generally looks good! Left a few comments. You can just pull from dev and use the newly added django-tables2 requirement added by another team. Also make sure CI tests are being passed

src/chigame/users/tables.py
henrietta-k: image
I think the name field is not optional but if it is, would having an email field help?

elizabethli31: It is optional. The name attribute lets blank=True. Email would work to make something show up. However, for the future functionalities that you guys want, this might not make sense as it will display emails to all users. What do you think? Maybe have a separate display for admin only with email and name for all users?

elizabethli31: Sorry missed the user-detail page, so I see you have one with the email!

henrietta-k:
Update: we met two times during the sprint to work on this issue. The first was on November 8th when we worked together to figure out the specific goals that this issue would address (setting up a page so that in the future, tournament information for all users on the platform can be displayed here) and how we wanted to go about accomplishing this goal. We met again on November 12th to finalize some issues (linting issues) and fine-tune what our goals were related to this task.

henrietta-k:
Update: we're reevaluating how to implement the UserTable after the two reviews.

src/chigame/users/tables.py
henrietta-k: I've added the `first_name` and `last_name` attributes from the User model, but since `name` and these two attributes are all optional, if a User doesn't have any of these attributes, I'm not sure what we can do about it.

henrietta-k:
Update: conflicts with URL paths between this branch and the dev branch have been fixed.

src/chigame/users/tables.py
elizabethli31: Okay, this is fine with me!

elizabethli31 (changes requested):
Still getting DS_store on these - can you remove them?

elizabethli31 (approved):
Thanks for addressing my comments. Good work!

AndrewM131:
Closing statement: We successfully implemented a view of all users on the platform. Any user can access this view. This 
Files changed: src/chigames/users/tables.py, src/chigames/users/urls.py, src/chigames/users/views.py, src/templates/users/user_list.html
This is something we should return to after tournaments and matches team deploy features such that we can create a leaderboard with the user list and ranking them by points, and such.

elizabethli31:
Issue Score: Excellent

Comments:
Nice job providing additional functionality for admins. Also, I like that you guys provided updates immediately once a fix was made, as that really helped with testing and knowing what was updated.

PR #108, opened 2023-11-01
(user profiles) Add friend invitations endpoints

omarelkashef:
Create endpoints to allow users to send and cancel friend requests. Profile pages need to reflect the status of a friend request. That is, it should display an Add Friend button if users are not friends and there is no friend invitation already and Remove Friend button if they are already friends and Cancel Friend Request if users are not friends but there is friend invitation already.

PR #130, opened 2023-11-03
Create a notification model and update friending endpoints to make use of the model 

### LINKED ISSUE #127: Design Notification model and edit friending endpoints to make use of it
We need a Notification model that will have updates for the user such as friend requests. A user need to be able to mark notification as read. The model should be general enough to handle all types of notifications. Notification can be marked notification as read, unread or deleted. The model will have another model instance as an "actor" who creates the notification.

omarelkashef:
This PR implements a Notification model that will have updates for the user such as friend requests and allows for models and parts of the app to register notifications. The model allows for the user to mark notification as read, unread or deleted. The model includes an actor which is the model that created the notification and a receiver which is a User as well as an optional message body. We also created a custom queryset class that expands models.QuerySet to remove redundancy in querying Notification. We also modified the friending viewpoints in PR #108 to create notifications when sending a friend invitation and make those notifications deleted(invisible) when an invitation is cancelled. Each instance of FriendInvitation has its own notification object even if the invitation is between the same sender and receiver.


henrietta-k (commented):
Looks great to me! Love how detailed the model is with the `first_sent` and `last_sent` attributes. Before I approve it, perhaps having a type attribute might come in handy in the future. Eg. FRIEND_REQUEST, REMINDER, UPCOMING_MATCH, etc. Up to what you intend on using this Model for.

omarelkashef:
> Looks great to me! Love how detailed the model is with the `first_sent` and `last_sent` attributes. Before I approve it, perhaps having a type attribute might come in handy in the future. Eg. FRIEND_REQUEST, REMINDER, UPCOMING_MATCH, etc. Up to what you intend on using this Model for.

Yes having a type would be helpful. My initial thought was that type would be inferred from actor_content_type which stores the model that issues the notification so that I don't restrict the types of models. But I will go ahead and add a type that we can expand on whenever we need to.

henrietta-k (approved):
Looks good to me!

elizabethli31 (approved):
Really great job! Tested and it ran smoothly. Everything from the renewed notifications to the added notification types shows all the details put into this PR. The model and views are clearly very well-thought-out. Again, excellent work! My only comment is that you guys should document the Notification model on the Wiki. This can be a quick issue for this sprint or the next.

omarelkashef:
Closing statment: we created a Notification model that stores updates such as friend request for the users. The user is able to mark a notification as read, unread or deleted.  We also created a custom queryset class to remove redundancy in querying Notification. The friending viewpoints have been modified to make use of the new Notification model.

@omarelkashef  worked on implementing the Notification model and its queryset
@24raniwalar  worked on using the new model in the friending endpoints.

elizabethli31:
Issue Score: Excellent

Comments:
Really solid PR and review! Very detailed error handling as always. I also liked that the notification model considered Chigame as a whole and handled different match related notifications despite you guys focusing primarily on internal user/friending features.

PR #136, opened 2023-11-05
Create tournaments/models.py based on the tournaments/ERD.md

### LINKED ISSUE #85: Design models for single-elimination tournament
## Description

We are going to try convert the ERD table into the Django models to utilize the ORM and map our entities in the database; we are going to implement the ERD in the models.py file for the tournaments App.

## Goal

Have a viable and reliable models.py file that successfully maps the ERD and serve the relationships between the entities correctly

MikeChen012345:
A models.py file has been created based on tournaments/ERD.md in the wiki. In the file, I have implemented the Tournament, Notification, Chat, and Message models based on the tournaments/ERD.md in the wiki.


src/chigame/tournaments/models.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class Chat(models.Model):
+    """
+    Represents the live chat feature for both players and spectators.
+    It contains a list of messages, which are not mixed with the messages from
+    other Chats.
+    """
+
+    match = models.ForeignKey(Match, on_delete=models.CASCADE)
```
rqueiroz20: The Matches team has not responded to my messages regarding how they're setting up their matches in lobbies. Until that is figured out, we might have to change how Chat is initialized in our model. I took a look at their model, and we should probably decide whether we want to initialize lobbies specifically for tournament matches or some other method for distinguishing them from non-tournament lobbies.

rqueiroz20 (commented):
This `models.py` file fits the ERD seen in the wiki. It looks good for now, but I would suggest holding off with the merging until the Matches team responds to clarify how lobbies are created specifically for tournaments.

src/chigame/tournaments/models.py
```diff
@@ -0,0 +1,89 @@
...
+
+    chat = models.ForeignKey(Chat, on_delete=models.CASCADE)
+    content = models.TextField()
+    timestamp = models.DateTimeField(auto_now_add=True)  # timestamp of the
+    # moment the message was created.
+    sender = models.ForeignKey(User, on_delete=models.CASCADE)  # Even if the user is deleted, the message
+    # will still exist.
+
+    def __str__(self):
+        return "Message from " + self.sender + ": " + self.content
```
jb-01: This matches the ERD, but the `__str__` method may not work as intended because it tries to concatenate a User object with a string, which would raise a TypeError. Instead, it should reference a string attribute of the sender, such as `self.sender.username`.

src/chigame/tournaments/models.py
jb-01: This is true. An additional thing to keep in mind: The `Chat` model only has a ForeignKey to `Match`, which implies a Many-to-One relationship, not a One-to-One as stated in the ERD.

jb-01 (changes requested):
This `models.py` is a great start to a faithful implementation of our ERD.

To summarize my comments:

1. **Chat Relationship**: We need clarity from the Matches team regarding lobby setup to properly initialize Chats. Currently, the Chat model is set up with a ForeignKey to Match, suggesting multiple Chats could be associated with a single Match, which conflicts with our ERD's specification of a one-to-one relationship. We should switch to a `OneToOneField` for accuracy.

2. **__str__ Method in Message Model**: The current `__str__` implementation in Message may result in a TypeError due to the concatenation of a User object with a string. Modify it to use `self.sender.username` to ensure we're returning a string representation.

Please address these issues and request a re-review.

src/chigame/games/models.py
```diff
@@ -63,3 +63,97 @@ class MatchProposal(models.Model):
...
+            + self.end_date.strftime("%m/%d/%Y")
+        )
+
+
+class Notification(models.Model):
+    """
+    A notification, which can be sent to multiple users.
+    """
+
+    receipients = models.ManyToManyField(User, related_name="notifications")
```
elizabethli31: typo

src/chigame/games/models.py
```diff
@@ -63,3 +63,97 @@ class MatchProposal(models.Model):
...
+    timestamp = models.DateTimeField(auto_now_add=True)
+
+    # visibility: a smallPositiveIntegerField representing the visibility of
+    # the notification. This may be added later.
+
+    def get_all_receipients(self):
+        return self.receipients.all()
+
+    def __str__(self):  # may be changed later
+        receipients_str = "&".join([str(receipient) for receipient in self.get_all_receipients()])
```
elizabethli31: Same with the comment Joshua made about messages and User. I believe you'll need to do something like recipient.name. Also typo again

elizabethli31 (changes requested):
Generally, everything looks good and reflects the ERD. I left some minor comments on typos and the str method. Have you guys tested making any objects yet?

src/chigame/games/models.py
```diff
@@ -63,3 +63,97 @@ class MatchProposal(models.Model):
...
+    timestamp = models.DateTimeField(auto_now_add=True)
+
+    # visibility: a smallPositiveIntegerField representing the visibility of
+    # the notification. This may be added later.
+
+    def get_all_recipients(self):
+        return self.recipients.all()
+
+    def __str__(self):  # may be changed later
+        recipients_str = "&".join([str(recipient) for recipient in self.get_all_recipients()])
```
elizabethli31, line 226: This is incurring an infinite loop. You should just do self.recipients.all() instead of the get_all_recipients function.

elizabethli31 (changes requested):
Change notification to announcements like we decided in discussion today. Added a comment about an infinite loop.

jb-01 (approved):
Thanks for making the changes I mentioned and fixing the infinite string error.

Great work!

elizabethli31 (approved):
Working now and looks good!

MikeChen012345:
Thank you so much!

MuyanXie:
Closing statement:

In this PR, @MuyanXie  and @MikeChen012345 worked on hard coding the scratch for models.py of tournaments. @Riyush worked on interpreting the ERD and make sure everything is in alignment. @jb-01 worked on reviewing and refining any errors in this PR

elizabethli31:
Issue Score: Excellent

Comments:
Huge PR and you guys did a great job implementing your ERD. Updates should have been provided in response to the reviews and any changes you made. Regardless, great work on providing the foundations of the tournament feature.

PR #140, opened 2023-11-06
Updated lobby model to Incorporate Functionality Described in User Stories

### LINKED ISSUE #113: Redesign and implement the lobby class according to user stories
**Goal:** [Data Modeling for Lobby] Based on our user stories from Sprint 1, the current lobby table needs more attributes (or perhaps a different way of modeling). For example, user stories mention match start times as well as the modified game features. We need to redesign and implement the Lobby class to account for these additional features.

**Possible Deliverables:** 
- ERD
- Pull Request to implement new tables/classes
- Documentation detailing changes

nugerhard:
This is my attempt at the minimal lobby model, as we described it in [https://github.com/uchicago-cs/chigame/wiki/Matches-~-User-Stories#findingjoining-a-lobby](url). This is mostly concerned with time control and displaying lobbies on the lobby list page. For the former, we need to only keep track of when the lobby is started, and when the game should start by. For the latter, we need to essentially be able to distinguish lobbies that may still be joined, and those which can only be spectated. More of these thoughts to be included in #113.


nugerhard:
Ok, I didn't upload my database migrations with the updated models.py file. Looks like the run-tests test failed, and it is probably due to that? What's the protocol on this.

jiyeongchloe:
> Ok, I didn't upload my database migrations with the updated models.py file. Looks like the run-tests test failed, and it is probably due to that? What's the protocol on this.

I tried pushing the migrations, but it still hasn't solved the issue. I'm not sure if I did it correctly?

jiyeongchloe:
Got it to pass the CI checks! (had to define defaults for some of the attributes)

xylucien (approved):
looks good to me. have tested it on my computer locally and it runs OK!

src/chigame/games/models.py
```diff
@@ -23,12 +24,25 @@ class Lobby(models.Model):
...
+
+    match_status = models.PositiveSmallIntegerField(choices=STATUS, default=1)
     name = models.TextField()
     game = models.ForeignKey(Game, on_delete=models.CASCADE)
+    game_mod_status = models.PositiveSmallIntegerField(choices=MODS, default=1)
     created_by = models.ForeignKey(User, on_delete=models.CASCADE)
     members = models.ManyToManyField(User, related_name="lobbies")
     min_players = models.PositiveIntegerField()
     max_players = models.PositiveIntegerField()
+    time_constraint = models.PositiveIntegerField(default=3)
```
elizabethli31: Is there a reason this is set to 3?

elizabethli31 (approved):
LGTM! Why did you guys set the time constraint to 3? Other than that, status and mods look good! Maybe add a lobby finished time for the future. You wouldn't need it now but I could see that factoring in for historical data more than lobby_created time. Once you guys update with main, I'll merge to dev.

src/chigame/games/models.py
jiyeongchloe: I was initially thinking 3 minutes, but I think it makes more sense to have seconds as the unit (changed the default to 300 seconds/5 minutes)

elizabethli31:
Issue Score: Excellent

Comments:
Great job! Simple changes that are definitely necessary for the functionalities you guys want to add. Also, good work debugging the CI checks. My only comment is that you are lacking a closing statement.

PR #147, opened 2023-11-07
Accepting and declining friend requests and profile view reflect status of friend invitations

### LINKED ISSUE #139: (user profiles) friending: accept/decline functionality
We need to display "Accept" and "Decline" options for users when visiting profile pages that send them a friend invitation.  Trying to accept/decline an invitation that does not exist or where you are not the receiver should redirect you back to your profile page with a proper error message. Accepting an invitation adds user2 to user1 friends list and adds user1 to users2 friends list while declining an invitation deletes that instance of FriendInvitation.

omarelkashef:
This handles the friending buttons in the profile views. The page displays accept or decline when visiting profile pages that send them a friend invitation. It displays remove friend if they are already friends. If they have send that user a friend invitation before, it displays cancel friend requests. If a user tries to accept/decline an invitation that does not exist or where they are not the receiver they are redirected back to their profile page with a proper error message. If they accept invitation, they become friends and if they decline the invitation is deleted.  Currently the remove friend button does not redirect to anything. This will be handled in a separate issue.

### DIFF
--- a/src/chigame/users/models.py
+++ b/src/chigame/users/models.py
@@ -78,10 +78,13 @@ class FriendInvitation(models.Model):
 
     def accept_invitation(self):
         sender = self.sender
+        sender_profile = UserProfile.objects.get(user__pk=sender.pk)
         receiver = self.receiver
-        sender.friends.add(receiver)
-        receiver.friends.add(sender)
+        receiver_profile = UserProfile.objects.get(user__pk=receiver.pk)
+        sender_profile.friends.add(receiver)
+        receiver_profile.friends.add(sender)
         self.accepted = True
+        self.save()
 
 
 class Group(models.Model):
--- a/src/chigame/users/models.py
+++ b/src/chigame/users/models.py
@@ -76,6 +76,13 @@ class FriendInvitation(models.Model):
     accepted = models.BooleanField(default=False)
     timestamp = models.DateTimeField(auto_now_add=True)
 
+    def accept_invitation(self):
+        sender = self.sender
+        receiver = self.receiver
+        sender.friends.add(receiver)
+        receiver.friends.add(sender)
+        self.accepted = True
+
 
 class Group(models.Model):
     """
--- a/src/chigame/users/urls.py
+++ b/src/chigame/users/urls.py
@@ -1,6 +1,7 @@
 from django.urls import path
 
 from chigame.users.views import (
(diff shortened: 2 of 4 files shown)


henrietta-k (changes requested):
On lines 93 - 94 in `src/chigame/users/views.py` what is the behavior after the message "You already have a pending friend invitation from this profile." shows up? Maybe having a return statement after this with a redirect might be helpful (similar to line 97). Other than that, great job catching all the scenarios that could happen with an invitation!

omarelkashef:
> On lines 93 - 94 in `src/chigame/users/views.py` what is the behavior after the message "You already have a pending friend invitation from this profile." shows up? Maybe having a return statement after this with a redirect might be helpful (similar to line 97). Other than that, great job catching all the scenarios that could happen with an invitation!

it will also return because if the function goes to 94 then it won't go into the else branch and would go directly to the return statement on 97 which returns with whatever messages got picked up along the way

henrietta-k (approved):
Thanks for clarifying! Looks good to me.

elizabethli31 (approved):
LGTM! Tested everything and all endpoints look great. Can you resolve the conflict with dev, and then I'll merge.

omarelkashef:
Fixing merge conflict was a little challenging because this both PR and #127 were changing in the same functions. So, I resolved the conflict and pushed again.

omarelkashef:
Closing statement: 
This addresses the functionality of friend request buttons within profile views. When visiting profile pages that have sent a friend invitation, the page now appropriately displays options to accept or decline the invitation. If the users are already friends, the option to remove the friend is shown. Additionally, if a friend invitation had been sent previously, the page now provides the option to cancel the friend request. In case a user attempts to accept or decline an invitation that does not exist or where they are not the intended recipient, they will be redirected back to their profile page with a relevant error message. Upon accepting an invitation, the users become friends, and declining results in the removal of the invitation. It's worth noting that the functionality of the "remove friend" button is pending redirection and will be addressed in a separate issue.

@24raniwalar  worked on designing the new required endpoints
@omarelkashef  worked on the logic of the friending feature

elizabethli31:
Issue Score: Excellent

Comments:
Nice PR that made it possible for users to become friends. Also, good summary description,  especially highlighting that the remove friend functionality was not covered in this PR.

Issue #39, opened 2023-10-23
(game management) Create updated game entity to include attributes we want to detail a game (game rules, title, difficulty, etc.)


willcravitz:
There are several attributes of a game (rules, difficulty, average play time, etc.) that are not currently reflected in the `Game` model. We will write a new entity (to be used in a later ERD) that includes the attributes which will be stored in the db and accounts for those which we will potentially retrieve from a remote API.

frowenz:
## Proposed Additions to the Game Model

Given the proposed integration of the [Board Game Geek API](https://github.com/uchicago-cs/chigame/issues/45#issuecomment-1782179661) to assist staff users with autofill capabilities when creating a new game, it's essential to align our attribute fields for compatibility. The BGG API also provides several useful categories which we should consider, one of which includes [mechanics](https://boardgamegeek.com/browse/boardgamemechanic). For further details, refer to [this wiki page](https://github.com/uchicago-cs/chigame/wiki/Games-~-BoardGameGeek-(BGG)-API).

**Current Game Model:**
```python
class Game(models.Model):
    """
    A game (Chess, Checkers, etc.)
    """

    name = models.TextField()
    description = models.TextField()
    min_players = models.PositiveIntegerField()
    max_players = models.PositiveIntegerField()
```

**Initial Additional Proposals for a game model:**
* Designer (usually one, but possibly many)
* Artists (likely many)
* Publisher (possibly many depending on the version)
* Year Published
* Game Thumbnail
* Playtime:
    * Minimum
    * Maximum
    * Expected
* Rules
* Suggested Age
* Complexity
* Popularity metrics:
    * Number of times played
    * Number of Favorites
* Corresponding BGG ID (could be useful for updating game's attributes) 

**Other considerations and open questions:**
- How do we handle rules?
    - Some Possible Options: 
        - Just text (probably simple and good for now)
        - Board Game Arena has an external wiki that they embed on their website. See [BGG Chess Rules](https://en.boardgamearena.com/gamepanel?game=chess)
        - Just link to the rulebook. This depends on things being online and may not capture nuances of online play. [Example](https://cdn.1j1ju.com/medias/88/21/c2-blood-rage-rulebook.pdf)
- How do we handle images? 
    - Some Options: serialize image or upload somewhere and link to the location.
- The richness of data in Board Game Geek, like internal popularity metrics, could be useful. However, it could be tricky to use if not every game on ChiGame is linked with a BGG game. 
[...]

PR #153, opened 2023-11-07
Update game model

### LINKED ISSUE #92: Complete the game model in django to have the updated attributes
We need to update the Game model in Django (in `chigames/games/models.py`) with the required attributes specified in #39. This will also necessitate updating the Game Detail View, Update View, and Create View to show these attributes. A provisional list of attributes to be added are rules, expected playtime, minimum playtime, maximum playtime, designers, artists, publisher, year published, game thumbnail/image, complexity, and BGG id.

frowenz:
Added the game attributes discussed in #39. This includes game information like designers, an image thumbnail, game category and more. Currently, there are multiple issues waiting on having a game model with these attributes.


src/chigame/games/models.py
```diff
@@ -6,13 +6,93 @@
...
     """
-    A game (Chess, Checkers, etc.)
+    A game like Chess, Checkers, etc.
     """
 
-    name = models.TextField()
+    # Basic information
+    name = models.CharField(max_length=255)
     description = models.TextField()
+    image = models.URLField(default="/static/images/no_picture_available.png")
```
willcravitz: Might not be relevant for this particular PR, but are we going to store all game images in this folder and in the repo?

src/chigame/games/models.py
```diff
@@ -6,13 +6,93 @@
...
+    year_published = models.PositiveIntegerField(null=True)
+
+    # Gameplay information
+    rules = models.TextField(null=True)
+
     min_players = models.PositiveIntegerField()
     max_players = models.PositiveIntegerField()
+    suggested_age = models.PositiveSmallIntegerField(null=True)
+
+    expected_playtime = models.PositiveIntegerField(null=True)
```
willcravitz: For game creators, is this in seconds, minutes, hours?

src/chigame/games/models.py
```diff
@@ -6,13 +6,93 @@
...
     description = models.TextField()
+    image = models.URLField(default="/static/images/no_picture_available.png")
+    year_published = models.PositiveIntegerField(null=True)
+
+    # Gameplay information
+    rules = models.TextField(null=True)
+
     min_players = models.PositiveIntegerField()
     max_players = models.PositiveIntegerField()
+    suggested_age = models.PositiveSmallIntegerField(null=True)
```
willcravitz: Is this minimum age? I.e. if this field is 10 then is the game "10+"?

src/chigame/games/models.py
```diff
@@ -6,13 +6,93 @@
...
+    expected_playtime = models.PositiveIntegerField(null=True)
+    min_playtime = models.PositiveIntegerField(null=True)
+    max_playtime = models.PositiveIntegerField(null=True)
+
+    complexity = models.PositiveSmallIntegerField(null=True)
+    category = models.ManyToManyField("Category", related_name="games")
+    mechanics = models.ManyToManyField("Mechanic", related_name="games")
+
+    # BGG information
+    BGG_id = models.PositiveIntegerField(unique=True, null=True)
```
willcravitz: Another very small change, but I think this should not be unique just in case we end up supporting multiple versions of the same game

willcravitz (approved):
I ran the new migrations and created a new game object using the Django shell. All of the new attributes worked as expected. Here is the output on the admin page:
image

I left a few comments on exactly how to interpret some of the attributes. This won't affect the model but will affect how we display info on games.

I also flagged a quick question about how we're handling image storage more generally. 

Great job!

src/chigame/games/models.py
frowenz: No. Suppose we had thousands and thousands of images, it would start bloat the repo. 

I googled around and the recommendations I found are that  images should be hosted on either a separate server or  blob storage service.

src/chigame/games/models.py
frowenz: Minutes. I will add a comment.

src/chigame/games/models.py
frowenz: Yes. I will add a comment.

src/chigame/games/models.py
frowenz: Good suggestion!

frowenz:
I added some comments to address vagueness. I removed the BGG unique id constraint.

src/chigame/games/models.py
majorsylvie: this would actually be a good thing to comment in the code itself! That way anyone looking at the game model for whatever reason will be able to see that there is a more durable longer term solution that can/eventually should be implemented

src/chigame/games/models.py
```diff
@@ -6,13 +6,95 @@
...
+
+    DESIGNER = 1
+    ARTIST = 2
+
+    ROLE_CHOICES = (
+        (DESIGNER, "Designer"),
+        (ARTIST, "Artist"),
+    )
+
+    name = models.CharField(max_length=255)
```
majorsylvie: Is there a particular reason you're setting this max length?

src/chigame/games/models.py
```diff
@@ -6,13 +6,95 @@
...
+    """
+    A person associated with a game, such as a designer or artist.
+    """
+
+    DESIGNER = 1
+    ARTIST = 2
+
+    ROLE_CHOICES = (
+        (DESIGNER, "Designer"),
+        (ARTIST, "Artist"),
```
majorsylvie, line 58: Given that there are only currently two roles listed, how will you decide what other roles to add?
Things like gameplay tester, director, translator, etc... 

Also can this model reasonably handle the situation where one person has multiple roles on the same work? Such as a solo dev doing all of the parts themselves, or a duo team coding and writing everything?

majorsylvie (changes requested):
Overall, looks quite good! Very happy to see much more information get associated with the model, and that many of these choices were influenced by Board game geek (and then properly documented as such).

I've got a few questions on functionality and design decisions that should be answered before merging.

As well, tests have been added in CI, and those are failing due to migration related error. Please locally merge the newest changes from dev and resolve the tests such that they would only fail from discrepancies of the game model, instead of migration issues.

src/chigame/games/models.py
frowenz: I was swayed by this

How to use CharField ?
CharField is used for storing small sized strings in the database. One can store First Name, Last Name, Address Details, etc. CharField should be given an argument max_length for specifying the maximum length of string it is required to store. In production server, after the Django application is deployed, space is very limited. So it is always optimal to use max_length according to the requirement of the field. Let us create an instance of the CharField we created and check if it is working.

from here:

https://www.geeksforgeeks.org/charfield-django-models/#

src/chigame/games/models.py
frowenz: These two roles are listed on board game geek which makes them easy first choices. We were thinking that other roles could just be added as needed. While still have not decided what other ChiGame specific people information needs to be tracked, this person model seems ideal for tracking simple relationships (e.g. developer of a game) as opposed to something more complex (staff members who created or edited a game). 

In regards to the second question, I think it is fine to have one person with multiple roles have different entries for each role. See Pascal Trigaux in [this](https://boardgamegeek.com/xmlapi2/thing?id=20&stats=1) BGG API response.

frowenz:
I have added more descriptive comments about design choices, pulled the most recent version of the dev branch in and removed the char fields.

src/chigame/games/models.py
majorsylvie: I talked with borja about this, he stated that in modern day string limits are not particularly necessary since fixed length storage optimizations are no longer that important.
For the purposes of chigame, we can leave the `CharField`'s without `max_length` option.

src/chigame/games/models.py
majorsylvie: Sounds good! Thank you for the clarification and citation.

src/chigame/games/models.py
```diff
@@ -6,13 +6,101 @@
...
+
+    expected_playtime = models.PositiveIntegerField(null=True)  # In Minutes
+    min_playtime = models.PositiveIntegerField(null=True)
+    max_playtime = models.PositiveIntegerField(null=True)
+
+    complexity = models.PositiveSmallIntegerField(null=True)  # 1-5, 1 being the easiest
+    category = models.ManyToManyField("Category", related_name="games")
+    mechanics = models.ManyToManyField("Mechanic", related_name="games")
+
+    # ================ OTHER ================
```
majorsylvie, line 40: Appreciate the sub-divisions of fields. For a particularly important and large model like the game model this makes it significantly more readable. This is _especially true for those outside the team._

majorsylvie (approved):
Thank you for your clarifications on your choices for the person model.
Overall, having the model be informed by the BGG api makes sense for your aspirations of the autocomplete and new-game-autofilling.

With CI passing this is good to go! 
Thank you for your work.

ptlin25:
The Game model now has the attribute fields for year published, rules, expected playtime, min/max playtime, BGG id, image, suggested age, and relationships with Publisher, Person, Mechanic and Category as specified by #39. 
@ptlin25 created the branch and added attribute fields for year published, rules, expected playtime, minimum and maximum playtime, and BGG id.
@danielngira created an initial migration for the basic attributes and created a Person model
@frowenz finished the attributes and added new models for Publisher, Mechanic, and Category

majorsylvie:
Issue Score: Excellent

Comments:
Incredible job all of you! 

@willcravitz commendable review! Asking clarifying questions about the code at this smaller level is very important and valueable.

@ptlin25 @danielngira @frowenz great job formatting the attributes of the new game model as well as taking feedback about code comments!

Great work and keep moving forward y'all!

PR #155, opened 2023-11-07
matches/Updated Match and Player models to track information pertinent to match outcomes

### LINKED ISSUE #109: (matches) data modeling for matches (Annika)
Goal: We need a way of storing match outcome data, such as win records, point totals, and resources gained. This data will need to be referenced by future matches and by the users who have played in the match.

Deliverables: An ERD and tables for match outcome data, and documentation detailing why certain decisions were made and intended uses for the data structures.

ambr-lise:
PR for Issue #109. Updated games/models.py according to the User Stories for matches. 

Player (throughtable for User and Match): 
- added team, role, outcome, and victory_type parameters to track the user's team and role (if applicable), whether the user won or withdrew, and how the user won (if applicable)

Match: 
- removed winners, which made tracking draw/withdraw/neutral outcomes more difficult (this information has essentially been migrated to Player)
- added date_played to track when the match was played


jiyeongchloe (approved):
Everything looks good to me! Is the `vicotry_type` attribute in Player implemented with specific games in mind? I'm just wondering what we'd use that for if we have an `outcome` attribute

ambr-lise:
> Everything looks good to me! Is the `vicotry_type` attribute in Player implemented with specific games in mind? I'm just wondering what we'd use that for if we have an `outcome` attribute

That's correct -- `victory_type` would essentially expand upon a "win" outcome. Certain games have different win conditions -- for example, in Root you could win via a Points Victory or a Domination Victory. However, `victory_type` is optional as many games only have one win condition. I also think it allows for some additional flexibility in case there's a game that falls outside of the win/loss/draw/withdraw outcomes.

src/chigame/games/models.py
```diff
@@ -61,8 +61,24 @@ class Player(models.Model):
...
+        (DRAW, "draw"),
+        (LOSE, "lose"),
+        (WITHDRAWAL, "withdrawal"),
+    )
+
     user = models.ForeignKey(User, on_delete=models.CASCADE)
     match = models.ForeignKey(Match, on_delete=models.CASCADE)
+    team = models.TextField(blank=True, null=True)
+    role = models.TextField(blank=True, null=True)
+    outcome = models.PositiveSmallIntegerField(choices=OUTCOMES)
```
elizabethli31: This should be nullable in case a player is currently in a match that hasn't been completed yet.

elizabethli31 (changes requested):
Looks good besides the one comment I made! Make sure you change that and update with dev, and I'll review again.

elizabethli31 (approved):
I like the addition of a player! It makes a lot of sense to separate user from player and simplifies the match process. LGTM! I'll merge after dev changes are made.

ambr-lise:
Closing Statement: Added attributes for additional data tracking to Player and Match tables, and migrated some functionality from Match to Player.

elizabethli31:
Issue Score: Excellent

Comments:
Nice work on this PR. I really like the clear distinction between what was done for the match model and player model in the description of this PR.

PR #156, opened 2023-11-07
Initial inbox feature includes button and list of notification IDs

### LINKED ISSUE #124: Notifications: create inbox
Create an inbox that contains all new updates such as new friend requests. It should show unread notifications from most recent to least recent. Users should have the ability to delete notification from their inbox. Clicking on an update should redirect to a page about this notification depending on what type of notification it is. For example, if it is a friend invitation, it should direct to a the profile of sender. The profile page should have a button to accept or reject the invitations which is handled in #125 . Refreshing a profile page should show new incoming notifications if there is any. We might want to display notifications in real time in the future.

24raniwalar:
This PR involved adding a "view" function for our inbox feature, adding the "inbox" button to the base template and only allowing access if a user is logged in, and adding a template for the inbox that currently just displays the notification numbers of the current notifications a user has. Another PR has a tables.py file that we will use to create a notifications table once it has been pushed.


ishiii-s (commented):
Overall, I don't think there are any changes I need to request for this code. Changes to migration and views appear logical to make considerations for user authentication and resulting access to inbox, and url structure seems to match these needs. Changes to template files to add inbox button and eventually display notification numbers also appear logical. Everything looks good to me.

src/chigame/users/views.py
```diff
@@ -120,3 +120,11 @@ def cancel_friend_invitation(request, pk):
     else:
         messages.error(request, "Something went wrong please try again later!")
     return redirect(reverse("users:user-profile", kwargs={"pk": request.user.pk}))
+
+
+@login_required
+def user_inbox_view(request, pk):
```
elizabethli31, line 187: You should add some sort of error handling, so that the current user can only see their own inbox. I tested with objects user1, user2, and a notification sent to user1. When I'm logged in as user2, I can't see anything in inbox. But when logged in as user1, I see the same notification at "users/inbox/pk" for any pk. This would be fixed simply by limiting to only being able to view the user's own inbox.

elizabethli31 (changes requested):
Inbox looks good and generally works properly! I left one comment about making sure that the user cannot view inboxes at other urls besides the one associated with their own.

src/chigame/users/views.py
24raniwalar: Modified!

elizabethli31: I think it would make more sense to just redirect back to your page and also have a message that says "Not your inbox". When I try to access inboxes of users that haven't been created yet, it still gives the same error, but these users shouldn't have an inbox page at all.

elizabethli31 (changes requested):
Your fix is a great start, but I think it needs to be a little more robust. I left a comment explaining why.

elizabethli31 (approved):
Looks good! Thanks for making the changes I suggested. Update w dev and I'll merge

AndrewM131 (changes requested):
When trying to access the inbox through the nav item, I get directed to the template user_inbox.html, but I get an "Error during template rendering" from line 4 in user_inbox.html. It says "no such table: users_notification." Since it's directing me towards the HTML correctly, and I don't see any issues with your code, your urls.py, views.py, and base.html edits look good to me. The only issue is it's not rendering. Currently, the user I set up trying to access this page has no notifications, so it may be due to this? If this is something just on my end, please let me know so I can resolve it.

Also, this may be more of a style comment, but the My Inbox in the Nav bar should be before sign out.

Edit: Rishabh suggested I make sure I make migrations, and it helped. It now looks good to me. Thank you for helping me out and addressing my style comment.

24raniwalar:
I don't think you are running the right branch / commit as there is no users_notifications table being used. Alternatively, you may not have migrated the new models that use notifications. I have moved the inbox to be before the sign out button.

AndrewM131 (approved):
Thanks for addressing my style comment, your code is good to go!

24raniwalar:
Closing Statement: This PR added an inbox button with initial functionality that displays the notification ID of any notifications currently assigned to a user. Changes requested and addressed include allowing users to only access their own inbox and redirect users otherwise with an error message along with a design change that shifted the inbox button to before the sign out button. Omar and I worked on this functionality together.

elizabethli31:
Issue Score: Excellent

Comments:
Great initial inbox PR that's been drastically improved upon since. Only comment is that I would have liked to see more updates in response to the changes made in each review.

PR #160, opened 2023-11-09
Added querying capabilities to the Game List API Endpoint

### LINKED ISSUE #159: Adding Filtering capabilities to the Game List View API Endpoint
Adding querying capabilities to the Game list view endpoints: `/api/games/?property=key`

For example, we may want to find all the games that have `min_players=1` or `max_players=5`. We may want to add the functionality of having some way to query the our Games dataset instead of always just returning all the games in our database. To do this, we can use the Filtering class in Django.

Example Endpoints:

`/api/games/?min_players=1` or `/api/games/?max_players=5`

We will know when this issue is completed when we can correctly filter our list of games using the `max_players`, `min_players`, and `name` attributes of the Game model

abrahmasandra:
Added querying capabilities to the Game list view endpoints: `/api/games/?property=key`

Instead of just getting a list of all Games, a user may want to find games, such that a certain condition is true. For example, find all the games that have `min_players=1` or `max_players=5`. We may want to add the functionality of having some way to query the our Games dataset instead of always just returning all the games in our database. To do this, we can use the Filtering class in Django through `django-filters`.

To query using the `name` of the game, the API will return all games such that the query string is contained in the actual name of the game. This gives a similar functionality to searching in a database, rather than abiding by a strict equality.

Some example endpoints:

`/api/games/?min_players=1` OR `/api/games/?max_players=5`


cuyakonwu (approved):
Code is looking good, may have to look back at game class

src/chigame/api/filters.py
```diff
@@ -0,0 +1,13 @@
+import django_filters
+
+from chigame.games.models import Game
+
+
+class GameFilter(django_filters.FilterSet):
+    name = django_filters.CharFilter(lookup_expr="icontains")
```
majorsylvie, line 9: I would add a comment to this line with a reference to documentation where we can see what `icontains` means.
As well as a short explanation similar to what you wrote in the description, detailing how its a substring match.

src/chigame/api/filters.py
```diff
@@ -0,0 +1,13 @@
...
+    min_players = django_filters.NumberFilter(lookup_expr="exact")
+    max_players = django_filters.NumberFilter(lookup_expr="exact")
+
+    class Meta:
+        model = Game
+        fields = ["name", "min_players", "max_players"]
```
majorsylvie, line 15: With the merge of the new game model into dev, this will need to be updated to include the new fields. This can be in a new issue, and for now the game endpoint and filtering will still work, even if not displaying all of the info for the game.

src/chigame/api/views.py
```diff
@@ -1,13 +1,17 @@
...
 
+from chigame.api.filters import GameFilter
 from chigame.api.serializers import GameSerializer
 from chigame.games.models import Game
 
 
 class GameListView(generics.ListCreateAPIView):
     queryset = Game.objects.all()
     serializer_class = GameSerializer
+    filter_backends = (DjangoFilterBackend,)  # Enable DjangoFilterBackend
```
majorsylvie, line 14: Thank you for the code comments :)

majorsylvie (changes requested):
Really cool functionality to add!

There are some small and fixable merge conflicts to address, as well as merging again from dev and adding comments.

Make separate issues to update the endpoints and querying in correspondence with the new game model.

Soon to be good!

src/chigame/api/filters.py
abrahmasandra: added comment about `icontains`!

src/chigame/api/filters.py
abrahmasandra: I will create a new issue to update the filtering fields for the games.

majorsylvie (changes requested):
The endpoint works well and supports simple compound queries which is nice, ie: `http://localhost:8000/api/games/?min_players=2&max_players=2`

As mentioned in previous comments the updates to the game model means more attributes will have to be added, though this is still a functional endpoint. 

Make another issue as mentioned to update the filtering and another one to update the game detail/list to support more of the new fields.

Something which could be interesting to try is adding support for multiple queries to the same field in the same request.
eg: `http://localhost:8000/api/games/?max_players=3&?max_players=4` gets you all the games with either 3 or 4 `max_players`

This is not entirely necessary, since for queries like this we can shift the burden of making more requests to the end-user. Making them request `?max_players=3` and `?max_players=4` separately.

There is a small merge conflict in the imports for `models.py`, fix those and this is good to merge!

abrahmasandra:
@majorsylvie I did create another issue to update the Game filtering. I will also create an issue now to update the filtering to support multiple filtering queries for the same attribute, i.e. `http://localhost:8000/api/games/?max_players=3&?max_players=4`

majorsylvie (approved):
Thank you for resolving the code conflict! 

Excited to see how querying could progress over time now that we have this solid foundation.

abrahmasandra:
Closing Comment: 

Added querying capabilities to the Game list view endpoints: `/api/games/?property=key`. Now a user can filter the list view of Games instead of just getting all games. Specifically, added the filtering capabilities for the Game attributes: `name`, `min_players`, `max_players`

majorsylvie:
Issue Score: Excellent

Comments:
This issue would have benefitted from more update comments! 
For example there was the commit: https://github.com/uchicago-cs/chigame/pull/160/commits/2114c2152ed0facb7b4d09e17bc91d538905be3f about "removing unnecessary field" from the filter.
A statement that that happened as well as why would be a great update.

Besides that, great summary and closing comment.

PR #165, opened 2023-11-09
Update API Endpoints dealing with Game model (List and Detail Views)

### LINKED ISSUE #166: Update the API Tests dealing with Game model (Testing List and Detail Views)
In issue #164, we detailed the ways that the Game endpoints needed to be changed in order to have functional API Endpoints that display and allow for the creation of Game objects with all the new attributes added to the models. Now, we also need to update the tests for these API Endpoints, because the old tests did not take into account all the new attributes that would be added to the Game model. 

Changing these tests are mandatory for the PR #165 to pass, because these tests need to pass before the new code for the List and Detail views can be merged to the `dev` branch of the repository.

abrahmasandra:
Pull Request https://github.com/uchicago-cs/chigame/pull/153 changed the Game model to be more fleshed out, taking inspiration from BoardGameGeeks. Due to the changes in the model, we changed our API Endpoint Views dealing with the Game model, so that the model attributes are returned correctly. There are additions to the model, as well as changes to the existing attributes of the Game model.

Specifically, the `GameSerializer` now has the line `fields = "__all__"`, which implies that we want to return all the fields in the model, when serializing and deserializing a Game. This is also dynamic, as we don't have to add fields/rename fields manually, when the Game class changes, as the serializer will just return all the fields in the model.


Esterello2 (approved):
Adding the "__all__" for the field works fine.

majorsylvie (changes requested):
You've got merge conflict with dev right now. Please resolve those and make sure your Game Endpoints work as expected (List/Detail GET, POST, PATCH, and DELETE) with no merge conflicts then this will be good to go!

abrahmasandra:
This PR has been closed, but its content has been duplicated in PR #289

PR #179, opened 2023-11-10
Fixed bug where fields that are supposed to be nullable were required on forms

### LINKED ISSUE #178: Make new Game model attributes nullable
Many of the new attributes that have been added to the Game model should be nullable (not required). However, this is currently not working because `blank=True` was not set for these attributes. As a result, trying to fill out a form without filling out these attributes creates an error. The attribute `blank=True` needs to be added to all of the attributes that are not required.

ptlin25:
Added `blank=True` to nullable fields, so Django form validation does not fail when the fields are not filled in. The image attribute will also be stored as a text field rather than a URL because our default image is stored locally and is not a valid URL.


frowenz (approved):
This all looks good. Nice job catching that `URLField` needed to change to `TextField`.

majorsylvie (approved):
Looks good! Thank you for catching the URLField. 

In general when you want something to be optional in django, you will basically always be doing both `null=True` and `blank=True`

ptlin25:
The image field in the Game model is now a text field instead of a URL field. The optional attributes in the Game model are now `null=True` and `blank=True`, so they can now be left blank on forms.

ptlin25:
`blank=True` has been added to the optional attributes in the Game model, so they do not need to be filled in on forms. This fixes the issue of the optional fields not being optional when creating and editing forms, so this task has been completed

majorsylvie:
Issue Score: Excellent

Comments:
Great minor PR updating a noticed bug! 
Great job following the issue/PR guidelines even for a smaller task.

PR #183, opened 2023-11-11
add ignore=E203 to flake8

MikeChen012345:
The line "ignore = E203" is added to the setup.cfg file such that the conflict between flake8 and black can be resolved. The error is raised by flake8 when it detects whitespace before ':' which is against PEP 8 style guide recommendations for slices.
However, there's a known discrepancy between flake8 and black regarding this specific rule. black formats code with whitespace before and after the colon in slices for better readability, while flake8 raises a warning for this.


borjasotomayor:
This looks good to me but, for future reference, can you provide a link to where this is called out as a "known discrepancy"?

borjasotomayor:
Please also provide more details on why W=503 was added (you can just include the explanation you provided on Slack)

MikeChen012345:
The W503 warning is raised by flake8 when it detects a line break before a binary operator. This was once against PEP 8 recommendations, which suggested that line breaks should occur after the binary operator.
However, PEP 8 was updated and now recommends that line breaks should occur before the binary operator. This makes the code more readable.

MikeChen012345:
The E203 error was discussed in a bunch of issues. Below are two of them:
https://github.com/PyCQA/pycodestyle/issues/373
https://github.com/psf/black/issues/280

And the W503 error was discussed here:
https://www.flake8rules.com/rules/W503.html

borjasotomayor (approved):
Thanks for the additional information. This change looks good to me.

elizabethli31 (approved):
Looks good! Please update with dev and I'll merge.

elizabethli31:
Issue Score: Excellent

Comments:
Good job flagging and fixing this! General change still needs a closing statement.

PR #184, opened 2023-11-11
Helper methods( Filtering and Updating) for Notification model

### LINKED ISSUE #182: Add helper filtering methods to Notification
We need helper methods that facilitate using the new Notification model. Currently we can only get and filter by actor, but other important methods include get and filter by receiver in order to display in inbox and also filter by unread, read, deleted and mark multiple notifications as deleted, undeleted, read, unread. Another helpful method is to filter by type of notification.

omarelkashef:
This PR adds helper methods  that facilitate provide important operations to the Notification model. That is, we can now get and filter by receiver which would be very useful for the inbox. Also, we can filter by unread, read, deleted, and the type of the notification and mark multiple notifications as deleted, undeleted, read, unread.

### DIFF
--- a/src/chigame/users/models.py
+++ b/src/chigame/users/models.py
@@ -94,31 +94,43 @@ class GroupInvitation(models.Model):
 
 
 class NotificationQuerySet(models.QuerySet):
-    def filter_by_actor(self, actor, **kwargs):
+    def filter_by_actor(self, actor, include_deleted=False, **kwargs):
         try:
             actor_content_type = ContentType.objects.get(model=actor._meta.model_name)
             actor_object_id = actor.pk
-            return self.filter(actor_content_type=actor_content_type, actor_object_id=actor_object_id, **kwargs)
+            queryset = self.filter(actor_content_type=actor_content_type, actor_object_id=actor_object_id, **kwargs)
+            if not include_deleted:
+                queryset.is_not_deleted()
+            return queryset
 
         except ContentType.DoesNotExist:
             raise ValueError(f"The model {actor.label} is not registered in content type")
 
-    def get_by_actor(self, actor, **kwargs):
+    def get_by_actor(self, actor, include_deleted=False, **kwargs):
         try:
             actor_content_type = ContentType.objects.get(model=actor._meta.model_name)
             actor_object_id = actor.pk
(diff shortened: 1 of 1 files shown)


src/chigame/users/models.py
```diff
@@ -93,24 +93,68 @@ class GroupInvitation(models.Model):
...
+        self.update(read=True)
+
+    def mark_all_deleted(self):
+        self.update(visible=False)
+
+    def restore_all_deleted(self):
+        self.update(visible=True)
+
+    def is_read(self):
+        return self.filter(read=True)
```
AndrewM131, line 147: My only question is when is_read, is_unread, is_deleted, and is_not_deleted, how does the system know to only mark a specific notification as read, unread, deleted or not deleted? Does it need an additional input?

AndrewM131 (changes requested):

src/chigame/users/models.py
omarelkashef: These methods are called on querysets so for example for is_read, the queryset will have Notification where read is True

AndrewM131: And does the filter individualize it so only one is affected, and not all?

omarelkashef: All Notification that fit this criteria are included in the queryset.

AndrewM131: I see, thank you for the clarification. the helper methods look properly implemented in the Notification model.

AndrewM131 (approved):
LGTM, thank you for clarifying my questions.

elizabethli31 (approved):
Tested all methods, and everything works great. Very complete helpers as always. Good work! Update from dev and I'll merge

omarelkashef:
Closing statment:  In this pull request, we introduced helpful methods that streamline essential operations for the Notification model. With these additions, we can now retrieve and filter notifications by receiver. Additionally, we enabled filtering by unread, read, deleted, and notification type the ability as well as marking multiple notifications as deleted, undeleted, read, or unread. 

@24raniwalar worked on the mark_x() and is_x() methods
@omarelkashef  worked on the filter and get methods

elizabethli31:
Issue Score: Excellent

Comments:
Nice set of functions, and really clean code. Great job!

Issue #107, opened 2023-10-31
Details view for player


rqueiroz20:
This task involves creating a details view for a player in a tournament. They should be able to see the tournament type, the rules of the tournament, a list of players in the tournament, and some other details that have not been fully established yet.

MikeChen012345:
For the detail view, staff, non-admin users, and guests (not logged in) are all able to see the information related to the tournament. The fields displayed are ("name", "game", "start_date", "end_date", "max_players", "description", "rules", "draw_rules", "matches", "players", and "winners"). Note that we have merged issues #107 and #105  because currently, we recognize no need to create separate detail views for spectators and players, so only one detail view is created.

MuyanXie:
Closing Statement:

@rqueiroz20 worked on creating a scratch for what the classes and attributes should look like in the details view
@MikeChen012345 and @MuyanXie worked on implementing the idea in code

elizabethli31:
Issue Score: Unsatisfactory

Comments:
Marking this as unsatisfactory for grading purposes because this issue was merged with #105

Issue #105, opened 2023-10-31
Tournament details view for spectator


rqueiroz20:
This task requires us to create a details view for a spectator of a tournament. This shouldn't be too different from the details view of a player in a tournament, but we must ensure that a spectator can only join matches in a tournament to spectate a live game instead of playing in one. We could also add some high level information exclusive to spectators, but I'm not sure what those details would entail.

MikeChen012345:
For the detail view, staff, non-admin users, and guests (not logged in) are all able to see the information related to the tournament. The fields displayed are ("name", "game", "start_date", "end_date", "max_players", "description", "rules", "draw_rules", "matches", "players", and "winners"). Note that we have merged issues #107 and #105  because currently, we recognize no need to create separate detail views for spectators and players, so only one detail view is created.

MikeChen012345:
This issue is closed by PR #205 , we have implemented a basic create and detail view for the tournament model. During the process,
- Mike, Sam, and Joshua worked on the code. In particular, Mike and Sam mainly worked on the implementation of the views and HTML templates, while Joshua contributed a lot to the discussion of implementation details and mainly worked on giving feedback and suggestions from the List view that he was working on. 
- Rafael did the review. It was a pull request that took a long time, and he surely spent a lot of time reviewing the code. The effort is greatly appreciated!

elizabethli31:
Issue Score: Excellent

Comments:
This issue provided much needed functionality for the tournaments feature. Updates and closing comment were very detailed as well.

Issue #88, opened 2023-10-30
Create a create view for tournaments


MikeChen012345:
In this issue, we will create a create view for the tournaments. This page allows users (restrictions may apply, but are not determined currently) to create a new tournament and specify certain properties of the tournament (e.g., number of participants, tournament type, rewards, etc.).

MikeChen012345:
On the create view, users are able to create a tournament and assign values to most of the attributes (the link to the create view is located on the list view and only appears when the user is a staff member). Currently, we only allow staff to create tournaments, which may be changed later with restrictions. The code is on pull request #187

MuyanXie:
Closing Statement:

@MikeChen012345 and @jb-01 worked on coding the classes and functions for create view. @MuyanXie worked on reviewing the structure of the code as well as checking everything is under the right track for CRUD pages

elizabethli31:
Issue Score: Excellent

Comments:
Good work explaining exactly what this view is for in the update!

Issue #86, opened 2023-10-30
Create the list view for tournaments


MikeChen012345:
In this task, we would like to create a list view for tournaments, on which we will be able to see all the tournaments created. This is a part of the CRUD pages.

MikeChen012345:
For the list view, staff, non-admin users, and guests (not logged in) are all able to see the information related to the tournament. The fields displayed are ("name",  "game", "start_date", "end_date", "max_players", "description", "rules", "draw_rules", "matches", "players", "winners"). The code is on pull request #187

MuyanXie:
Closing Statement:

@MikeChen012345 worked on coding the soluting for the list view together with @MuyanXie . @jb-01 worked on researching the best way to represent a list view for tournaments

elizabethli31:
Issue Score: Excellent

Comments:
Nice work on this PR and for fixing all the merge conflicts.

PR #187, opened 2023-11-12
Add create and read views to the tournaments webpages

MikeChen012345:
We have implemented the create view, list view, and detail view for the tournaments (issues #88 , #86 , #107 , and #105 ). On the create view, users are able to create a tournament and assign values to most of the attributes (the link to the create view is located on the list view and only appears when the user is a staff member). Currently, we only allow staff to create tournaments, which may be changed later with restrictions. For the list view and detail view, staff, non-admin users, and guests (not logged in) are all able to see the information related to the tournament. Note that we have merged issues #107 and #105 because currently, we recognize no need to create separate detail views for spectators and players, so only one detail view is created.


MikeChen012345:
The pull request is closed at the suggestion of @elizabethli31 . A new pull request would be created to make the code look better and cleaner.

MikeChen012345:
The new Pull Request is #205

Issue #172, opened 2023-11-10
Add support for non-admin users to create or edit tournaments


MuyanXie:
This issue is a direct continuation of issue #173 . In this issue, we will discuss how we would allow non-staff users to create or edit tournaments, with a thorough documentation of the methods and a coming PR finalizing the changes

Documentations on Non-staff users:

- Allowed to use tokens to create Tournaments (contact the User team to get this done, which is also considered a scope within this issue).
- Allow to have partial edit control over their own tournaments; some actions shall be done by contacting staff users for assistance.
- Also for editing: would not be able to remove/add matches/players to a tournament.
- Allow to join tournaments that they have access to.
- Are not allowed to delete/disable the tournament they created but should contact staff users to do so because the deletion of a tournament can lead to a bunch of different events that need to be handled.
- Cannot access the tournaments that are not open to them.

MuyanXie:
@MikeChen012345 I think one issue regarding a possible PR is that should we include a way for the admin users to create/edit tournaments in the form of views/apis? Definitely they can just create a tournament by modifying the database, but I'm just wondering if you think we should provide a different function in views.py for admin users to create/edit tournaments

MikeChen012345:
For admin users to create, update, and delete a tournament, the approach is to use the link that only appears on the detail view and list view if users are logged in as an admin user. This has been implemented in the CRUD pages. However, since the CRUD pages have been split into 3 pull requests, currently the links are removed. We may need to add them back later.

MuyanXie:
Reminder: we need to add support to distribute new tokens every month

MuyanXie:
@MuyanXie worked on @Riyush worked on finalizing the logic of non-admin creating tournaments. @MikeChen012345 worked on coding and doing the details

elizabethli31:
Issue Score: Excellent

Comments:
Great start to making the tournaments feature more dynamic. Also, nice description that clearly defines that this task covers. My only comment is that this issue took two sprints to complete.

Issue #151, opened 2023-11-07
Create the update view for tournaments


MikeChen012345:
In this task, we would like to create an update view for tournaments, on which users (currently restricted to admins, and will get changed later when we figure out a way to allow non-admin users to create or edit tournaments) will be able to update (edit) all the tournaments created (or the tournaments created by themselves for non-admin users). This is a part of the CRUD pages.

MikeChen012345:
We have implemented the update view for the tournaments. Currently, only staff members are allowed to edit most of the information about a tournament. This would be changed later to allow non-admin users to access the functionalities with restrictions (issue #172 ). The related pull request is #188

MuyanXie:
Closing Statement:

In this issue, @jb-01 worked on discussing with the team about the necessity of having an update view. @MikeChen012345 and @MuyanXie worked on implementing these features in code

elizabethli31:
Issue Score: Excellent

Comments:
Nice job implementing this view! Also, great comments detailing the changes to be covered in future PRs.

Issue #87, opened 2023-10-30
Implement a `Delete` view for tournaments


MuyanXie:
Details:

We will implement the Delete in CRUD for our models in the views.py file, enabling the web server to accept authenticated and authorized users to delete a specific tournament.

Goals:

Upon the conclusion of the issue, we should have a robust views.py file with valid delete methods for the entities with correct relationships mapped between the entities.

Notes:

We should pay attention to give the delete right to only authenticated and authorized users

MikeChen012345:
We have implemented the delete view for the tournaments. Currently, only staff members are allowed to delete a tournament. This would be changed later to allow non-admin users to access the functionalities with restrictions (issue #172 ). The related pull request is #188

MuyanXie:
Closing Statement:

@MikeChen012345 and @jb-01 worked on coding the classes and functions for delete view. @MuyanXie worked on reviewing the structure of the code as well as checking everything is under the right track for CRUD pages

elizabethli31:
Issue Score: Excellent

Comments:
Same comment as to the other CRUD related PRs. Good work on splitting it up and fixing merge issues between the team.

PR #188, opened 2023-11-12
Add update and delete views to the tournaments webpages

MikeChen012345:
We have implemented the update and delete views for the tournaments (issues #151 , #87 ). Currently, only staff members are allowed to edit most of the information about a tournament and delete a tournament. This would be changed later to allow non-admin users to access the functionalities with restrictions (issue #172 ).


MikeChen012345:
Based on the suggestion of @elizabethli31 , we would create a new pull request to make the code better and cleaner. The current pull request is closed.

MikeChen012345:
The new pull request is #210

Issue #163, opened 2023-11-09
Create and show single-elimination brackets on the tournament-detail-view


MikeChen012345:
We want to let the users see the match brackets in the tournament. The number of brackets depends on the number of players, and if needed, some brackets may contain only one player (when there is an odd number of players in that round). When the tournament is created, all the players will be randomly assigned to a match with another player. Since this is for a single-elimination tournament, each time the tournament gets to the next round, the number of brackets is divided by the maximum number of players that can play the game (rounds up), and the match assignment for that round is randomly created based on the result of the last round.

MikeChen012345:
We have implemented the brackets feature. Currently, when a tournament is created, players in the tournament will be automatically randomly assigned to different matches (brackets). When one round of matches is done, a function is provided in tournaments_function.py to randomly assign winners of the last round to brackets for the next round. Note that the concept of a "round" is not implemented right now since games have not been implemented and hence the result of a match cannot be determined. We may add this "round" later in other pull requests. Also note that currently, the brackets are created as soon as the tournament is created. This is for testing purposes. However, after the concept of "registration deadline" is added, the brackets should be created after the registration deadline has passed so that users have time to sign up for and quit a tournament. The related pull request is #189 .

MuyanXie:
Close statement:

@MikeChen012345 worked on the main code logic for brackets
@MuyanXie worked on the logic of single elimination and discussed details about elimination together with brackets
@Riyush Helped with putting together the documentation and discussion details about the brackets and best practices

elizabethli31:
Issue Score: Excellent

Comments:
The functions related for this issue are a really nice addition for both the backend functionality and future renditions of frontend. My only comment is that this issue spanned two sprints, but great work providing a thorough explanation for the issue and detailed closing comment.

PR #189, opened 2023-11-12
Add tournaments brackets feature

MikeChen012345:
We have implemented the brackets feature mentioned in issue #163 . Currently, when a tournament is created, players in the tournament will be automatically randomly assigned to different matches (brackets). When one round of matches is done, a function is provided in tournaments_function.py to randomly assign winners of the last round to brackets for the next round. Note that the concept of a "round" is not implemented right now since games have not been implemented and hence the result of a match cannot be determined. We may add this "round" later in other pull requests. Also note that currently, the brackets are created as soon as the tournament is created. This is for testing purposes. However, after the concept of "registration deadline" is added, the brackets should be created after the registration deadline has passed so that users have time to sign up for and quit a tournament.


src/templates/tournaments/tournament_list.html
```diff
@@ -15,7 +15,6 @@ <h2>
         <p>Description: {{ tournament.description }}</p>
         <p>Rules: {{ tournament.rules }}</p>
         <p>Draw rules: {{ tournament.draw_rules }}</p>
-        <!-- Matches may be added later if needed -->
         <p>Winners:</p>
```
rqueiroz20: I think we should change how winners are displayed on ChiGame. Currently, if the tournament is in progress, the tournament has no winner, so why should the list view show "Winners" at all? I would recommend hiding that attribute/ changing what it says if there are no winners to something other than an empty space until the tournament has winners.

src/chigame/games/tournaments_function.py
rqueiroz20, line 1: The `create_tournaments_brackets()` function works well, but I'm wondering if we should include a restriction (at least for single-elimination tournaments) that the number of players in a tournament should be an even number/power of 2 so that every user is guaranteed to have an opponent in the first round. I was still able to create a tournament with an odd number of players, but they were not paired up with someone else. 

Also, the `next_round_tournament_brackets()` and `end_tournament()` functions don't seem to be used in this PR. Specifically, they don't appear to work since there is no `winners` attribute for matches/lobbies.

src/chigame/games/views.py
```diff
@@ -157,12 +158,29 @@ class TournamentCreateView(CreateView):
     # Note: we may remove the "matches" field later for the same reason,
     # but we keep it for now because it is convenient for testing.
 
+    # This method is called when valid form data has been POSTed. It
+    # overrides the default behavior of the CreateView class.
+    def form_valid(self, form):
+        response = super().form_valid(form)
+        create_tournaments_brackets(self.object)  # This should be changed later
```
rqueiroz20: I believe there should be a way to block someone from creating a tournament if the number of players in the tournament is greater than the maximum number of players in that tournament. I was able to create a tournament with `max_players` equal to 1 and add 2 users to create a match between them.

rqueiroz20 (commented):
In general, the functionality to create a tournament with randomized users works well, but there are some edge cases that I think are worth looking into. I've mentioned them in the individual reviews, but here is a summary of them:

- `next_round_tournament_brackets()` and `end_tournament()` don't work because of a missing attribute in the Match model
- There should be some restriction on creating a valid tournament (number of players in tournament should not be greater than max players in the tournament)
-  Change how winners are displayed if a tournament is in progress

Other than those comments, this is a good start towards implementing this feature! One final note: if a user leaves the tournament (caught cheating or some other issue), there should be some way to keep their opponent in the tournament. I was thinking either through advancing them to the next round automatically or creating a new match if that happens to multiple users.

src/chigame/games/tournaments_function.py
MikeChen012345: Hi Rafael, the create_tournaments_brackets() is designed to allow an odd number of players in a way that reflects how people deal with the situation in real life. People should not be rejected to join a tournament if they are at odds, or otherwise, they have to find someone else to join the tournament together to keep the number of players even. This would cause some trouble for the players. Currently, one player chosen at random can get the first round for free if there is an odd number of players, while other players would all be in pairs.

src/chigame/games/views.py
MikeChen012345: Hi Rafael, this is a good catch! It will be checked by the join_tournament functionality in a later pull request! Currently, the create_tournament_brackets feature only creates the brackets literally, and later we will implement a way for the users to utilize this feature.

src/templates/tournaments/tournament_list.html
MikeChen012345: That is a good idea! I will make a change on this.

src/chigame/games/tournaments_function.py
MikeChen012345: The two functions have also been updated now based on the feedback from the Match group. Please refer to my new commit for that.

rqueiroz20 (changes requested):
Overall, very nice job addressing the comments I made! I was able to successfully create a tournament with randomized matches. There is one that escaped my original review that I noticed upon re-review, but I'm not sure how important it will be:

- If I create a tournament with 8 users in it and decide to update the tournament to remove one of the users, that user's match still appears with their username in it. I'm not sure we want that to be the case.

Apart from that comment, I think this looks good to me!

MikeChen012345:
Hi Rafael, it seems that currently there is no way to remove a player from the match through the update view provided by Django. Therefore, we may try to use a different approach if we want to remove a player later.

rqueiroz20 (approved):
Very nice job addressing my suggestion about the display of tournament winners! I understand that we have come to an agreement about what to do about the issue I raised in my previous review regarding removing a player from a tournament and adjusting matches accordingly:

- Leave it for another PR/another tournaments team

Other than that, the functionality works! Nicely done!

MikeChen012345:
Thank you for your feedback! I will put this in my close comment as well.

src/chigame/games/tournaments_function.py
```diff
@@ -0,0 +1,132 @@
...
+            match_status=Lobby.Lobbied,
+            game=game,
+            game_mod_status=Lobby.Default_game,
+            created_by=players[i],  # the field is currently set to the first player in the list
+            min_players=game.min_players,
+            max_players=game.max_players,
+            # time_constraint, # default: 300
+            # lobby_created, # default: the field is set to the current time
+        )
+        players_in_match = players[i : i + tournament.game.max_players]  # the players in the match
```
elizabethli31: Seems like this line is adding all the players into the match. Shouldn't it just be one additional player besides the one in the loop? When I create the tournament, I get this: 
image
You could fix this by incrementing through the players list by 2 and creating matches that way. 
`for i in range(0,len(players),2)` and then `players_in_match = players[i:i+1]`.

elizabethli31 (changes requested):
I could be misunderstanding the functionality of your brackets functions, but I left a comment regarding what I am seeing when I create a tournament. Conceptually, it doesn't make sense to have 4 players against each other in the match.

MikeChen012345:
Hi Elizabeth,
The bracket functionality is currently built in a way such that it allows matches that have more than 2 players. The number of players that is assigned to each bracket (match) is determined by the maximum player of the game that the tournament uses. If you set the max_player field of the game to 2, then you will be able to have 1 vs 1 brackets. Hope this answers your question!

src/chigame/games/tournaments_function.py
```diff
@@ -0,0 +1,132 @@
...
+    for i in range(0, len(brackets), tournament.game.max_players):
+        game = tournament.game
+        lobby = Lobby.objects.create(
+            match_status=Lobby.Lobbied,
+            game=game,
+            game_mod_status=Lobby.Default_game,
+            created_by=brackets[i].winners.all()[0],  # the field is currently set to the first player in the list
+            min_players=game.min_players,
+            max_players=game.max_players,
+            # time_constraint, # default: 300
```
elizabethli31: Remove these lines. This doesn't need to be documented

MikeChen012345:
The functions in tournaments_function.py have now been moved into the tournament model in models.py

src/chigame/games/models.py
```diff
@@ -266,6 +268,126 @@ def __str__(self):  # may be changed later
...
+            lobby = Lobby.objects.create(
+                match_status=Lobby.Lobbied,
+                name=self.name + " " + str(i),  # the name of the lobby can be changed later
+                game=game,
+                game_mod_status=Lobby.Default_game,
+                created_by=players[i],  # the field is currently set to the first player in the list
+                # members, # this field is not set yet (it is unclear what it should be set to)
+                min_players=game.min_players,
+                max_players=game.max_players,
+                # time_constraint, # default: 300
```
elizabethli31: Can you remove these commented out lines? It isn't necessary to explain that these are set by default. Also, in general, a lot of the commented lines are unnecessary (like the ones explaining save, add, clear, etc.). Please remove those lines, and only keep ones that are absolutely necessary for explaining the logic of your code.

elizabethli31 (changes requested):
See comment. After you remove the commented out lines, I'll approve!

MikeChen012345:
Hi Elizabeth, the comments have been removed now

elizabethli31 (approved):
Thanks for removing the comments! Can you fix the linter test, and I'll merge?

MikeChen012345:
In this pull request, @MikeChen012345  and @MuyanXie worked on writing the code and testing the functionalities, and @Riyush focused on coming up with potential ways of improvement in displaying the brackets.
@rqueiroz20 helped a lot in finding issues related to the pull request and discussing if the issues could be resolved. Thank you!

elizabethli31:
Issue Score: Excellent

Comments:
Nice set of functions to provide added functionality for creating matches.

PR #193, opened 2023-11-12
Created a view for user's historical game information

### LINKED ISSUE #143: Add historical information view to display user's stats

AndrewM131:
In this PR related to issue #143, we implemented a URL path to a view called user_history.html that will display the historical information of a player. It will need to be revisited to actualize the information, but this is a setup for a future task.


24raniwalar (approved):
View, url, and template appropriately set up and extend from base. LGTM.

src/templates/users/user_history.html
```diff
@@ -0,0 +1,18 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h2>User Historical Information</h2>
+  <ul>
+    <li>Username: {{ user.name }}</li>
```
elizabethli31: Because there is a username field in the model (even though it's currently set to none), I would change this to Name instead of Username.

elizabethli31 (changes requested):
Good work! Left a minor comment that's more or less stylistic. The only other thing I had a question on is whether historical information of a user should be visible to all users? If yes, then once you pull from dev, I can approve and merge!

src/templates/users/user_history.html
AndrewM131: Historical information is intended to be visibile to all users. I just git pulled from dev, and will push with the style change you asked me to do.

elizabethli31 (approved):
LGTM! Good work and great job fixing the small change!

AndrewM131:
Closing Statement: We have successfully created a new view that will allow any user to see another user's historical information, but the actual implemented information is yet to be incorporated. This will be returned to so that we can properly display the information.

elizabethli31:
Issue Score: Excellent

Comments:
Nice job creating a framework for displaying historical information!

Issue #144, opened 2023-11-06
Add list of friends to user profile view


AndrewM131:
Summary:
When a user goes to their profile page, they can see a table of their friends' usernames. The goal of this issue is first to set up a table view of all of a user's friends so that in the future, new features can be implemented in this table, such as friend rankings for who has won the most number of tournaments, etc. If the user is an admin or the User instance does not have a corresponding UserProfile instance, then the friends table will not show up on the profile page.

AndrewM131:
Description: We want to implement a part of the profile view that allows users to see their list of friends.

AndrewM131:
Update: We created a PR and requested a review. After the first review, we'll need to make changes to the code, and re-request a review.

henrietta-k:
Update November 30: Fixed merge conflicts and changed the position of the if statement on user_detail.html so that the Friends table only shows up if the user has friends.

henrietta-k:
Update December 3: PR has been approved.

henrietta-k:
Closing statement: the issue is completed. When a user goes to their profile page, if the user is not logged in as an admin and they have a corresponding UserProfile, they are able to see a table of all their current friends. This feature will allow future features to be implemented, such as seeing which friend is most highly ranked, etc. All the assignees worked together to write the code for this but I was in charge of making the GitHub PR.

elizabethli31:
Issue Score: Excellent

Comments:
This issue took two sprints to complete since the PR wasn't closed until sprint 3. But I like that you explained the exact goals of what this issue was to accomplish and what that was to look like for different users.

PR #195, opened 2023-11-13
User profile displays a list of all friends

henrietta-k:
Addresses issue https://github.com/uchicago-cs/chigame/issues/144

When a user goes to their profile page, they can see a table of their friends' usernames. The goal of this issue is first to set up a table view of all of a user's friends so that in the future, new features can be implemented in this table, such as friend rankings for who has won the most number of tournaments, etc. If the user is an admin or the User instance does not have a corresponding UserProfile instance, then the friends table will not show up on the profile page.


henrietta-k:
Update: we met two times during the sprint to work on this issue. The first was on November 8th when we decided what the specific goal of this task was. We wanted a user to be able to see their friends so that when the tournament pages are implemented, users can see their friends' rankings, etc. We met again on November 12 to finalize the task and figure out an issue that we had, which was the scenario where a User did not have a corresponding UserProfile instance.

.DS_Store
omarelkashef, line 1: Maybe include .DS_Store in .gitignore so it does not get attached to any push.

src/.DS_Store
omarelkashef, line 1: same comment about .DS_Store

src/chigame/users/views.py
omarelkashef, line 1: the PR and issue description specify that the friends table is in the profile page but here it is implemented in the user list page. I think it makes more sense to have in the profile page as described in the issue

src/chigame/users/views.py
```diff
@@ -51,8 +52,16 @@ def get_redirect_url(self):
...
 def user_list(request):
     users = User.objects.all()
+    friends_table = None
 
-    return render(request, "users/user_detail.html", {"users": users})
+    user_profile = UserProfile.objects.filter(user=request.user).first()
+
+    if user_profile:
+        friends = user_profile.friends.all()
+        friends_table = FriendsTable(friends)  # Passing friends data directly to the table
```
omarelkashef: you are specifying the model in the FriendsTable as UserProfile but you pass in User because user_profile.friends is User

src/templates/users/user_detail.html
```diff
@@ -25,13 +25,21 @@ <h2>{{ object.name }}</h2>
         </div>
       </div>
       <!-- End Action buttons -->
+      {% if friends_table %}
```
omarelkashef: all of this if block is inside another   {% if object == request.user %} and I don't see any object passed into the context

omarelkashef (changes requested):

src/chigame/users/views.py
henrietta-k: The user list page is a profile page when `request.user` is not an admin. It only becomes a user-list page when an admin is trying to access it. In another PR, I changed the URL to make the name `user-list` less confusing but it hasn't been merged yet.

src/templates/users/user_detail.html
henrietta-k: This is because we only want the list of friends to show up when `object == request.user`. The list of friends should not show up when `object == request.admin`.

src/chigame/users/views.py
henrietta-k: This is the intended behaviour because UserProfile has a `friends` attribute while User does not. If the User does not have a corresponding UserProfile (1 - 1), the table will not show up because a User itself cannot have friends.

henrietta-k:
Update November 30: Fixed merge conflicts and changed the position of the if statement on user_detail.html so that the Friends table only shows up if the user has friends.

omarelkashef (approved):
Thank you for fixing the changes! I think it works as expected now. Good job!

src/chigame/users/views.py
```diff
@@ -52,25 +52,22 @@ def get_redirect_url(self):
...
-    table = UserTable(users)
-    context = {"users": users, "table": table}
-
-    # Add information about top ranking users, total points collected, etc.
-
-    return render(request, "users/user_list.html", context)
-
-
 @login_required
 def user_detail(request):
```
elizabethli31: I think all of this functionality needs to be in user_profile_detail_view and not user_detail. The issue and PR description state that the list of friends will be displayed under the user's profile page, but right now it's under the url at user-detail. I would scrap this view entirely and place all the code in the user_profile_detail_view (with the additional check that the profile the user is currently on is their own).

elizabethli31 (changes requested):
FriendsTable itself looks good to me, but the functionality of this PR is not as described. See my comment for more detail, but this table should be placed in the profile page. Also, it doesn't make sense to me for admin to not be able to have a FriendsTable as well. They should be able to see both their own friends and all users. I think you could keep the user_list view purely for admin to see all users, but the FriendsTable should be moved into user_profile_detail_view.

24raniwalar:
The new commit changes the functionality so that admins can view a page with all users with the capability for users to see their friends provided in #357

elizabethli31 (approved):
LGTM now! As an fyi this PR might end up just being a minor issue now since it only made the table with pretty much the same code as UserTable. Associated issues will still be major.

henrietta-k:
Closing statement: the issue is completed. When a user goes to their profile page, if the user is not logged in as an admin and they have a corresponding UserProfile, they are able to see a table of all their current friends. This feature will allow future features to be implemented, such as seeing which friend is most highly ranked, etc. All the assignees worked together to write the code for this but I was in charge of making the GitHub PR.

elizabethli31:
Issue Score: Excellent

Comments:
Great job creating the FriendsTable. Like I said in the review comment, I'm marking this as minor.

PR #197, opened 2023-11-13
Implemented path to user detail

### LINKED ISSUE #119: Implementing profile path

AndrewM131:
This PR addresses issue #119, and in the process of coding, we realized there were actually two sort of paths to a profile. One led to a view that allowed one to change their name and update email. Another led to a profile page that any user can access, and if the own user accesses this page, they can change their name, but if a separate user accesses this page, they can send a friend request. 
What we decided to do was implement a link in the nav bar that links to a user's personal information or a personal "profile" and will display the user's personal information. It's made such that only the user of that information can access it, and no one else can access this view. Before, the user_detail view can only be accessed by knowing the URL, but this way, a user can access it by clicking a link. Also, if a user goes to "My Profile" in the nav bar, after editing their name, they will be taken to this page, and will be able to update their name or email.


src/templates/users/user_detail.html
```diff
@@ -3,25 +3,43 @@
...
+              User Display Name: {{ object.name }}
+              <a class="btn btn-primary" href="{% url 'users:update' %}" role="button">Update Name</a>
+            </li>
+            <li class="list-item">
+              Email: {{ object.email }}
+              <a class="btn btn-primary"
+                 href="{% url 'account_email' %}"
+                 role="button">Update E-Mail</a>
+            </li>
+            <li class="list-item">Password: {{ object.password }}</li>
```
omarelkashef, line 36: it might be a good idea to not show the password and instead have a Change Password button that takes you to change password view that can be implemented in future issues

src/templates/users/user_detail.html
```diff
@@ -3,25 +3,43 @@
...
+            <li class="list-item">Bio: {{ object.bio }}</li>
```
omarelkashef: User does not have bio attribute. UserProfile is the one that stores the bio

src/templates/users/user_detail.html
omarelkashef: it might be a good idea to not show the password and instead have a Change Password button that takes you to change password view that can be implemented in future issues

src/templates/users/user_detail.html
```diff
@@ -3,25 +3,43 @@
...
+            <li class="list-item">Date Joined: {{ object.date_joined }}</li>
+            <li class="list-item">Is Admin: {{ object.is_staff }}</li>
+            <li class="list-item">Friends: {{ object.friends }}</li>
```
omarelkashef: User does not have friends either. friends is in UserProfile

src/templates/users/user_detail.html
```diff
@@ -3,25 +3,43 @@
...
 {% endblock title %}
 {% block content %}
   <div class="container">
     <div class="row">
       <div class="col-sm-12">
-        <h2>{{ object.name }}</h2>
+        <h2>My Details:</h2>
       </div>
     </div>
     {% if object == request.user %}
```
omarelkashef, line 15: currently a user can use the url and get to the page but they won't see personal info which is great but if you want to prevent users from accessing others personal info pages, you can use  UserPassesTestMixin in the class-based view where the test_func would be something like

`
def test_func(self):
        return self.request.user == self.get_object()
`

https://docs.djangoproject.com/en/4.2/topics/auth/default/#django.contrib.auth.mixins.UserPassesTestMixin

also if you want to control what happens to "illegal" users, you can use AccessMixin and change the get_permission_denied_message()

https://docs.djangoproject.com/en/4.2/topics/auth/default/#django.contrib.auth.mixins.AccessMixin

src/templates/users/user_detail.html
```diff
@@ -3,25 +3,43 @@
...
-  User:
-  {{ object.name }}
+  My Details:
 {% endblock title %}
 {% block content %}
   <div class="container">
     <div class="row">
       <div class="col-sm-12">
-        <h2>{{ object.name }}</h2>
+        <h2>My Details:</h2>
```
omarelkashef: I think My info and My profile will be confusing to the user. Can you explain the difference between them?

omarelkashef (changes requested):

src/templates/users/user_detail.html
AndrewM131: I will update it so that it reflects UserProfile

src/templates/users/user_detail.html
AndrewM131: You're right, I'll implement that feature.

src/templates/users/user_detail.html
AndrewM131: I'll update it so it accesses UserProfile

src/templates/users/user_detail.html
AndrewM131: So my info will just display all their information, and can only be accessed that user. It's not like a profile where any user can access it.

src/templates/users/user_detail.html
AndrewM131: Thank you for the idea, I just implemented it, and it works as expected.

AndrewM131: Hey, so after pushing my code, I'm failing one test in which the test case TestUserDetailView.test_authenticated is failing due to a PermissionDenied exception being raised. This happens when the UserPassesTestMixin in my UserDetailView denies permission for the test user, which it's doing properly, but since the test is formatted this way, I can't implement this properly.

src/templates/users/user_detail.html
omarelkashef: ok got it!

src/templates/users/user_detail.html
```diff
@@ -3,25 +3,48 @@
...
+                 href="{% url 'account_email' %}"
+                 role="button">Update E-Mail</a>
+            </li>
+            <li class="list-item">
+              Password:
+              <a class="btn btn-primary"
+                 href="{% url 'users:password_reset' %}"
+                 role="button">Reset Password</a>
+            </li>
+            <li class="list-item">Bio: {{ object.UserProfile.bio }}</li>
```
omarelkashef: This needs to be object.userprofile.bio

This goes the same for all references to the user profile. It should be lowercase.

src/templates/users/user_detail.html
```diff
@@ -3,25 +3,48 @@
...
+            <li class="list-item">Date Joined: {{ object.date_joined }}</li>
+            <li class="list-item">Is Admin: {{ object.is_staff }}</li>
+            <li class="list-item">Friends: {{ object.UserProfile.friends }}</li>
```
omarelkashef: same issue with UserProfile. Just mentioning it again since it is a little far from the first one so you can notice it

src/chigame/users/urls.py
```diff
@@ -21,4 +22,8 @@
     path("cancel_friend_invitation/<int:pk>", view=cancel_friend_invitation, name="cancel-friend-invitation"),
     path("user-list/", views.user_list, name="user_list"),
     path("user_history/<int:pk>", views.user_history, name="user-history"),
+    path("password_reset/", auth_views.PasswordResetView.as_view(), name="password_reset"),
+    path("password_reset/done/", auth_views.PasswordResetDoneView.as_view(), name="password_reset_done"),
+    path("reset/<uidb64>/<token>/", auth_views.PasswordResetConfirmView.as_view(), name="password_reset_confirm"),
+    path("reset/done/", auth_views.PasswordResetCompleteView.as_view(), name="password_reset_complete"),
```
omarelkashef: I am actually getting the following error when i try to reset my password

Reverse for 'password_reset_confirm' not found. 'password_reset_confirm' is not a valid view function or pattern name.

I followed django documentation here https://docs.djangoproject.com/en/4.2/topics/auth/default/#all-authentication-views 

and added the following 
`    path("accounts/", include("django.contrib.auth.urls")),
`
to src/config/urls.py and it seems to work. The line above automatically adds the necessary paths for password resetting. Just note that I changeed accounts/ to account/ because there is already accounts/ path. I was not sure what else to name it.

src/templates/users/user_detail.html
omarelkashef: Which test is it failing? If the test is looking for access to a page that it should not have access to, maybe we should actually change the test.

omarelkashef (changes requested):

src/templates/users/user_detail.html
AndrewM131: Just addressed, thank you for the input!

src/templates/users/user_detail.html
AndrewM131: Just addressed, thank you for the input!

src/chigame/users/urls.py
AndrewM131: Just implemented, thank you for finding me the resource, I appreciate that a lot.

AndrewM131:
From what I understood, this is the test case:
def test_authenticated(self, user: User, rf: RequestFactory):
    request = rf.get("/fake-url/")
    request.user = UserFactory()
    response = user_detail_view(request, pk=user.pk)
But since this user is not the same as the user we are trying to view, the test_func will return False, and the PermissionDenied exception will be raised.
If we were to change the test case, wouldn't we have to talk to Borja about it since he implemented it?

src/chigame/users/urls.py
omarelkashef: it works now! but I don't think you need lines 29-32 if you just replace 
`  href="{% url 'users:password_reset' %}"` with 
`href="{% url 'password_reset' %}`
because the latter will use the 
`  path("account/", include("django.contrib.auth.urls")),`
that you added

omarelkashef (changes requested):
Great! It work. Just a minor fix.

omarelkashef:
> From what I understood, this is the test case: def test_authenticated(self, user: User, rf: RequestFactory): request = rf.get("/fake-url/") request.user = UserFactory() response = user_detail_view(request, pk=user.pk) But since this user is not the same as the user we are trying to view, the test_func will return False, and the PermissionDenied exception will be raised. If we were to change the test case, wouldn't we have to talk to Borja about it since he implemented it?

oh ok I thought we added that test. But still maybe we need to fix that test but that should be another issue.

src/templates/users/user_detail.html
```diff
@@ -3,25 +3,48 @@
...
+            <li class="list-item">
+              Password:
+              <a class="btn btn-primary"
+                 href="{% url 'users:password_reset' %}"
+                 role="button">Reset Password</a>
+            </li>
+            <li class="list-item">Bio: {{ object.userprofile.bio }}</li>
+            <li class="list-item">Date Joined: {{ object.date_joined }}</li>
+            <li class="list-item">Is Admin: {{ object.is_staff }}</li>
+            <li class="list-item">Friends: {{ object.userprofile.friends }}</li>
```
elizabethli31: If this field is none, it should just be blank. I'm currently getting "users.User.None".

src/chigame/users/urls.py
```diff
@@ -26,4 +27,8 @@
     path("accept_friend_invitation/<int:pk>", view=accept_friend_invitation, name="accept-friend-invitation"),
     path("decline_friend_invitation/<int:pk>", view=decline_friend_invitation, name="decline-friend-invitation"),
     path("user_history/<int:pk>", views.user_history, name="user-history"),
+    path("password_reset/", auth_views.PasswordResetView.as_view(), name="password_reset"),
```
elizabethli31: I can't seem to reset my password for any test accounts (i.e. user1@example.com) because it brings me to a fake page. This probably has something to do with the Django auth views, but do you know if there's any way to fix this?

elizabethli31 (changes requested):
Left a few comments with the primary one being that I can't actually reset my password with a fake email. Also, it's a bit redundant to have separate "my info" and "my profile" pages. Intuitively, I think it makes more sense to just have the "my info" details under the "my profile" page only if the user is logged in and on their own profile page. Let me know what you think.

AndrewM131:
Update: removed nav link since it'll be more confusing and merged conflicts.

AndrewM131:
Update: I addressed the review comments by removing the page from nav bar, and it can now be accessed through user profile. I removed the password option as the django built in doesn't really have a viable reset password unless you're in the admin view page and can manually change it. I also updated the view that if a user isn't an admin, it doesn't say anything about being an admin, and it now properly displays a list of friends and if a user has no friends, it says "no friends yet". The path should also be fixed to path("account/", include("django.contrib.auth.urls")).

omarelkashef (changes requested):
the friends list working well along with changing the email. But when i try to view someone else's details I get a blank page without any message that my request is forbidden because it is not my account.

AndrewM131:
Good catch, I should address that in my code. I just implemented it, and should work now.

omarelkashef (approved):
Looks very clean now. Great Job, Andrew!

src/templates/users/user_detail.html
```diff
@@ -3,45 +3,56 @@
...
-          <!-- Admin-specific content goes here -->
           <ul>
-            {% for user in users %}
-              <li>
-                {{ user.username }}
-                {{ user.email }}
-              </li>
-            {% endfor %}
+            <li class="list-item">
+              User Display Name: {{ object.name }}
```
elizabethli31: It's a little unclear to have this as display name because that's a different field for UserProfile. I would change this to just show "Name:"

src/templates/users/user_detail.html
```diff
@@ -3,45 +3,56 @@
...
+            <li class="list-item">
+              Email: {{ object.email }}
+              <a class="btn btn-primary"
+                 href="{% url 'account_email' %}"
+                 role="button">Update E-Mail</a>
+            </li>
+            <li class="list-item">Bio: {{ object.userprofile.bio }}</li>
+            <li class="list-item">Date Joined: {{ object.date_joined }}</li>
+            <li class="list-item">
+              Friends:
```
elizabethli31: Remove all of this friend stuff, as it is implemented in PR #357.

elizabethli31 (changes requested):
Really nice work on this Andrew, and thanks Omar for the very thorough review!! Great error handling, and the additional work on this PR since it was opened clearly shows. From your most recent comment, it sounds like the admin should still be marked as "admin", but I don't see that reflected in the code. See my comments for a couple minor changes.

src/templates/users/user_detail.html
AndrewM131: Got it, will do that in the next commit!

src/templates/users/user_detail.html
AndrewM131: You're correct, I'll change that in the user_detail.html

AndrewM131:
According to my comment, I should have the title of admin in the user_detail.html, so I implemented it in user_detail.html, and after the title, I wrote "Admin of Chigame" in bold to emphasize. This is only possible if the user is an admin.

elizabethli31 (approved):
Thanks for addressing my changes - LGTM!

AndrewM131:
Closing comment: At the start, the goal was to create a link in nav bar, and have personal information displayed that only the user of that information can access. That goal changed because it would be too confusing for a user to have a profile and account in the nav bar, so the link for account can be accessed through the user profile. I also added information in the account view page, such as name, a link to changing name, email, a link to changing email, bio, and it highlights if a user is an admin of chigame.

elizabethli31:
Issue Score: Excellent

Comments:
Nice work!

PR #199, opened 2023-11-13
Expanded actions for friends to allow for invitations to games, invitations to tournaments

### LINKED ISSUE #194: Actions for friends -> invite to game and tournaments
Description: This issue will address creating more actions for friends within the chigame app. It will work to allow friends to be able to invite other friends to games and tournaments, as well as allow friends to remove friends.

ishiii-s:
This PR involves removing friends, inviting them to games, and inviting them to tournaments. These functionalities would only be accessible if a user is logged in. Next steps would be to create functionality to accept/decline these requests, and incorporate the notification framework.


AndrewM131:
When I try to run your code using "python3 manage.py runserver", I get the following error "ImportError: cannot import name 'Group' from partially initialized module 'chigame.users.models' (most likely due to a circular import) (/root/chigame/src/chigame/users/models.py)"
From what I understand, this is when "Circular imports can occur when two or more modules mutually depend on each other." You'll have to fix this in order for the code to be properly reviewed, and this may also be the cause of failing the checks.

src/chigame/users/views.py
```diff
@@ -120,3 +120,31 @@ def cancel_friend_invitation(request, pk):
...
+    sender_profile = sender.profile
+    sender_profile.friends.remove(receiver)
+    messages.success(request, "Friend removed successfully.")
+    return redirect(reverse("users:user-profile", kwargs={"pk": request.user.pk}))
+
+@login_required
+def invite_to_game(request, pk, match_id):
+    sender = User.objects.get(pk=request.user.id)
+    receiver = User.objects.get(pk=pk)
+    match = User.objects.get(pk=match_id)
```
AndrewM131: when defining match, I think you defined it incorrectly, and it should be Match.objects instead of User.objects, as the User model doesn't have a match_id

src/chigame/users/views.py
```diff
@@ -120,3 +120,31 @@ def cancel_friend_invitation(request, pk):
...
+    match = User.objects.get(pk=match_id)  
+    GameInvitation.objects.create(sender=sender, receiver=receiver, group=group, match=match)
+    messages.success(request, "Invitation to game sent successfully.")
+    return redirect(reverse("users:user-profile", kwargs={"pk": request.user.pk}))
+
+@login_required
+def invite_to_tournament(request, pk, tournament_id):
+    sender = User.objects.get(pk=request.user.id)
+    receiver = User.objects.get(pk=pk)
+    tournament = User.objects.get(pk=tournament_id)
```
AndrewM131: Same thing here where it should be Tournament.obkects instead of User.objects , as the User model doesn't have a tournament_id

src/chigame/users/views.py
```diff
@@ -120,3 +120,31 @@ def cancel_friend_invitation(request, pk):
...
+    sender_profile.friends.remove(receiver)
+    messages.success(request, "Friend removed successfully.")
+    return redirect(reverse("users:user-profile", kwargs={"pk": request.user.pk}))
+
+@login_required
+def invite_to_game(request, pk, match_id):
+    sender = User.objects.get(pk=request.user.id)
+    receiver = User.objects.get(pk=pk)
+    match = User.objects.get(pk=match_id)  
+    GameInvitation.objects.create(sender=sender, receiver=receiver, group=group, match=match)
```
AndrewM131: When did you define group in the function? Currently it isn't defined, and may be why you aren't passing the checks.

AndrewM131 (changes requested):
Make sure you address the circular import issue, and pass the checks.

ishiii-s:
Reviewed change requests and modified code in views.py accordingly to properly define match and tournament. Also made changes to imports in models.py to account for circular dependencies.

src/chigame/users/views.py
```diff
@@ -118,7 +120,9 @@ def send_friend_invitation(request, pk):
     if new:
         messages.success(request, "Friendship invitation sent successfully.")
         notification = Notification.objects.create(
-            actor=invitation, receiver=other_user, type=Notification.FRIEND_REQUEST
+            actor=invitation,
+            receiver=other_user,
+            type=Notification.FRIEND_REQUEST,
```
AndrewM131: I like that you made the style cleaner and more readable

src/chigame/api/tests/factories.py
```diff
@@ -90,12 +90,15 @@ class Meta:
...
-    max_players = 16
+    registration_start_date = Faker("date_this_year")
+    registration_end_date = Faker("date_this_year")
+    tournament_start_date = Faker("date_this_year")
+    tournament_end_date = Faker("date_this_year")
+    max_players = Faker("pyint", min_value=1, max_value=1000)
     description = Faker("text")
     rules = Faker("text")
     draw_rules = Faker("text")
+    num_winner = Faker("pyint", min_value=1, max_value=1000)
```
AndrewM131: When creating a new tournament, the website view field is called "Num winner," you should have a better name for it so it's a bit clearer what you're asking when creating the tournament. It's a good name for coding, but it should be changed for the user view on the website.

src/chigame/games/models.py
```diff
@@ -232,21 +232,118 @@ class Tournament(models.Model):
...
-    end_date = models.DateTimeField()
+    registration_start_date = models.DateTimeField()  # the start date of the registration period
+    registration_end_date = models.DateTimeField()  # the end date of the registration period
+    tournament_start_date = models.DateTimeField()  # the start date of the tournament
+    tournament_end_date = models.DateTimeField()  # the end date of the tournament
     max_players = models.PositiveIntegerField()
     description = models.TextField()  # not limited to 255 characters
     rules = models.TextField()  # not limited to 255 characters
     draw_rules = models.TextField()  # not limited to 255 characters
+    num_winner = models.PositiveIntegerField(default=1)  # number of possible winners for the tournament
```
AndrewM131: Good style that it has a default of 1 for better user usability

src/chigame/games/models.py
```diff
@@ -232,21 +232,118 @@ class Tournament(models.Model):
...
+        elif (
+            self.tournament_start_date > timezone.now()
+        ):  # the registration period has ended but the tournament has not started yet
+            return "registration closed"
+        elif self.tournament_end_date > timezone.now():  # the tournament has started but not ended yet
+            return "tournament in progress"
+        else:  # the tournament has ended
+            return "tournament ended"
+
+    def clean(self):  # restriction
```
AndrewM131: Good error checking!

AndrewM131 (changes requested):
I left a few comments, but overall the invitation to game and invitation to tournaments work perfectly, it's really awesome. 
One comment:
-When a tournament has ended, it shouldn't give the option to join or withdraw the tournament

ishiii-s:
Update: Rishabh removed "remove friend" functionality due to that now being addressed in a separate PR/Issue. Merge conflicts have been resolved and checks have passed.

src/chigame/api/tests/factories.py
ishiii-s: I don't think these are changes handled by us. The only files we changed are src/chigame/urls.py, src/chigame/models.py, and src/chigame/views.py.

AndrewM131 (approved):
Clarification was given over slack that num winner and the buttons for join and withdraw are not incorporated in this PR and deals with tournaments. The code LGTM!

src/chigame/users/urls.py
```diff
@@ -31,8 +32,10 @@
     path("add_friend/<int:pk>", view=send_friend_invitation, name="add-friend"),
     path("remove_friend/<int:pk>", view=remove_friend, name="remove-friend"),
     path("cancel_friend_invitation/<int:pk>", view=cancel_friend_invitation, name="cancel-friend-invitation"),
+    path("invite_to_game/<int:pk>", view=invite_to_game, name="invite-to-game"),
```
elizabethli31: I don't think these urls are correct because they don't have the pk for the match which is needed for the view

src/chigame/users/views.py
```diff
@@ -185,6 +185,25 @@ def cancel_friend_invitation(request, pk):
 
 
 @login_required
+def invite_to_game(request, pk, match_id):
+    sender = User.objects.get(pk=request.user.id)
+    receiver = User.objects.get(pk=pk)
```
elizabethli31: Everything should be in a try statement because there is no error handling for if the receiver or match don't exist. Same goes for tournament

elizabethli31 (changes requested):
Nice addition for invitations, but there are some errors in the views and urls. Please test with a match and tournament instance

henrietta-k:
Update: got rid of GameInvitation but kept TournamentInvitation. Tested it and the URL works as expected.

src/chigame/users/views.py
```diff
@@ -185,6 +185,18 @@ def cancel_friend_invitation(request, pk):
 
 
 @login_required
+def invite_to_tournament(request, pk, tournament_pk):
+    try:
+        sender = get_object_or_404(User, pk=request.user.id)
```
elizabethli31, line 190: Need to do one more check where sender and receiver are not the same. You shouldn't be able to send an invite to yourself.

src/chigame/users/views.py
```diff
@@ -185,6 +185,18 @@ def cancel_friend_invitation(request, pk):
...
+        receiver = get_object_or_404(User, pk=pk)
+        tournament = get_object_or_404(Tournament, pk=tournament_pk)
+        TournamentInvitation.objects.create(sender=sender, receiver=receiver, tournament=tournament)
```
elizabethli31, line 193: Along with the check on user id, the invitation object should only be created if that check succeeds.

elizabethli31 (changes requested):
A few minor changes, but generally looks good!

PR #203, opened 2023-11-14
Delete notification model and add announcemnt model [V2]

### LINKED ISSUE #162: Change Notification model into Announcement
During our group meeting on Monday, Rafael contacted the User Profile group and found that the group has created a Notification model for sending notifications. Since we don't want overlapping here, we would like to change the Notification model in our games/models.py into an Announcement model, which is used to send announcements to multiple users by sending a Notification to each user in the "recipients" attribute of the Announcement model.

MuyanXie:
PR description for Issue #162 

Overview:

We tried to the models to use instead an announcement model
We imported and used the notification model from the User's team
We provided a method to send notifications by auto-generating notifications to the correct receiver

Key changes:

deleted the notification model in games views
added announcement model for games views using notification from user views
merged discrepancies in database scheme


src/chigame/games/models.py
```diff
@@ -223,24 +224,57 @@ def __str__(self):  # may be changed later
...
-    recipients = models.ManyToManyField(User, related_name="notifications")
+    REMINDER = 2
+    UPCOMING_MATCH = 3
+    MATCH_PROPOSAL = 4
+
+    ANNOUNCEMENT_TYPES = (
+        (REMINDER, "REMINDER"),
+        (UPCOMING_MATCH, "UPCOMING_MATCH"),
+        (MATCH_PROPOSAL, "MATCH_PROPOSAL"),
+    )
```
rqueiroz20, line 240: Do we want a notification type for letting users know it's time to join a tournament? Or are we putting that into the upcoming match notification?

rqueiroz20 (commented):
Everything looks good to me! I was able to successfully run the migrations and start up the website. Other than the tournament type question I asked, I think it is ready to be merged.

src/chigame/games/models.py
MuyanXie: We can put that in the upcoming_match notification type. We only have three types here as we are actually basing our announcement model on the notification model from users/model.py, which only provides these options and one inapplicable friend request option.

rqueiroz20 (approved):
Alright, everything looks good now! Glad we resolved the notification question I had. I was able to successfully run the migrations and load up the website. Good job!

elizabethli31 (approved):
This looks really good! Tested and everything works perfectly. Great work Sam and Mike! Update with dev and I'll merge!

elizabethli31 (approved):
Looks good and thanks for fixing migrations!

MuyanXie:
What each person did in this PR:

1. Sam Xie: wrote the code to import notification from users/models.py and make everything integrated
2. Mike Chen: wrote the logic to use different fields from notifications built by user's team
3. Rafael Queiroz: reviewed the code and provided constructive feedbacks

elizabethli31:
Issue Score: Excellent

Comments:
Change to announcement makes much more sense, and helper functions will prove useful once announcements are implemented. Nice work!

PR #205, opened 2023-11-14
Add create and read views to the tournaments webpages

MikeChen012345:
We have implemented the create view, list view, and detail view for the tournaments (issues https://github.com/uchicago-cs/chigame/issues/88 , https://github.com/uchicago-cs/chigame/issues/86 , https://github.com/uchicago-cs/chigame/issues/107 , and https://github.com/uchicago-cs/chigame/issues/105 ). On the create view, users are able to create a tournament and assign values to most of the attributes (the link to the create view is located on the list view and only appears when the user is a staff member). Currently, we only allow staff to create tournaments, which may be changed later with restrictions. For the list view and detail view, staff, non-admin users, and guests (not logged in) are all able to see the information related to the tournament. Note that we have merged issues https://github.com/uchicago-cs/chigame/issues/107 and https://github.com/uchicago-cs/chigame/issues/105 because currently, we recognize no need to create separate detail views for spectators and players, so only one detail view is created.


src/templates/tournaments/tournament_create.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Create/Update Tournament</h1>
```
rqueiroz20: I would suggest changing this to just "Create Tournament" since this PR is not allowing us to edit any tournament information.

src/templates/tournaments/tournament_create.html
rqueiroz20, line 1: There is something weird going on with the Matches attribute for creating a tournament. I'm able to click that vertical line, making it go away like a checkbox. Is there any way we can fix this, or is this just a result of Matches being a Many-to-Many foreign key?
Screenshot 2023-11-13 at 8 06 47 PM

rqueiroz20 (changes requested):
Creating and viewing tournaments looks very good! I'm able to access information about a specific tournament, create one, and view the created one's information. A few general questions to consider:

1. Are we going to include detailed information about the tournament in the list view? So far, it looks like the information is just copy-pasted in both the list view and the detailed view, which I'm not sure we're looking for.
2. I mentioned it in the comments, but there is something weird when creating a tournament about initializing the Matches attribute. It doesn't affect my ability to create tournaments, but I'm not sure it looks good for people creating tournaments. Also, this would mean the person who creates the tournament would have to decide who is paired up with who, which I believe is not what we want.

Other than those details, your code looks good! Let me know when these changes have been made (or if we decide not to change them right now).

src/chigame/games/views.py
```diff
@@ -68,3 +72,57 @@ def get_context_data(self, **kwargs):
...
+        "name",
+        "game",
+        "start_date",
+        "end_date",
+        "max_players",
+        "description",
+        "rules",
+        "draw_rules",
+        "matches",
+        "num_winner",
```
rqueiroz20: Remove this field as it is causing an error: 
django.core.exceptions.FieldError: Unknown field(s) (num_winner) specified for Tournament

rqueiroz20 (changes requested):
I am unable to access the create tournament page. I get the following error when doing so: 

django.core.exceptions.FieldError: Unknown field(s) (num_winner) specified for Tournament

Fix that and then I will re-review your code to make sure the functionality works as intended.

src/chigame/games/views.py
MikeChen012345: Sorry, I forgot to remove the field here

rqueiroz20 (changes requested):
I am unable to create a tournament still. I get the following error: 

NOT NULL constraint failed: games_tournament.num_winner

I've looked through the code from this PR, but it doesn't look like there's an attribute for num_winner anywhere in the Tournaments model. So I am uncertain what is causing this error. If you could resolve this, that would be much appreciated.

rqueiroz20 (approved):
The Create/Read views both work for tournaments of my end! Good job with making the HTML change I requested! Nice work!

src/templates/tournaments/tournament_detail.html
```diff
@@ -0,0 +1,70 @@
...
+  <p>Draw rules: {{ tournament.draw_rules }}</p>
+  <p>Players:</p>
+  <ul>
+    {% for player in tournament.get_all_players %}
+      <li>
+        <p>
+          {% if player.name %}
+            <a href="{% url 'users:detail' player.id %}">{{ player.name }}</a>
+          {% else %}
+            <a href="{% url 'users:detail' player.id %}">{{ player.email }}</a>
```
elizabethli31, line 20: Possibly a data breach because email is visible to anyone and not just staff. Also there might be concerns with users who have made their accounts through API and don't have associated users. I'm okay with this for now, but what do you guys think? (You could also just leave this as something for future iterations of the class to deal with).

elizabethli31 (changes requested):
Generally looks good! Nice checking of possibilities in the html pages. In a future PR, it'd be nice to have a create button show up on the tournaments list page. Left a minor question. Can you also remove the changes in migration pages and update with dev?

src/templates/tournaments/tournament_detail.html
MikeChen012345: Hi Elizabeth,
I think the change can be dealt with later because currently the user only need to provide an email when they create an account on the website (which means that the username would just be blank). As a result, if we do not have email displayed here, and the users have not setup their usernames, the way that a user is displayed would be a bunch of blank spaces, which is not ideal.

MikeChen012345: Thank you for the review! The branch has been updated. For the changes in migration pages, I am not sure what causes the difference, but they should have the same content?

elizabethli31: Okay I'm okay with that decision! For the migration pages, I'm not really sure what's going on or why they are different, but could you just remove them from the PR adds?

elizabethli31 (changes requested):
Still seeing the migration changes. Could you try following one of these and make sure they are not part of the commit? https://stackoverflow.com/questions/38743912/remove-a-file-from-a-git-pull-request
Once you do that and rebase, I'll approve and merge!

elizabethli31 (approved):
LGTM and great work!

MikeChen012345:
In this pull request, we have implemented a basic create and detail view for the tournament model. During the process,
- Mike, Sam, and Joshua worked on the code. In particular, Mike and Sam mainly worked on the implementation of the views and HTML templates, while Joshua contributed a lot to the discussion of implementation details and mainly worked on giving feedback and suggestions from the List view that he was working on. 
- Rafael did the review. It was a pull request that took a long time, and he surely spent a lot of time reviewing the code. The effort is greatly appreciated!

elizabethli31:
Issue Score: Excellent

Comments:
Great work! Thanks for breaking this down into multiple PRs. I know it was a lot of work to refactor and make those changes, but you guys did a really wonderful job. Also, like mentioned on the other CRUD PR, make sure for future PRs, to only open them, once you are sure it's ready to be pushed, so we can avoid a ton of unnecessary commits.

PR #236, opened 2023-11-15
Created categories and mechanic fixtures

frowenz:
Corresponding issue: #235

## Things that were done:
1. Updated `category` to `categories` in the `Games` model to be consistent with the `mechanics` attribute. Also, updated this in corresponding places
2. Added an image url category to `Category` and `Mechanic` models
3. Created a web scraper viewable [here](https://github.com/uchicago-cs/chigame/wiki/Games-~-Scraping-Categories-and-Mechanics-from-BGG) to create fixture files for `Category` and `Mechanic` models
4. Commit the output of the code, `/src/chigame/games/fixtures/mechanics_categories_fixtures.json`

Screenshot 2023-11-15 at 5 11 11 PM

## Future work:
1. Parsing the Category and Mechanic descriptions from BGG more intelligently
2. Better Category and Mechanic selection in Game Create page
3. Adding in Category and Mechanic Details page that lists corresponding Games
4. Adding in autofill support for Categories and Mechanics

PR #239, opened 2023-11-16
Added support to search bar for querying for games by the publisher's name, by an artist's name, or by a designer's name

### LINKED ISSUE #170: Add support to games search bar to allow users to search games by author as well as title
Currently, the game search bar can only search by the game name or by category names. We will add functionality to the game search bar, so that users can search the games by author. This involves editing the filter options in `games/views.py`, so that the games can be queried by the publisher's name or by the artists and designers

ptlin25:
The search results for querying for games has been altered. Now, when a user enters a query, along with games whose name matches the query and games whose category matches the query, games whose publisher's name or games who have an artist or designer whose name matches the query are shown in the search results. This was accomplished by adding more Q objects to the object filer. More information about Q objects can be found at https://docs.djangoproject.com/en/4.2/topics/db/queries/#complex-lookups-with-q-objects


src/chigame/games/views.py
```diff
@@ -91,6 +91,26 @@ def test_func(self):
...
+
+    """
+    The Q object is an object used to encapsulate a collection of keyword
+    arguments that can be combined with logical operators (&, |, ~) which
+    allows for more advanced searches. More info can be found here at
+    https://docs.djangoproject.com/en/4.2/topics/db/queries/#complex-lookups-with-q-objects
+    """
+    object_list = Game.objects.filter(
+        Q(name__icontains=query)
+        | Q(category__name__icontains=query)
```
willcravitz: This might need an update after PR #236 when "category" becomes "categories".

willcravitz: Did you also think about adding search support for mechanics?

willcravitz (approved):
Here are quick checks for each type of query. The results are based on the small 5 game fixture.

**Name**
search input: "ride"
results: Ticket to Ride

**Category**
search input: "strategy"
results: Settlers of Catan, Ticket to Ride, Pandemic, Scythe

**Publisher**
search input: "kosmos"
results: Settlers of Catan

**People**
search input: "Leacock"
results: Pandemic

Everything functions as expected. This seems like a simple and worthwhile update. I mentioned that it might also make sense to add support for "mechanics", but I'll leave that up to you.

ptlin25:
I agree that the search bar should support searching by mechanics, but I think that should be left for a separate issue and pull request.

src/chigame/games/views.py
majorsylvie: #236 has been merged!

majorsylvie (changes requested):
As Will mentioned, since #236 has been merged, search needs to be updated to use `categories` instead of `category`. 

Please do so!

ptlin25:
In `src/chigame/games/views.py`, search has been updated to use `categories` instead of `category`

src/chigame/games/views.py
```diff
@@ -91,6 +91,26 @@ def test_func(self):
...
+    The Q object is an object used to encapsulate a collection of keyword
+    arguments that can be combined with logical operators (&, |, ~) which
+    allows for more advanced searches. More info can be found here at
+    https://docs.djangoproject.com/en/4.2/topics/db/queries/#complex-lookups-with-q-objects
+    """
+    object_list = Game.objects.filter(
+        Q(name__icontains=query)
+        | Q(categories__name__icontains=query)
+        | Q(people__name__icontains=query)
+        | Q(publishers__name__icontains=query)
```
majorsylvie, line 107: Appreciate the syntax here to separate each query component onto a new line with the operation at the front.

Both of these decisions drastically improve readability and ease of extension.

src/chigame/games/views.py
```diff
@@ -123,10 +143,6 @@ def get_context_data(self, **kwargs):
         # Additional context can be added if needed
         return context
 
-    # check if user is staff member
-    def test_func(self):
-        return self.request.user.is_staff
-
```
majorsylvie, line 129: Why was this helper removed?

majorsylvie (changes requested):
Code looks good! 
Appreciate your readable syntax.

FYI there is a merge conflict with `views.py` on dev with the old version of your `search_results()` code.

Please resolve the merge conflict, _verify your code works as intended_, then rerequest review and lets get this merged!

src/chigame/games/views.py
ptlin25: This helper was removed because it was an error caused from a merge. Somehow with one of the merges, this helper got duplicated and pushed from `GameEditView` to `TournamentListView`.

majorsylvie (approved):
This code works and is ready to merge.

However, there should definitely be updates to how the games search results are displayed as a result of how this is coded.

Since you do substring and partial matches, consider the following situation:

you have multiple games and multiple people working on each game. 
I created a set of games and users to simulate this:

image

if I search for "selin" or "se", then just games 1 and 4 show up (as expected)
image

However, both "Tom" and "Tom York" have worked on multiple games, if I look up `York` or `Yo, then I see this: (only "Tom York"'s games, as expected)
image

But if I look up `tom` or `to` I get this:
image
Duplicate games! where the duplicates come from each different tom.

If I look up just `t` it gets worse, since now "Cait"'s games also are included:
image

It get's even worse if I look up "game":
image

as there are multiple varying amount of duplicates per game.

You will want to think about how to address this and what your desired/expected behaviour is when looking up a person that matches multiple names. **As well as generally what to do with duplicate results**

You could consider adding extra information to each search result saying how/why it was matched by the user's query.

You may also consider having some way to limit the searches to only people associated with a game, (like maybe a dropdown option?).

I'm still approving, more work is needed to flesh out the search results.

ptlin25:
This pull request has been merged as the search bar successfully supports querying for games by the publisher's name, artist's name, or designer's name. However, there are many issues to fix. Most notably, the search results currently return duplicates. This should be fixed by adding `distinct()` after filtering the Game objects. We may also want to make searching by a person's name part of a different drop down menu option. For example, the `Game` option would query by game name and `Game – by related people` option would query by producer's name, artist's name, or designer's name.

majorsylvie:
Issue Score: Excellent

Comments:
Wonderful closing comment @ptlin25 ! I appreciate you calling out where this code falls short as well as suggesting future functionality and actually explaining it with example.

Then also thank you for actually writing updates in response to the reviews done by me and will. 

@willcravitz thank you for your thorough review! 

Be proud of you work y'all! You deserve to be :)

Issue #226, opened 2023-11-14
Add auto-create chat when a tournament is generated


MuyanXie:
This issue seeks to add the functionality to automatically create a live chat whenever a new tournament is created. Specifically, the main way to implement this would be to add a function that instantiates a new `Chat` object that has the current tournament as its parameter.

MuyanXie:
Update: Due to the update in our model. I wonder if we would like to close this issue or re-define our scope to be "add auto-create chat when a tournament is generated" @rqueiroz20

rqueiroz20:
> Update: Due to the update in our model. I wonder if we would like to close this issue or re-define our scope to be "add auto-create chat when a tournament is generated" @rqueiroz20 

This is a good idea, I will update the title right now.

MuyanXie:
@rqueiroz20 Are you still able to do this PR before the Soft Freeze? If not, we will have to push it to back log

rqueiroz20:
> @rqueiroz20 Are you still able to do this PR before the Soft Freeze? If not, we will have to push it to back log

Yes I am, I just finished the code for it now. Will push my branch now.

rqueiroz20:
This issue led to tournaments automatically having a live chat associated to them when that tournament is created. Some of the specific changes include the following:
- Modifying the `TournamentCreateView` class to also instantiate a new `Chat` object associated with the tournament.
- Adding a button labeled with the tournament name that takes the user to the frontend live chat display.

@MikeChen012345 and @MuyanXie talked over how such a feature would be implemented and the pseudo-code necessary. @rqueiroz20 worked on implementing the code in the branch itself, and @Riyush reviewed the code and made sure it aligned with the pseudo-code.

elizabethli31:
Issue Score: Excellent

Comments:
Nicely implemented auto-creation of chat!

PR #240, opened 2023-11-16
Tournaments/implement display chat and display chat content

### LINKED ISSUE #225: Implement display of chat box and chat content at the frontend
In this issue, we will create a very basic frontend display module for the chat content.

## TODO

1. Implement a tournament match chat display page
2. implement a js file that contains the logic for displaying, fetching, and auto-fetching/display changes
3. implement user-friendly features for users to use this as a chat box

## metrics

1. the user can now navigate to the page
2. the user can have auto-updates 
3. the user can scroll down the page
4. the latest message will always be at the bottom of the chat box

MuyanXie:
In this PR, we will try to address issue #225 . We will implement a chat page, a nice auto-fetch of message, a nice auto-display of new messages, and a nice scroll down feature

## Finished Features
- basic chat page
- auto fetch of messages
- auto scroll down when there are new messages
- mock messages

## For reviewers
Please navigate to ```games/tournaments/<pk of match>/chat``` once you have already created a match


src/chigame/games/views.py
```diff
@@ -196,6 +196,12 @@ class TournamentDeleteView(DeleteView):
     success_url = reverse_lazy("tournament-list")
 
 
+def TournamentChatDetailView(request, pk):
```
rqueiroz20, line 199: This function doesn't feel like it's exclusive to tournaments, but could be generalized to matches. I think we should talk to the Matches team and see if individual matches should have a live chat.

src/static/js/getChatUpdate.js
```diff
@@ -0,0 +1,123 @@
...
+    const message = fakeMessage[tokenId];
+    return message;
+  } catch (err) {
+    console.log(err);
+    return null;
+  }
+}
+
+// Function to simulate receiving messages from the backend
+function updateView(tokenId) {
```
rqueiroz20, line 53: This `update_view()` function assumes that users will be able to edit messages sent via a live chat. I don't know if that's what we want to implement since I'm not sure whether most live chat features allow you to edit a message that you sent. We should look into this though.

src/templates/tournaments/tournament_match_chat.html
rqueiroz20, line 1: I like the overall format of the live chat, however we should probably try to distinguish visually who sent what message. Perhaps one user's username is red and the other person's username is blue?

This is also unrelated, but I had assumed that the chat feature was going to be within the same page as the game itself. Is this just to demonstrate that a live chat is possible?

rqueiroz20 (commented):
Overall, very nicely done having a live chat displayed on the HTML page! A few comments: 

- The messages are hard-coded for now, though I assume at some point we will have to work with the APis team to allow users themselves to send messages into the live chat. That's another PR, though. 
- We should talk with matches to see if they want a live chat for individual matches, though I think they should and it would make sense given other websites like chess.com have one.
- We should distinguish who sends what message in the chat display itself (perhaps color-coded?) and I'm not really the biggest fan of the word "From" in front of the username. I think just the username followed by a colon would be sufficient.

Very well done! Re-assign me to review this PR once we've discussed the changes!

src/templates/tournaments/tournament_match_chat.html
MuyanXie: I think for now just having a separate page for a demonstration of chat is enough; I will work on deleting the From in front of the message; About the color, we can leave it for later PRs if we find it useful as per the current code structure it doesn't seems feasible to easily add the color

src/static/js/getChatUpdate.js
MuyanXie: We have discussed this already. We think we will reserve edit for later PRs; Please pay attention that the updateView here is purely for updating the display of the page to reflect new messages, and doesn't involve editing messages

src/templates/tournaments/tournament_match_chat.html
rqueiroz20: Sounds good to me!

src/chigame/games/views.py
MuyanXie: It seems per our chat model we are only assigning match as a foreign key to chat. We cannot really exclusively assign chat to each tournament but have to assign it to matches; Also for now, I think just per demonstration purpose it is fine and in later PRs in which we will "auto generate chat for each match (see issue #226 )", we will work on this if this is still a concern

rqueiroz20: Sounds good to me! We can always revisit this if you want.

src/static/js/getChatUpdate.js
rqueiroz20: I'm not even sure we need an edit for live chat messages. I'm not sure that functionality exists for a lot of live chat services. For example, in OverWatch 2, once you send a chat message, the user cannot edit it. So I'm not sure if our live chat should even have an edit feature.

rqueiroz20 (approved):
Nice job implementing the changes I requested. Of course, we can continue to talk more about the live chat and its specifications in future PRs, but for now this is a very nice demo of what's to come for a live chat. Great work!

src/static/js/getChatUpdate.js
MuyanXie: Yeah that's true, and that's why we have actually make this probably that last feature we will implement for chat; for now, we can just ignore edit

rqueiroz20: Sounds good to me!

src/chigame/games/views.py
```diff
@@ -196,6 +196,12 @@ class TournamentDeleteView(DeleteView):
...
+    match = Match.objects.get(pk=pk)
```
elizabethli31: image
I'm getting this error when trying to access chat. I think it's because the pk being used for match is the one from the tournament. I am on a tournament with pk=3, but the only match on this page has pk=2, and I do not have a match with pk=3. I'm a bit confused as to why you want the chat functionality only for matches right now. With the indexing and because it is being implemented into tournaments, it makes more sense to me to have a chat for the entire tournament as a whole.

elizabethli31 (changes requested):
Unable to view a chat. Pk isn't matching between match and tournament

src/chigame/games/views.py
MuyanXie: We matched chat to match as when we were defining the model for chat, we only assigned match as a foreign key to chat, as we want each chat to exclusive to each match. We can try to change the model to map toward tournament, but I think we intend to make chat mapped to match whening designing the ERD?

MuyanXie: Please see latest commit~

MuyanXie:
Made some changes to the foreign keys (in the models.py); Had to set a default value for the foreign key to not intervene with current scheme per django rules.

```
You are trying to add a non-nullable field 'tournament' to Chat without a default value; Address this by adding a default value.
```

I addressed this by adding "None" as a default value for existing chat rows. To solve this issue forever, we can delete all migration history and make one directly (not recommended). Please notice that I didn't really intend to make a default value -- it is to address the django makemigrations rules.

Now, it seems like the chat is mapped to tournaments and can be accessed to tournaments/{pk}/chat

src/static/js/getChatUpdate.js
```diff
@@ -0,0 +1,123 @@
+// Simulated frontend data, replace with actual implementation
+let currentTokenId = 0; // Assume the frontend starts with token ID 0
+const name = document.getElementById("name").value;
+console.log(name);
```
elizabethli31, line 4: Remove this print

elizabethli31: Remove this print

src/static/js/getChatUpdate.js
```diff
@@ -0,0 +1,123 @@
...
+  var element = document.getElementById("messageContainer");
+  element.scrollTop = element.scrollHeight - element.clientHeight;
+}
+
+function getMessages(tokenId) {
+  try {
+    const message = fakeMessage[tokenId];
+    return message;
+  } catch (err) {
+    console.log(err);
```
elizabethli31, line 48: Maybe needs some more robust error handling? I'm ok with this for now

src/static/js/getChatUpdate.js
```diff
@@ -0,0 +1,123 @@
...
+  });
+
+// Simulate initial message retrieval
+updateView(currentTokenId);
+
+// Simulate periodic message updates (you may use WebSocket or other real-time mechanisms)
+setInterval(() => {
+  console.log("Updating messages...");
+  updateView(currentTokenId);
+  scrollToBottom();
```
elizabethli31: This function is buggy because it doesn't let me scroll back to the top and remain there. I think you need to end the setInterval once we reach the last message.

elizabethli31 (changes requested):
Great work! This mock looks really good so far, and I'm excited to see the finished result! Remove the console.log statements, and also, I would prefer if you guys removed some of the comments as well except ones you believe are absolutely necessary (especially ones for the CSS are not necessary). In general, your code should speak for itself instead of through comments. Mostly, I'm requesting changes for the buggy scroll bar which should be a quick fix.

src/static/js/getChatUpdate.js
MuyanXie: Thanks! Done

src/static/js/getChatUpdate.js
MuyanXie: Gotha. Will implement more robust error handling in later Prs

src/static/js/getChatUpdate.js
MuyanXie: Opsi. Sorry for this. I just fixed this in the latest commit (also merged with the latest dev branch). This fix the issue -- you will still be taken down to the bottom of the page if you receive new message; If there's no new message, you can definitely stay at the top without any interference

elizabethli31 (approved):
Thanks for fixing the scroll. Looks good to me!

MuyanXie:
Closing Statement:

In this PR, @MuyanXie and @MikeChen012345 wrote the majority of the code in the logic about display the chat and chat content, especially in the frontend.

@jb-01 guided us through the research about the frontend implementation, and @rqueiroz20 helped us review the PR and finished the logic

elizabethli31:
Issue Score: Excellent

Comments:
Very cool first iteration of the chat and nicely done review.

PR #244, opened 2023-11-16
Set format_js in pyproject.toml to be false

### LINKED ISSUE #242: Set format_js to false
Currently, with format_js set to be true in the djLint configurations, Django template tags, when used with inline JavaScript, is being reformatted due to stylistic issues and causing it to no longer work. Disabling format_js may be an interim solution.

ptlin25:
Setting format_js to false will fix the issue of the linter reformatting and breaking Django template tags in JavaScript.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,7 +90,7 @@ generated-members = [
 blank_line_after_tag = "load,extends"
 close_void_tags = true
 format_css = true
-format_js = true
+format_js = false
 # TODO: remove T002 when fixed https://github.com/Riverside-Healthcare/djLint/issues/687
 ignore = "H006,H030,H031,T002"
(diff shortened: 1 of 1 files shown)


borjasotomayor:
Looking into this in more detail, it looks like it's possible to selectively turn off djlint as described here: https://www.djlint.com/docs/ignoring-code/ I think that may be a better approach than turning off Javascript formatting entirely. Could you check that wrapping the <script> tag in your original PR (the one where this issue came up) works? If so, let's just go with that and I'll close this PR without merging it.

ptlin25:
Wrapping the <script> tag with the djlint tags works correctly and the linter ignores the code in between. This pull request can be closed without merging.

PR #246, opened 2023-11-17
Added medium (50) and big (500) game fixtures

### LINKED ISSUE #98: Write up fixture files for games models that have the updated attributes
During the first sprint we were all testing the crud pages with random data we had created in our local databases. Now we want to create a set of game fixtures so that we can test on the same data and have controlled results.

Our initial goal is to create 3 fixtures:

1. A "small" fixture of about 5 games. This will be used for quick visual checks on crud and display pages.
2. A "medium" fixture of about 50 games. This will be used for visual checks that involve pagination, sort, etc.
3. A "large" fixture of about 500 games. This will be used to test search, sort, filter, and the efficiency of these operations.

frowenz:
Corresponding issue: https://github.com/uchicago-cs/chigame/issues/98

In this PR, we created game fixtures with 50 and 500 games. We did this by using the BGG API to pull game information en masse. Currently, the scripts we wrote to do this are comments on the corresponding issue. If it would be useful we could also upload them to the wiki or put them in a utility folder.

One other subtle changes that was made was changing the `category` attribute to `categories` in the previous fixture with 5 games to align with the changes that will be made in PR https://github.com/uchicago-cs/chigame/pull/236 (hence, this PR is dependent on that one).

Notably, the `categories` and `mechanics` attributes are not filled out as they are a bit harder to create from API information. This is a good future item to work on once https://github.com/uchicago-cs/chigame/pull/236 gets merged.


danielngira (commented):
The structure of the fixtures are accurate and contain the correct information. All the fields are present in the fixtures. The image URLS appear to be working as well. I have no changes to suggest for this issue.

majorsylvie (changes requested):
Currently, even after pulling from dev, the game model still has the field name as `category` not `categories` which causes all 3 fixtures to unsuccessfully run when tested with

`$ p3 manage.py testserver src/chigame/games/fixtures/games-fixture-500` (which is the same as calling the fixture with `.json`)

with error:

```bash
ls/options.py", line 683, in get_field
    raise FieldDoesNotExist(
django.core.exceptions.FieldDoesNotExist: Game has no field named 'categories'
```

Then after manually changing the `category` field in the game model to `categories`, the 5 and 50 item fixtures pass, but the 500 does not.

This is because model `pk=173`, Go, has a `year_published` == -2200
Which then causes this error:

```bash
  File "/Users/fkb/ucclasses/year3/f23/220/chigame/venv/lib/python3.11/site-packages/django/db/backends/sqlite3/base.py", line 328, in execute
    return super().execute(query, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
sqlite3.IntegrityError: CHECK constraint failed: year_published
```

This is monkeypatched by manually updating Go's publishing year to be non-negative (eg: 2200)

---
Tasks to come out of this:
- You will want to update the BGG wiki documentation to include info on what fields are associated with a game and values they can have (thus including `year_published` making a note that pre 0 years are just marked as negative.

- You also should update the games model to support negative years if you want to keep it as an integer field without adding BC/AD information to every year.

- Then you should update the game model to actually have the attribute `categories` instead of `category`, such that running `$ p3 manage.py testserver src/chigame/games/fixtures/games-fixture-500` works.

majorsylvie:
poking deeper, it seems you did not commit the discussed changes for making the `year_published` an `IntegerField`. Please add those!

frowenz:
The mention of changing `category` to `categories` in the opening comment was only in reference to the old 5 game fixture since other PRs were meant to address this.

frowenz:
Forgot to push this change
https://github.com/uchicago-cs/chigame/issues/98#issuecomment-1815589177

majorsylvie:
Thank you for the clarification! #236 has been merged now.

majorsylvie (approved):
After merging #236 all that was needed was pushing the `year_published` update commit! 

These features look good! 

Next #239 will need to update the search to use `categories` instead of `category` 

as right now, searching on on this branch with what's in dev will lead to:
```
FieldError at /games/search/

Cannot resolve keyword 'category' into field. Choices are: BGG_id, categories, complexity, description, expected_playtime, id, image, lobby, match, matchproposal, max_players, max_playtime, mechanics, min_players, min_playtime, name, people, publishers, rules, suggested_age, tournament, year_published
```

danielngira (approved):
The data in the 3 fixture files look correct and the update to the year_published attribute in the game model is consistent with the expected field type.

frowenz:
Closing comment:

As we set out to do, this PR resulted in the creation of 2 fixtures: games-fixture-50.json with 50 games, games-fixture-500.json with 500 games. These fixtures were generated by a script which takes BGG ids and outputs json models of the games using the BGG api. The script used can be found [here](https://github.com/uchicago-cs/chigame/wiki/Games-~-Creating-Fixtures-Using-BoardGameGeek-API).

The next big step is to add the capability to scrape the category and mechanic attributes. The main difficulty is the fact that these exist as separate entity in the database.

Developer contributions:
@willcravitz and @frowenz developed the script the generate fixtures from BGG api, used it to create the 50 and 500 game fixtures, and made some updates to the game model.
@ptlin25 provided feedback and tested.

majorsylvie:
Issue Score: Excellent

Comments:
Great work!
Appreciate the very good closing comment @frowenz 

Then these fixtures have been very useful for testing your other functionality like searching and filtering! 
@danielngira thank you for, even when approving the PR, adding detail of what works and how. This makes me more confident that you've actually engaged with the code than you saying "Looks good" with no other comments.

@frowenz @willcravitz wonderful work with the fixture generation script! Simply cool work that has significantly helped the project. 

All of you deserve to be proud of the work you've done!

PR #257, opened 2023-11-20
Added filters for the lobby list page

### LINKED ISSUE #219: Implement django filter on Lobby List view
**Goal:** The lobbies list page lists all lobbies with no specific organization. To make lobby searches more efficient, we want to implement advanced filtering capabilities where users can choose to display only lobbies with certain games, players, or match start times. We've done the research for what we need to do in order to have the filter appear on the view. We need to now decide which attributes we want to allow users to filter in their search.

**Deliverables:**
- Pull Request

**possible mini-steps:**
- create `filters.py` with LobbyFilter class
- identify which attributes we want to allow filters for
- for each attribute:
    - identify the lookup expressions we want
    - add it to `Meta.fields`
- add function to `views.py`
- add filters to HTML template

jiyeongchloe:
The lobbies list page lists all lobbies with no specific organization. To make lobby searches more efficient, we want to implement advanced filtering capabilities where users can choose to display only lobbies with certain games, max number of players, or lobby names.
`filters.py` was created to achieve this.


nugerhard (approved):
This works for me! I think this is sufficient for the current PR, but some notes that I had were:

- It would be cool if we could get the filters to align vertically with the fields we are searching on. I really have no idea how we would do this, as the the Django-filters and Django-tables features are sort of self-contained (in that, I don't know how we would get them to coordinate in the html).
- Players definitely should be able to filter on max number of players, as we have implemented here. As such, the table should probably display the max players of the lobby. This is a quick fix, so I'll probably just add this in one of my PRs later this week.

elizabethli31 (approved):
The filters look great, and I really like that you've accounted for lowercase and subsets too instead of an exact search. Agreed with Nat that there could be more frontend design changes, but for now this is good with me!

jiyeongchloe:
**Closing Comment:**
PR is merged and finished.
We added three different filters (that can be used in conjunction with each other) that allow users to search for lobbies that contain the queries the users type in.

elizabethli31:
Issue Score: Excellent

Comments:
Same comments as the associated issue, but nice job on this PR!

Issue #93, opened 2023-10-30
Create a basic search that returns search results for games


rakrusch:
We will create an initial search bar that returns results without any autocompletion or suggestions. We will implement a basic search bar that has a dropdown menu that can be used to specify whether the query is for games or users. This search bar will be placed in the navigation bar.

ptlin25:
Created new branch `game-management/basic-search-bar` for this issue

ptlin25:
Added html code for search bar

ptlin25:
Search bar has been created. There is a dropdown menu next to the search bar to specify whether the user is querying users, games, etc. The dropdown menu dynamically changes the action attribute of the search bar. In other words, when the user switches the selected option in the dropdown menu, the search will link to different pages.

ptlin25:
I was thinking about implementing full text search, but it requires using PostgreSQL while ChiGame is currently using SQLite. if we switch to PostgreSQL in the future, we can think about using PostgreSQL's more advanced querying options

ptlin25:
The basic search bar has been complete and added to the navigation bar. The search bar includes a dropdown menu to specify if the search is for games, users, etc. This issue will now be closed and completed via pull request #176. The next steps include adding autocompleting for game names (#95) and adding support for searching for games by the producer, artist, or designer (#170).

majorsylvie:
Issue Score: Excellent

Comments:
@ptlin25 In the future, if you're editing someone elses comment significantly (like the summary here), please mark that you edited the comment with your name, since it's not obvious that _someone else_ edited the comment when all it says is `(edited)`.

For the task, great work! I particularly appreciate you mentioning the desired expansion to PostgreSQL as well as linking relevant future issues and related PR's.

Makes this issue well-documented and easy to understand.

Keep going!

PR #262, opened 2023-11-22
(DELETED) Implement the logic for users searching 

omarelkashef:
This PR builds on what has been done by the games team regarding a universal search bar #93 . We are handling the logic specific for user searching but keeping things consistent. That is, we are using the same search results landing page but adding conditional check that displays specific information about each matching result according to the model being queried. We also added a message to be displayed if there are no matching result and prevented the form from being submitted if it is empty regardless of the model. In the search bar, you can search by username or email. First and last name are not currently handled since they are None in the User model.


src/chigame/users/urls.py
```diff
@@ -26,4 +27,5 @@
     path("accept_friend_invitation/<int:pk>", view=accept_friend_invitation, name="accept-friend-invitation"),
     path("decline_friend_invitation/<int:pk>", view=decline_friend_invitation, name="decline-friend-invitation"),
     path("user_history/<int:pk>", views.user_history, name="user-history"),
+    path("search-results", user_search_results, name="user-search-results"),
```
AndrewM131: This should be view=user_search_results or view.user_search_results so it can properly go into views.py

AndrewM131 (changes requested):
Your views.py looks very similar to the games team views.py for the search results, but you took into consideration the case of no users, which is good. The updated HTML file looks properly implemented. The only thing I see is updating urls.py.

PR #267, opened 2023-11-22
Functionality of Placeholder Game Constructed

### LINKED ISSUE #266: Create Placeholder Game Functionality
We are looking to create view(s) which stands in for a game, after a lobby has met its start conditions. This should be informative for future classes, to see at what points database instances are spun out: when the lobby, match, and player entities are created. 

Deliverables: a user should be able to click through some view, which subsequently tells the user whether they won or lost. Corresponding match entity should be created. For this task, because it is substantial, we are not concerned with linking this to the rest of the site yet: we just want to be able to access and walk through the game.

### LINKED ISSUE #323: Prevent crashing behavior in Placeholder Game
This is the second half of the actual implementation task for the dummy game we constructed. Essentially, we initially built the game to simply crash if there was an integrity error in the database (e.g. if someone tried to play in a match twice and thus created two player records, or if two matches were created from the same lobby). This issue is meant to address that.

nugerhard:
See issue #266. Larger placeholder game issue broken up into two issues, as creating the coin flip game was slightly more involved than anticipated, and as the status of the join lobby button is unclear.


xylucien (changes requested):
This is a fun feature to add! Code structure looks good to me, but when I try to play around with it on http://127.0.0.1:8000/games/lobby/1/coinflip (after loading the fixtures), the first run works well, but the second time I make a guess the app crashes.

xylucien (approved):
After discussing it makes sense that each user can flip the coin only once in a lobby. All lgtm now and I would expect a separate issue to prevent a user from flipping the coin 2+times!

src/chigame/games/views.py
```diff
@@ -209,3 +211,42 @@ def search_results(request):
...
+@login_required
+def check_guess(request, pk):
+    user_guess = request.POST.get("user_guess")
+    coin_result = choice(["heads", "tails"])
+    correct_guess = user_guess == coin_result
+
+    lobby = get_object_or_404(Lobby, id=pk)
+
+    # Create Match instance linked to the fetched Lobby
+    match = Match.objects.create(
```
elizabethli31: I'm getting an error here because I already have a match for this particular lobby. 
image
I know this is just a dummy game, but I'm not sure if building on top of a pre-existing lobby is the best option because of this possibility. An alternative could be to check for match instances first then prevent this page from being reached if there is one.

elizabethli31 (changes requested):
Looks really good! I left a comment regarding error handling which I believe is in the scope of lobbies. As for Xinyu's comment, I'd prefer some sort of error message instead of the app crashing if the user guesses twice. You could just put everything in the check_guess function into an if statement to see if the user has guessed already and redirect back to the coin flip page. I believe this is the same error that I mentioned in my comment, so you could knock out both together.

nugerhard:
Ok, had some issues with merging- I should've merged with the remote branch prior to implementing the fixes suggested by @elizabethli31. That is resolved now; since this non-crashing behavior took some time to implement, I created a second issue for it. A description of the changes can be found associated with that issue: #323.

src/templates/games/game_already_played.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+
+<!DOCTYPE html>
+{% block title %}
+  Error: Already Played
+{% endblock title %}
+{% block content %}
+  <h1>Already Played</h1>
+  <p>Looks like you have already played this game. How'd you get here?</p>
```
xylucien, line 9: LOL

xylucien (approved):
I have played around and things work just fine. Visiting the coinflip page after a coin flip won't crash the app now.

elizabethli31 (approved):
Thanks for fixing the changes! Game is working properly now with a pre-existing lobby and allows users to flip multiple times. LGTM!

nugerhard:
Closing Comments: This is a toy example of a game (unclear if a coinflip can even be considered a game). But it has a notion of winning and losing, and an arbitrary number of people can "play." And since it is integrated with lobbies and creates the relevant match and player data, it should serve as a temp.

elizabethli31:
Issue Score: Excellent

Comments:
Like comments on the issue, this PR was well implemented.

PR #272, opened 2023-11-23
Integrated nav bar with forums page

### LINKED ISSUE #215: Add a link to the forum to the chigame navigation bar
The purpose of this task is to add a link to the forum to the chigame navigation bar. This will ensure that the forums can be easily reached from rest of the chigame website.

### LINKED ISSUE #216: Add the chigame navigation bar to the forum pages
The purpose of this task is to add the chigame navigation bar to the top of all forum pages. This will ensure that the forums are integrated with the rest of the chigame website and that users are able to easily navigate to other parts of the platform.

Benjoman12:
Per issue #215 and issue #216 I have added both the nav bar to the the forums page but also a link to the forums page to the nav bar.


bburgess19:
I've pulled the changes down, and I've only noticed one small thing. It may not be related to your changes, but when clicking on the forums tab, you link to `/forums` as expected, but the "Home" tab label is prepended with "(current)," and that doesn't happen when clicking on any of the other tabs.

One inkling I have is that django-machina's imported stylesheets overwrite or don't style the "visually-hidden" label applied to the html containing "(current)" as it clearly is styled away when inspected with developer tools. Just something to keep in mind. Otherwise, things look great @Benjoman12!

src/templates/navbar.html
joshuamonroe, line 1: As far as I can tell, it shouldn't be necessary to break the nav bar out into its own file like this. It's probably better if we include or extend from base.html itself, even though it may require a bit more work to get working properly, so that we don't lose any functionality on the forum page (e.g. search related JS) now or later down the line. It also appears to be contributing to the [wonky behavior](https://github.com/uchicago-cs/chigame/pull/272#issuecomment-1826423475) mentioned by @bburgess19. Would be interested to hear your reason behind doing it this way in case I'm missing something here though.

src/templates/navbar.html
```diff
@@ -0,0 +1,64 @@
...
+      <span class="navbar-toggler-icon"></span>
+    </button>
+    <a class="navbar-brand" href="{% url 'home' %}">ChiGame</a>
+    <div class="collapse navbar-collapse" id="navbarSupportedContent">
+      <ul class="navbar-nav mr-auto">
+        <li class="nav-item active">
+          <a class="nav-link" href="{% url 'home' %}">Home <span class="visually-hidden">(current)</span></a>
+        </li>
+        <li class="nav-item active">
+          <a class="nav-link" href="{% url 'forum:index' %}">Forums</a>
```
joshuamonroe: Is there a reason that we can't simply include this element in base.html?

src/templates/forum/board_base.html
```diff
@@ -0,0 +1,73 @@
...
+{% block css %}
+  {{ block.super }}
+  <link rel="stylesheet"
+        href="{% static 'css/machina.board_theme.vendor.min.css' %}" />
+  <link rel="stylesheet"
+        href="{% static 'css/machina.board_theme.min.css' %}" />
+{% endblock css %}
+{% block body %}
+  {% block header %}
+    {% include 'navbar.html' %}
```
joshuamonroe: Changing this line to include base.html instead, without changing anything, else fixes the issues that @bburgess19 mentioned but also removes some of the styling that is present otherwise and shows the "Coming soon!" text at the top of page. Although it may require a bit more work to configure properly in the short term, I think we should consider finding a way to extend from chigame's base.html as an alternative approach.

Also, nuking the existing django-machina nav bar code from this block cuts off our access to django-machina's search functionality. If we can't easily roll this functionality into chigame's nav bar as a part of this PR, we should at least open a separate issue to get that on record and, ideally, integrated soon after.

src/templates/forum/board_base.html
```diff
@@ -0,0 +1,73 @@
+{% extends MACHINA_BASE_TEMPLATE_NAME %}
```
joshuamonroe, line 3: While we could change this line to directly extend from chigame's base.html. We can also set the MACHINA_BASE_TEMPLATE_NAME variable to point to base.html in settings/local.py. However, doing so using either method, for some reason that I haven't been able to ascertain yet, removes a few elements that should be present, like the breadcrumb links (e.g. "Forum Index"), Subscriptions, View My Posts, etc. If we can figure out how to get those elements to show up properly, I think it makes more sense to extend from base.html, rather than include it.

src/config/settings/base.py
```diff
@@ -197,6 +197,7 @@
         # https://docs.djangoproject.com/en/dev/ref/settings/#dirs
         "DIRS": [
             str(BASE_DIR / "templates"),
+            str(BASE_DIR / "templates/forum"),
```
joshuamonroe, line 202: This line seems less than ideal to me, since we are already including the templates directory. However, I see why you did it this way. Because of where django-machina is expecting the templates to be, putting them straight into the templates directory could get messy, and this prevents that by keeping them contained in their own directory. I wish we could find a cleaner solution though.

joshuamonroe (commented):
I few issues that I think need to be addressed before merging this PR. 
1. The issue explained by @bburgess19 in [this comment](https://github.com/uchicago-cs/chigame/pull/272#issuecomment-1826423475) and pictured below. I will also note that the nav bar shifts positions slightly to the right when on the forum page and highlights both the 'Home' and 'Forums' links.
brave_screenshot (1)

2. We should discuss whether it is necessary to break the nav bar out into a separate file or if we can (and, I believe, should) simply use base.html itself. Furthermore, we should determine if it would be more appropriate to figure out how to extend from, rather than include this file. See the code comments for more detailed thoughts on this.

3. We should be aware that by nuking the existing django-machina nav bar code we are removing access to django-machina's search functionality. If we can't easily roll this functionality into chigame's nav bar as a part of this PR, we should at least open a separate issue to get that on record and, ideally, integrated soon after.
brave_screenshot (2)

src/config/settings/base.py
majorsylvie: If it has to be this way to comparmentalize machina files, then a code comment should for sure be put in here explaining with pretty good detail why this is the case.

src/templates/forum/board_base.html
majorsylvie: Which parts of this file have been modified by you vs coming from Machina? I know for sure this line is you @Benjoman12

src/templates/navbar.html
majorsylvie: I could see it as a decision to make the code more modular, such that as more is added to the navbar, a conceptually distinct component, it is all contained in one independent file.
@Benjoman12 any comment?

majorsylvie (changes requested):
I concur with Josh that losing the Machina-provided search bar is not ideal. 

After addressing the problem with the `(current)` showing up as mentioned multiple times.

Another way that this could be approched is actually treating the forum as a somewhat distinct entity, where upon entering the forum, the site shifts to being forum-centric, with things like the nav bar, theme, and search bar all changing to only be for the forum.

Then if a user wants to go back to the "main site", they could navigate backwards / click on a big `ChiGame` button in the top left.

This is effectively what is on `dev` right now, minus a link to the forum on the "main site" nav bar.

I'm not saying this is the best solution, it is one you can consider. 

If you're looking for a unified navbar across both, you could extend `base.html` and then replace the game / user search bar with the machina one for the forum. 

Another option would be to investigate how the machina forum search bar works and see if that functionality can be rolled into the "main site" search bar:
image

At this point, there are multiple valid ways forward with how you want to create this nav bar, talk as a team and figure out what you want to try and why. 

I'm requesting changes because at minimum the `(current)` bug needs to be fixed.

src/templates/navbar.html
Benjoman12: The only reason for this was to increase the modularity of the code. Currently, customization of a specific element within the base.html is made harder then it needs to be due to its conglomerated nature. Increasing the modularity allows us to customize with more ease. But, if this modularity is causing more problems then its worth we can definitely discuss just extending the base.html

src/config/settings/base.py
Benjoman12: To override pre made django-machina html templates we have to add files to the projects templates folder that have the same name and we have to follow the same file structure as the django-machina templates folder. The issue is that this means for the program to actually override the file in question, board_base.html, without telling the program to specifically search in the templates/forum folder we would have to store board_base.html in the base of the main projects templates folder. This is messy and creates organization issues. The solution to this is to put it in a folder called forums and tell the program to search that folder for it.

Benjoman12: All of that being said if we are fine keeping this line:  str(BASE_DIR / "templates/forum"), in the settings file then I will definitely add a comment to it.

joshuamonroe: Comment looks good, you might also add something to the effect of "... discover the correct templates without cluttering the templates base directory" to clarify further the necessity of this line.

src/config/settings/base.py
```diff
@@ -318,9 +320,10 @@
 
 
 # Add additional configuration below:
-# ------------------------------------------------------------------------------
+# -----------------------------------------------------------------------------
```
joshuamonroe: The changes here, and on line 326 seem harmless but also unnecessary and inconsistent with the rest of the file.

src/templates/base.html
```diff
@@ -119,7 +126,6 @@
         {% endfor %}
       {% endif %}
       {% block content %}
-        <p>Coming soon!</p>
```
joshuamonroe: It doesn't seem necessary to remove this line and leaving it in doesn't seem to cause any issue with the forum page now either.

src/templates/forum/board_base.html
```diff
@@ -0,0 +1,82 @@
+{% extends "base.html" %}
```
joshuamonroe: Can we keep the original extends statement from the django-machina template and modify the variable in django-settings? That way it remains consistent with the django-machina docs. I've tested it and it doesn't seem to cause any issues to do it this way.
```
{% extends MACHINA_BASE_TEMPLATE_NAME %}
```
Then put this at the end of settings/base.py. I've also included the setting to change the page title.
```
# DJANGO-MACHINA SETTINGS
# ------------------------------------------------------------------------------
# https://django-machina.readthedocs.io/en/stable/settings.html
MACHINA_FORUM_NAME = "ChiGame Forums"
MACHINA_BASE_TEMPLATE_NAME = "base.html"
```

joshuamonroe (changes requested):
Looking good now! A few minor changes requested in the code comments and then I think this is probably ready to merge.

It may also be worthwhile to put something like the following at the top of board_base.html for future devs to quickly understand where this file came from and compare to the original, this would also help to address this [comment](https://github.com/uchicago-cs/chigame/pull/272#discussion_r1405266782) by @majorsylvie.
```
# This file was derived from the following django-machina template.
# https://github.com/ellmetha/django-machina/blob/main/machina/templates/machina/board_base.html
```

joshuamonroe (approved):
Good to merge as far as I'm concerned. Great work!

majorsylvie (approved):
As of now, there is a slightly different offset on the forums version vs not.

Besides that works just fine! 

I'm approving, get working on the rest of your PRs!

https://github.com/uchicago-cs/chigame/assets/86740254/bd80e8fc-261c-4a56-9978-fbd0196798fe

joshuamonroe:
Thanks to @Benjoman12 for making the PR and incorporating numerous pieces of feedback during the code review process, which was conducted by myself (along with merge deconfliction) with additional comments from @bburgess19. The following issue has been opened to address @majorsylvie's [comment](https://github.com/uchicago-cs/chigame/pull/272#pullrequestreview-1756620033) regarding the inconsistent nav bar offset.
- #324 

Additionally, further refinement of the search functionality has been deferred to the following issues.
- #307 
- #310

majorsylvie:
Issue Score: Excellent

Comments:
Phenomenal work all of you! 
Great closing comment @joshuamonroe 
Incredible work @Benjoman12 on both getting this integrated as well as taking feedback throughout the PR review process!

You deserve to be proud of what you've done.

Issue #261, opened 2023-11-22
add button to see / make reviews on game detail page


danielngira:
Include a button in the game detail page from #40 that allows a user to make a review of a game. Make a list view for the reviews as well to display a list of reviews.

danielngira:
Created branch game-management/reviews-page

danielngira:
Made a form ReviewForm for creating new reviews

danielngira:
Updated the GameDetailView with a FormMixin to allow a form to create reviews for games

danielngira:
Made a ReviewListView and a game-review template. Also updated the URLs to include a URL to the game-reviews page for a specific game

danielngira:
Updated the game-detail template page to include a form to create new reviews

danielngira:
Updated the `ReviewForm` to include `is_public` and a `rating` field

danielngira:
![Image](https://github.com/uchicago-cs/chigame/assets/81624602/a6a7c858-dc6b-4f46-a846-9a2c532184bf)

The review list page functions and only shows public games

danielngira:
This issue is closed as per PR #275. Tasks I worked on:

1. I made a `ReviewForm` to allow users to leave ratings and reviews for games. 
2. I updated the `GameDetailVIew` to render the form.
3. I made a `ReviewListView` to render  a list of reviews
4. I updated `game_grid.html` to display the review form
5. I made a `game_reviews.html` to display a list of reviews that a game has

majorsylvie:
Issue Score: Excellent

Comments:
Great work @danielngira !
Quite nice closing comment. Good job updating as you've worked! The kind of use of updates we want.

Issue #40, opened 2023-10-23
(game management) Create Individual Game Detail View and Template Page (Patrick, Ryan)


ptlin25:
Add detail template page for individual games. This page will show the name of the game, min and max number of players, game rules, and other game attributes specified in issue #39. A corresponding Django view must also be created to display the game detail page

rakrusch:
Created branch game-management/game-detail for this issue, as well as pull request #71

rakrusch:
Updated pull request per feedback. Updates include: adding links back and forth between game list and game detail pages, extended game detail page template from base.html, removed code for registering models (to be completed in separate issue)

ptlin25:
Completed as of #71, basic detail page for each game has been added 

Work done by each assignee:
Ryan: coded the detail view and template
Patrick: general feedback and basic review before pull request

majorsylvie:
Issue Score: Excellent


Comments:
Great job y'all!
I appreciate y'all using issue/PR numbers to directly associate this issue with related issues and PRs.
Thank you for including the closing comment with the developer contributions.

PR #274, opened 2023-11-23
Updated models, made review entity

### LINKED ISSUE #271: update game model with reviews entity
In this issue, we will update the game models in order to introduce a review entity. This is in response to issue #261 in which we will add a button to make reviews and view the reviews in the game detail page that was made in issue #40. This issue is consistent with the user stories that relate to users reviewing games which can be found here [https://github.com/uchicago-cs/chigame/wiki/Games-~-User-Stories#rating-viewing-and-favoriting-games](url)

danielngira:
Corresponding Issue: #271 

In this PR, I updated the game models in order to introduce a review entity. This change will be useful in issue #261 where we will add a button to make reviews and view the reviews in the game detail page from issue #40.


src/chigame/games/models.py
```diff
@@ -353,3 +353,11 @@ class Message(models.Model):
 
     def __str__(self):  # may be changed later
         return "Message from " + str(self.sender) + ": " + self.content
+
+
+class Review(models.Model):
+    "Represents a game review"
```
willcravitz, line 635: I have a few questions about how we want to handle reviews that could affect the way the fields of this model:
1. Can a user leave multiple reviews on a single game? This could mean using a through table or imposing uniqueness constraints.
3. Can a user update their rating or review?
4. If a user leaves a rating, do they have to write a review? If they write a review, do they have to leave a rating? Do they need to do one or the other? This could mean adding validation or adjusting the `null` arguments.
5. Are there visibility settings on reviews? This could mean adding an additional field.

willcravitz (commented):
I left a few questions about how we want to handle reviews on the site. I think these should be answered before we can determine whether the current model is sufficient.

src/chigame/games/models.py
danielngira: 1. Yeah a user can leave multiple reviews per game. I tried making the game and user attributes ManytoMany fields but then I realized that would mean multiple games and users can have the same review. Since each review will have a primary key and we'll use foreign keys from the user and game model, that will allow us to have a user make multiple reviews per game since each review will have its own primary key. Let me know if I've got that right.
2. Yeah we can allow a user to update a review. We can reflect this in the views with a new view for specifically editing a review.
3. I think a user can make a review, rate a game, or do both. I made the review and rating attributes nullable for this. Is that correct?
4. I agree there should be visibility settings. I added an attribute to determine whether the review will be public or not. We can implement this in the game detail page in issue #261.

willcravitz: 1. Sounds good. I think the way you have it set up makes the most sense. 
2. Also makes sense, no issue here.
3. I think what you have works but we should add an extra validation check that ensures at least the rating or the review is not null. This can be done by adding a `clean` method to the model.
4. Thanks for adding this. 

Thought of a couple more ideas for fields to potentially add:
5. Will reviews have an optional title? I'm fine with not having one but just wanted to raise it as a possibility. 
6. I think we should have a `created_at` field, which could impact the order in which reviews are displayed. One way to do this is with `created_at = models.DateTimeField(auto_now_add=True)`

danielngira: Thank you. I've addressed your comments in the following ways 

1. I added a `clean` method to validate that either a review is null or a rating is null but not both.
2. I added the `title` field
3. I added the `created_at` field

willcravitz: Great, thanks for addressing these!

src/chigame/games/models.py
```diff
@@ -471,3 +471,21 @@ def save(self, *args, **kwargs):
...
+    rating = models.PositiveSmallIntegerField(blank=True, null=True)
+    user = models.ForeignKey(User, on_delete=models.CASCADE)
+    game = models.ForeignKey(Game, on_delete=models.CASCADE)
+    is_public = models.BooleanField(default=True)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self):
+        return f"Review {self.id} by {self.user} for {self.game}: {self.review}"
+
+    def clean(self):
```
willcravitz, line 649: In order for the the custom `clean` method to take effect, you also need to write a custom `save` method. See either the `Game` or `Lobby` model as an example of how to do this.

willcravitz (changes requested):
Great job adding these extra fields! 

One more minor change is needed. There needs to be a custom `save` method for the validation you wrote to take effect. After this I'm good to approve.

src/chigame/games/models.py
danielngira: I've made this adjustment. Thanks for looking out!

willcravitz (approved):
All set! Thanks for going through all these changes.

src/chigame/games/models.py
```diff
@@ -517,3 +517,25 @@ def save(self, *args, **kwargs):
 
     def __str__(self):
         return "Message from " + str(self.sender) + ": " + self.content
+
+
+class Review(models.Model):
+    "Represents a game review"
+    title = models.TextField(blank=True, null=True)
+    review = models.TextField(blank=True, null=True)
+    rating = models.PositiveSmallIntegerField(blank=True, null=True)
```
majorsylvie: What is the expected range of the rating? Is it 1-5, 1-100? Something else?
If you have a specific one in mind please add a code comment **and also validation** that check whether the rating is in the correct range.

src/chigame/games/models.py
majorsylvie: When creating reviews via the admin page I can leave the title, summary, and rating blank. Were you able to test this validation and get it to work?

majorsylvie (changes requested):
Requesting changes to add documentation and validation to the `rating` attribute. 

Then, at least testing via admin, the validation is not run since I can leave all text fields blank and still save the model:

image

image

Please ensure the validation works and add what I've requested for the `rating`!

src/chigame/games/models.py
danielngira: I've added validation to the `rating` field and fixed the issue where blank rating and review fields were allowed.
Screenshot 2023-11-30 at 5 41 25 AM
I added a small comment under `rating`. I wasn't sure if that's what you meant by documentation or if there's another place/way I should go about it. Thanks!

danielngira:
In the latest commit, I run `make migrations` to reflect the changes I made to the model. I had forgotten to do this when I edited the model to reflect the changes @majorsylvie  requested

majorsylvie (changes requested):
As of now I am unable to properly create review entities via the admin interface:

image

Upon looking into this error I found these stack posts:
https://stackoverflow.com/questions/72062094/i-got-decimal-invalidoperation-class-decimal-conversionsyntax 

and

https://stackoverflow.com/questions/56458774/django-error-class-decimal-invalidoperation

Please either explain why I should not be able to save / add review models via the interface or fix this PR such that we can add models via admin.

I was allowed to do "save and add another" in sequence, and while doing that the all the restrictions worked: on needing one of a review or rating as well needing a user and game. 

So that part of your work is working as intended!

majorsylvie (approved):
> As of now I am unable to properly create review entities via the admin interface:
> ...

This previous error was a result of my local `db.sqlite3` database. 
I tried wiping the database, then re-adding review/users/games and then had no issues adding and viewing reviews in the admin interface.

The restraints on needing at least a body or rating + needing a user and game all work as intended.

I tested cases of:
_Valid, as expected:_
- title, body, and rating
- title, body
- title, rating
- body
- rating

_Invalid, as expected:_
- no title, body, nor rating
- title, but no body nor rating
- not supplying a user and/or game

Thank you for your work!

danielngira:
https://github.com/uchicago-cs/chigame/assets/81624602/54bf68f8-aadd-4a42-9887-919ead2b45fb

I'm currently trying to resolve the `decimal.InvalidOperation` error @majorsylvie run into. However, I am unable to recreate the same issue and the reviews are being saved on my end

danielngira:
- In this PR, I made a `Review` model that will store reviews and ratings that a user makes on a game. 
- The model has the following attributes:
1. Title -An optional review title
2. Review - An optional review
3. Rating - An optional rating between 1-5
4. User - A chigame user
5. Game - A game
6. Is Public - The visibility of a review
7. Created_at - The date the review was made
- This PR was useful in completing PR #275 where I incorporated a review form into the game detail page and made a template page for a game's reviews.

majorsylvie:
Issue Score: Excellent

Comments:
Phenomenal work! 
Awesome closing comment, listing individual components of your work makes what you've done very clear and readable!
Good work linking to relevant PR's.
Great job providing evidence of your work and current understanding through videos.
Great work taking the feedback I gave you on review.

@willcravitz Thank you for the good review! Appreciate you being willing to request changes.

Be proud of this work y'all!

PR #278, opened 2023-11-23
Unit tests for Notification model

### LINKED ISSUE #273: Add unit tests to Notification model
We need to add unit tests and factories for the newly constructed Notification models. Later, we want to test its integration with other models such as FriendInvitation.

omarelkashef:
This PR adds unit tests for the newly constructed Notification model. We created FriendInvitationFactory and BaseNotificationFactory that is a parent to specific notification type such as FriendInvitationNotificationFactory that we implemented.


AndrewM131 (approved):
From what I see, the files factories.py and test_models.py are properly implemented. What I really like is that it's very readable, and spaced out so it'll be a great format for future tests to be built off of and hopefully they'll replicate the same style

elizabethli31 (approved):
Tests all pass - great work! I like that you've grouped the mark_x and is_x methods which makes the tests condense and neat. LGTM!

omarelkashef:
Closing Statement: We created a number of unit tests for the newly constructed Notification model. The tests target all the methods that could be used to interact with the model. However, we are only testing notification coming from FriendInvitation so we could not test filter_by_type because that would require us to add factories for other models that can be actors. Future issues be handling this as well as the problem with get_by_actor() mentioned above.

elizabethli31:
Issue Score: Excellent

Comments:
Like comments on issue, good work on this PR.

PR #282, opened 2023-11-24
Implemented restricted versions of edit and delete views for lobbies

### LINKED ISSUE #280: Implement a delete view for lobbies
**Goal:** Users who created a lobby should be able to delete the same lobby in case it was created by mistake. We want to implement a Delete View of the CRUD pages to allow this to happen. We also need to figure out how to restrict access to this view to only the user who is the "host" of the lobby.

**Deliverables:**
- PR

### LINKED ISSUE #281: Implement an edit view for lobbies
**Goal:** Users who created a lobby should be able to edit the same lobby in case they change their mind or inserted a wrong number by mistake when initially creating the lobby. We want to implement an Update View of the CRUD pages to allow this to happen. We also need to figure out how to restrict access to this view to only the user who is the "host" of the lobby.

**Deliverables:**
- PR

jiyeongchloe:
Users who created a lobby should be able to delete and edit the same lobby in case it was created by mistake or there are some minor changes that need to be made. We implemented a Delete View and Edit View of the CRUD pages to allow this to happen. We also restricted access to these views to only the user who is the "host" of the lobby.

also, a minor change to the lobby_details HTML file was made to display user emails instead of names (which was set to None because of how the user model is set up)


agustinmoses (commented):
Looks good to me. Nice job ensuring only hosts of these lobbies can edit and delete!

agustinmoses (approved):
Looks good to me! Nice work on the changes!

src/chigame/games/views.py
```diff
@@ -58,6 +59,35 @@ class ViewLobbyDetails(DetailView):
...
+    template_name = "games/lobby_form.html"
+
+    def get_success_url(self):
+        return reverse_lazy("lobby-details", kwargs={"pk": self.object.pk})
+
+    def dispatch(self, request, *args, **kwargs):
+        # get the lobby object
+        self.object = self.get_object()
+        # check if the user making the request is the "host" of the lobby
+        if request.user != self.object.created_by:
```
elizabethli31: I don't think it was in your user stories, but it would make sense to also allow staff to edit or delete lobbies (in the case that lobbies have expired, modifications necessary, etc.)

elizabethli31 (approved):
LGTM! I left a comment you could adjust, but it's not necessary to. Nice error handling and simple buttons.

src/chigame/games/views.py
xylucien: Thank you for the suggestion. Adjusted!

xylucien:
The branch is now ready to be merged! Thank you for your time. @elizabethli31

jiyeongchloe:
**Closing Comment:**
We implemented edit and delete views that are only available to the lobby's host (nobody else can edit or delete). We also made sure to only display links to these on the frontend to the host.

elizabethli31:
Issue Score: Excellent

Comments:
Great job implementing CRUD for lobbies.

PR #284, opened 2023-11-24
Implement the logic for users searching

omarelkashef:
This PR builds on what has been done by the games team regarding a universal search bar https://github.com/uchicago-cs/chigame/issues/93 . We are handling the logic specific for user searching but keeping things consistent. That is, we are using the same search results landing page but adding conditional check that displays specific information about each matching result according to the model being queried. We also added a message to be displayed if there are no matching result and prevented the form from being submitted if it is empty regardless of the model. In the search bar, you can search by username or email. First and last name are not currently handled since they are None in the User model.


AndrewM131 (approved):
Thanks for addressing my requested change from my comment from the previous PR, everything else looks properly implemented!

src/templates/pages/search_results.html
```diff
@@ -6,10 +6,20 @@
...
-      <li>
-        <a href="{% url 'game-detail' game.id %}">{{ game.name }}</a>
-      </li>
-    {% endfor %}
+    {% if nothing_found %}
+      <p>Oops! No {{ query_type }} matching your search criteria. Please try again with different search parameters.</p>
+    {% elif query_type == "Users" %}
+      {% for profile in object_list %}
+        <li>
+          <a href="{% url 'users:user-profile' profile.user_id %}">{{ profile.display_name }}</a>
```
elizabethli31, line 14: I didn't catch it earlier, but there's an error in your `user_profile_detail_view` function in views.py that's affecting this line. When someone isn't logged in, they are still able to search which provides them with a list of user profiles matching the query. However, these pages aren't being made by the the detail view function because the request does not have an associated user id or pk (lines 89 and 93 in views.py). I think user profiles should still be visible even if not logged in so you could just change the `user_profile_detail_view` function to something like this `if request.user.pk == pk or not request.user.pk:` and add a log-in if statement to the associated html page to address the add friend button.

elizabethli31 (changes requested):
Search works really well, and I'm glad you've added in functionality to prevent empty searches! See my comment for the change. It's not from this PR, but it affects the search display and takes a quick fix.

src/templates/pages/search_results.html
omarelkashef: I have fixed this issue where now clicking on a link in the search results page to view profile while not logged in displays an Add Friend button. However, when clicking on that, it redirects to Login page.

src/chigame/users/views.py
```diff
@@ -87,14 +87,16 @@ def user_profile_detail_view(request, pk):
...
+        is_friend = None
         friendship_request = None
-        if not is_friend:
-            curr_user = User.objects.get(pk=request.user.id)
-            other_user = profile.user
-            friendship_request = FriendInvitation.objects.filter(
-                Q(sender=curr_user, receiver=other_user) | Q(sender=other_user, receiver=curr_user)
-            ).first()
+        if request.user.pk:
+            profile.friends.filter(pk=request.user.pk).exists()
```
elizabethli31: You would need to set this line equal to is_friend. Currently, even if a friend request is accepted, the profile page is still stuck on the accept or decline page.

elizabethli31 (changes requested):
Small change for friend request to work properly - see comment.

src/chigame/users/views.py
omarelkashef: Yes, sorry missed that!

elizabethli31 (approved):
LGTM!

omarelkashef:
Closing Statement: We added the user searching logic for the search box that takes username or email. Currently, the firstname and lastname fields are not queries since they are None in models.py  We used the same search results landing page that was created by the games team to keep things consistent. We also handled empty search and zero matching search queries. 
@omarelkashef  worked on the query logic for searching
@24raniwalar  worked on edge cases for empty searches and queries with no matching results

elizabethli31:
Issue Score: Excellent

Comments:
Great job incorporating changes to make userprofile views more robust as well as implementing the search option.

PR #196, opened 2023-11-13
BGG Autofilling Working on Create Pages

frowenz:
Corresponding issue: #97

## Description

In this pull request, the code in [this](https://github.com/uchicago-cs/chigame/wiki/Games-~-BoardGameGeek-(BGG)-API-Demo-Page) wiki page was integrated into the code base. This involved added views to search BGG games by ID or EXACT NAME. Because a lot of extra code was added to the views file, more concrete sections were created using comments. 

An AJAX script was added to the games form page to allow for searching and autofilling. Cu

https://github.com/uchicago-cs/chigame/assets/69015620/8a5760c4-bad4-4040-bd98-05603a3eddc8

## Future Work

1. Improve UI
2. Add in autofilling for mechanics and categories
3. Possibly add autofilling to the edit page

PR #287, opened 2023-11-25
Generated new game fixtures that correctly parse special XML characters

### LINKED ISSUE #283: Update medium and large fixtures to parse XML special characters
In issue #258 we noted that there are several characters in the game descriptions such as `&#10;` that we need to parse correctly. It turns out that it's hard to parse these characters on the frontend, and it's better to parse them when we make the call to the BGG API. This means updating the script to generate fixtures so that descriptions are parsed before they get stored.

willcravitz:
In previous game fixtures when we called the BGG API it returned a description with special XML characters instead of things like newline characters. We updated the [script](https://github.com/uchicago-cs/chigame/wiki/Games-~-Creating-Fixtures-Using-BoardGameGeek-API) used to generate those fixtures so that it parses special characters, then used that script to make new versions of 5, 50, and 500 game fixtures. We kept the old 5 game fixture but changed its name to `games-fixture-manual-5.json`. All game fixtures are stored in `src/chigame/games/fixtures/`.


frowenz (approved):
Nice work!

I tested each fixture and they all worked. I approving this PR.

One lingering questions is `Where does the script to generate this code belong?`. 

Currently, we have a bunch of scripts in wiki pages: [1](https://github.com/uchicago-cs/chigame/wiki/Games-~-Scraping-Categories-and-Mechanics-from-BGG), [2](https://github.com/uchicago-cs/chigame/wiki/Games-~-Creating-Fixtures-Using-BoardGameGeek-API), [3](https://github.com/uchicago-cs/chigame/wiki/Games-~-BoardGameGeek-(BGG)-API-Demo-Page). Given that some of this is code is:
1) being reused and improved upon and 
2) contains functions that are useful in other places (e.g., BGG autofilling , https://github.com/uchicago-cs/chigame/pull/196, would benefit from the same intelligent parsing),

I think it might be worth putting these scripts or parts of these scripts into a utilities folder, say `utils`. What do you think?

willcravitz:
I definitely agree that at the very least there should be a function somewhere in the codebase that, given a bgg api, generates a json object with all of the correctly parsed information. Another question is: do we need one implementation in python and another in JS?

majorsylvie (approved):
Looks good to me!

willcravitz:
Closing comments: this PR modified our existing BGG fixture generator to correctly parse XML special characters in the description, then used it to generate new 5, 50, and 500 game fixtures. The next steps are to find a permanent place in the codebase to put some of these helper functions we've developed, then use the correct XML parsing in game creation autofilling.

majorsylvie:
Issue Score: Excellent

Comments:
Wonderful work discussing the code and it's consequences in updates @frowenz @willcravitz !

@willcravitz Appreciate you being very clear on what files were generated and what you did with the old 5 fixture file. Makes it all super clear and understandable. 

Thank you for your work!

PR #291, opened 2023-11-26
Update the API POST Users endpoint (`/api/users`) so that it correctly creates a new User model

### LINKED ISSUE #218: Update the API POST Users endpoint (`/api/users`) so that it correctly creates a new User model
Currently, the API `POST` users endpoint at `/api/users/` treats the `User` model as a normal model, i.e. POSTing using a normal JSON post request (through Django REST Framework). This will actually not work to create new users because the `User` model has special attributes, specifically the `password` attribute. We will need overwrite the `create()` method of the `UserListView` in order to create a `User` object using the python command built into Django. 

We will know this issue is completed when the `POST` request to the users endpoint `/api/users` creates a new `User` model in our database. This will be reflected if we can see it when we make a `GET` request to the users endpoint.

abrahmasandra:
Previously, the API `POST` users endpoint at `/api/users/` treated the `User` model as a normal model, i.e. POSTing using a normal JSON post request (through Django REST Framework). This will actually not work to create new users because the `User` model has special attributes, specifically the `password` attribute. 

To fix this issue, the `create()` method of the `UserSerializer` was overriden in order to actually create and store a new `User`. Specifically, the `create_user()` method from the `UserManager` class is used to create new users.


cuyakonwu (approved):
Looks good! Is this done in collaboration with Esteban in regards to JWT?

abrahmasandra:
@cuyakonwu this does not incorporate JWT, but this is a backbone necessary (changing the serializer so that it creates users correctly) before we restrict the view to certain users (through JWT).

abrahmasandra:
Added a restriction on the `username` field of the `User` model, so that the `username` cannot be all numeric. This avoids any potential confusion between the user's `username` and `id`, when trying to find the detail view of a given user. This was pointed out as an issue by @cuyakonwu in PR #321

src/chigame/users/models.py
```diff
@@ -10,6 +11,14 @@
 from chigame.users.managers import UserManager
 
 
+def validate_username(value):
+    """
+    Validate that the username is not all numeric.
+    """
+    if value.isdigit():
+        raise ValidationError(_("Username cannot be all numbers."), code="invalid_username")
```
majorsylvie, line 20: You're changing a different teams model. You need to notify them about this.

src/chigame/users/models.py
```diff
@@ -22,7 +31,9 @@ class User(AbstractUser):
     first_name = None  # type: ignore
     last_name = None  # type: ignore
     email = models.EmailField(_("email address"), unique=True)
-    username = None  # type: ignore
+    username = models.CharField(
```
majorsylvie, line 35: Especially if you're adding a new field to a model you need to update the team and the whole project once you've got approval, since there could be many things that rely on the current set up of the `User` model that changing this would impact.

Only upon getting clearance from both the users team and after asking everyone in the general slack channel will I merge these changes.

majorsylvie (changes requested):
These changes are quite good! 
The big thing is that I will not approve changes to the user model until you have cleared any changes with the users team and have broadcasted any changes you make to everyone in the project.

Please reply with updates after you've communicated with the users team.

src/chigame/users/models.py
abrahmasandra: Screenshot 2023-12-01 at 12 28 03 PM

abrahmasandra: got clearance from the `users` team! will ask in the general slack channel

abrahmasandra: got a thumbs up in the general channel!!

24raniwalar (approved):
Nice job making password "write-only" in the UserSerializer class. I like having the max length of usernames set as well, along with the validating function that ensures that the username is not solely numeric. Was wondering if isdigit was the right function to use, but it takes a string not a char so LGTM!

majorsylvie (changes requested):
I was able to:
- POST a new user to `api/users/` with `name`, `username`, `email`, and `password`
  - then sign in with that user for the first time, finish confirming my email via the command line, then after using the confirmation link, signing in. 
  - signing out then signing in again with the same credentials, no confirmation needed.
- DELETE the `/api/users/<user_pk>/` for the just created user and was no longer able to sign in 

- POST a new user and confirm their email. Then PATCH each of the 4 fields except `id` (thus `name`, `username`, and `email`), being able to sign in with the same credentials regardless of `name` or `username` being updated. 
  - having to reconfirm email after PATCH'ing `username`.
  
**Weird Issue/Bug**
Then after PATCH'ing a new email and verifying it, I could still sign in with the old email and same password.
While simultaneously being able to sign in with the newly PATCH'ed email and originally POST'ed password.

image

I am not sure where this would come from. 

Please make an issue to clean up PATCH support for emails via the API and add a code comment to your Users endpoint stating this bug with PATCH'ing emails, then this will be good to merge.

abrahmasandra:
@majorsylvie Bug is documented in Issue #394 and code comment added to reference bug

src/chigame/api/views.py
```diff
@@ -56,6 +56,7 @@ class UserListView(generics.ListCreateAPIView):
     pagination_class = PageNumberPagination
 
 
+# Bug with PATCH'ing emails -- refer to Issue #394
```
majorsylvie, line 59: Thank you for the comment!

majorsylvie (approved):
Thank you for addressing the feedback and creating the new issue! 

Keep moving forward! Demo's almost here and quarter's almost done!

abrahmasandra:
Closing Comment: 

Updated the POST Users endpoint at `/api/users/` so that it correctly creates a new `User` model in the database. In order to do this, I needed to override the default method that the `UserSerializer` used for creating a new object, and instead needed to use the `create_user` method within the `UserManager` class.

Now, we can create new `User` objects along with inputting `passwords` through POST requests. In addition, a `username` attribute was added to the `User` class in order to act as `slugs` in other URLs

majorsylvie:
Issue Score: Excellent

Issue #243, opened 2023-11-16
Update Game Factory to have a unique category field for creation


majorsylvie:
Currently, the game factory can create models that have duplicates of certain fields such as categories.
However, there are uniqueness constraints on the game model. Which can cause integration tests to fail non-deterministically as sometimes random generation will not produce the same categories, but sometimes it will.

This task would involve updating the game factory to use some kind of method to generate attributes like categories in a way that maintains uniqueness. For example, and autoincrementing name like `Category 1` then `Category 2` etc...

frowenz:
Pull request, https://github.com/uchicago-cs/chigame/pull/236, has a fixture with all the categories and mechanics which could possibly be of use

cuyakonwu:
Just created Pull Request #299 to address this issue

abrahmasandra:
Closing Comment: 

This issue modified the `CategoryFactory` so that there is no chance of creating multiple `CategoryFactory` models with the same name during testing. Before, there were non-uniquely named `Category` objects being created, causing errors in testing. Now, Category factories are named using an `auto-incrementing` integer (Category 1, Category 2, etc.) so there is no chance of them having the same name. This change was implemented using the `Sequence` function within the `factory` module.

majorsylvie:
Issue Score: Excellent

Comments:
Great job! Thank you for fixing this bug and moving within the agile method.
@abrahmasandra this closing comment is particularly good! thank you for being very descriptive and fully describing the problem.

PR #299, opened 2023-11-27
Based on Issue #243, modify the Category Factory to prevent uniqueness errors in Continuous Integration

cuyakonwu:
Appropriately modify the GameFactory and CategoryFactory to incrementally create categories as opposed to randomly. The faker library sometimes ran into the issue of creating duplicate categories which would create an error as categories are considered unique. This will be circumvented using a lambda function to incrementally create categories which would ensure there are no duplicates.


src/chigame/api/tests/factories.py
```diff
@@ -1,5 +1,6 @@
 import random
 
+import factory
```
giomhern: Maybe for readibility sake, just have the imports into one line as in 
```
from factory import * 
```

src/chigame/api/tests/factories.py
```diff
@@ -10,7 +11,7 @@ class CategoryFactory(DjangoModelFactory):
     class Meta:
         model = Category
 
-    name = Faker("word")
```
giomhern, line 14: This is exactly what we needed! A sequence numerator that could autoincrement!

src/chigame/api/tests/factories.py
```diff
@@ -1,5 +1,6 @@
 import random
```
giomhern, line 1: Do we use the random import anywhere? Else, it is not really necessary!

giomhern (approved):
This seems to get the job done. Perhaps some extraneous code that could be removed but to address the issue, it looks like it is spot on and functional -- exactly how it is stated in the issue. I approve!

src/chigame/api/tests/factories.py
cuyakonwu: That raises several errors :(

src/chigame/api/tests/factories.py
cuyakonwu: Random is used in the game factory

src/chigame/api/tests/factories.py
giomhern: I thought it would be okay, but the in-place way works just fine! We might in the future want to make it cleaner (some way or another), but for now, looks good!

majorsylvie: in general it is a bad idea to import everything, since then it is much harder to tell what you are actually using from a given library.

It is better to do the work of writing out what you are doing (generally in alphabetical order) since then it is clear exactly what you are using from each file.

majorsylvie (changes requested):
Remove the `import factory` line and update your code to use the imports directly. 

Thank you for adding the autoincrementing

majorsylvie (approved):
With the imports standardized, this code is good to go!

abrahmasandra:
@majorsylvie I removed the `import factory` line and instead imported `Sequence` from the `factory` class individually

abrahmasandra:
Closing Comment:

This PR modified the `CategoryFactory` so that there is no chance of creating multiple `CategoryFactory` models with the same name during testing. Before, there were non-uniquely named `Category` objects being created, causing errors in testing. Now, Category factories are named using an `auto-incrementing` integer (Category 1, Category 2, etc.) so there is no chance of them having the same name. 

@cuyakonwu : wrote the lambda function to name using auto-incrementing integers
@abrahmasandra : reviewed code and fixed certain import statements

majorsylvie:
Issue Score: Excellent

Comments:
Thank you for the bugfix!
Very good closing statement which summarizes and explains the bug that this PR was fixing. 
I'm categorizing this as minor since the actual code changes for this is a few lines!

PR #309, opened 2023-11-28
Integrated forum search into search dropdown menu

### LINKED ISSUE #307: Add the forum search bar to the search bar options dropdown in the nav
The purpose of this task is to add the forums search bar to the search bar options dropdown found in the nav-bar. This will ensure that the forums can be searched even if a user is not on the forums part of the site.

Benjoman12:
Per issue #307 I have added the forum search option to the dropdown in the nav on the home page and have ensured that it actually searches.


src/templates/base.html
```diff
@@ -161,25 +162,23 @@
         document.addEventListener('DOMContentLoaded', function() {
           const search_bar = document.getElementById('search-bar');
           const query_type = document.getElementById('query-type');
-          const query_input = document.getElementById('query-input');
-
-          //do nothing if search bar is empty
-          search_bar.addEventListener("submit", function() {
```
joshuamonroe, line 173: This block of code needs to be preserved to maintain existing functionality that prevents searching when nothing is entered in the search bar.

src/templates/base.html
```diff
@@ -161,25 +162,23 @@
...
-          });
+          const query_input = document.getElementById('query');
+          const hidden_query_input = document.getElementById('hidden-query');
 
           query_type.addEventListener("change", function() {
-            if (query_type.value == "games") {
+            if (query_type.value === "games") {
               search_bar.action = "{% url 'game-search-results' %}";
-            } else if (query_type.value == "users") {
-              search_bar.action = "{% url 'users:user-search-results' %}";
```
joshuamonroe, line 184: This line needs to be preserved so as not to break existing user search functionality.

src/templates/base.html
```diff
@@ -111,9 +111,10 @@
                     <select id="query-type" name="query_type">
                       <option value="games">Games</option>
                       <option value="users">Users</option>
-                      <!-- Add more options here -->
+                      <option value="forums">Forums</option>
                     </select>
                     <input id="query" name="query" type="text" placeholder="Search..." />
+                    <input type="hidden" name="q" id="hidden-query" />
```
joshuamonroe: From what I can tell, you have included this "hidden-query" element because the django-machina search page is expecting to get the search query from an input element named "q",  while the the games search page is expecting to receive input from an element named "query".

It seems less than ideal to have this second hidden input and we should find a way to unify this into a single input element that can serve both the games and forums search. Although we could rewrite the django-machina search code to expect "query", it may actually end up being cleaner to go the other way and change the games search to expect input from "q" instead.

src/templates/base.html
```diff
@@ -111,9 +111,10 @@
                     <select id="query-type" name="query_type">
                       <option value="games">Games</option>
                       <option value="users">Users</option>
-                      <!-- Add more options here -->
+                      <option value="forums">Forums</option>
                     </select>
                     <input id="query" name="query" type="text" placeholder="Search..." />
```
joshuamonroe: I noticed that we broke games and user search in dev with PR #272 by changing the id of this element from "query-input" to "query", while "query-input" is still expected in the JS. We need to fix this and verify that all search modes are functioning before merging. My preference would be to revert to using "query-input" throughout, as it is a bit more descriptive.

src/templates/base.html
```diff
@@ -161,25 +162,23 @@
         document.addEventListener('DOMContentLoaded', function() {
           const search_bar = document.getElementById('search-bar');
           const query_type = document.getElementById('query-type');
-          const query_input = document.getElementById('query-input');
```
joshuamonroe, line 170: Should restore this line per comment on line 116 regarding broken search.

joshuamonroe (changes requested):
Found a few issues that need to be resolved before we merge this.

Firstly, now that forum search is integrated into the dropdown menu, I think it makes sense to have the same search bar across all pages, rather than changing it to a forum specific search on the forums page.

Please see code comments for remaining issues.

src/chigame/games/views.py
```diff
@@ -136,7 +136,8 @@ def test_func(self):
 
 
 def search_results(request):
-    query = request.GET.get("query")
+    # Must be "q" and not "query" as django machina framework expects a "q" input
```
joshuamonroe: While certainly valuable to justify these changes now, I'm a bit iffy on the value of these comments for future developers. Once you are no longer looking at a diff with the old line, the comment loses some of its meaning and might even be confusing. Perhaps these could be reworded or maybe it would be better to include a comment with the element in base.html instead, where the context will be a little bit more clear.

src/templates/base.html
```diff
@@ -163,21 +163,19 @@
           const query_type = document.getElementById('query-type');
           const query_input = document.getElementById('query-input');
 
-          //do nothing if search bar is empty
```
joshuamonroe, line 172: Assuming the code you have included below is functionally the same as this code, it would be good to retain this comment.

joshuamonroe (changes requested):
Looking much better. Although I do have a couple of very minor concerns related to comments. Also, I would still like to address this item from my previous review.
> Firstly, now that forum search is integrated into the dropdown menu, I think it makes sense to have the same search bar across all pages, rather than changing it to a forum specific search on the forums page.

joshuamonroe (approved):
Looking good now, approved to merge.

majorsylvie (changes requested):
code itself looks good and using the forum dropdown option takes me to the forum when i hit enter. But I don't understand how the actual search works! 

Searching for topic names, post names, individual letters, nothing worked and all returned 0 results. 

Please explain if / how the search works!
And resolve the merge conflict in game `views.py`

joshuamonroe:
Merge conflicts have been resolved. We are aware that search does not work currently and that is why #342 exists, to implement a working backend for search.

majorsylvie (approved):
Sounds good! #342 has been merged so the search backend you're using should now work! 

Clean work, keep moving forward!

Benjoman12:
All of the suggestions have been added, the forum search can now be found in the dropdown. The code found in this PR also lays the groundwork for how the search functionalities of any future frameworks incorporated into Chigame should be added to the base.html file. The PR has been successfully merged and closed. Thanks to @joshuamonroe  for reviewing the PR.

majorsylvie:
Issue Score: Excellent

PR #317, opened 2023-11-28
Add create page for forums

### LINKED ISSUE #292: Allow admins to create new forums on the forum landing page
# Goal
This issue seeks to allow creation of new forums, and this functionality should be made accessible on the forum landing page (`/forum`)

## Notes
- This functionality should only be available to admins
- Completion of this issue will require routing to a creation forum page

bburgess19:
# Goal
This PR seeks to add a creation page for new forums, closing #292 

## Notes
 - The creation page is accessible through a button on the forum landing page (`/forum`)
 - Completion of this required creation of a `forums` app within chigame which was, at the time of implementation, not yet created


src/chigame/forums/views.py
```diff
@@ -0,0 +1,15 @@
...
+from django.views.generic import CreateView
+from machina.apps.forum.models import Forum
+
+
+class ForumCreateView(LoginRequiredMixin, SuccessMessageMixin, CreateView):
+    model = Forum
+
+    success_message = "Forum successfully created!"
+    template_name = "forum_form.html"
+    fields = ["name", "slug", "description", "image", "type"]
```
giomhern: Is the slug here something based on the machina implementation and okay to be something inputted from the user? I imagine that the slug could be an autoincrementing attribute (or alike) rather than something provided from the user.

src/config/settings/base.py
```diff
@@ -198,6 +198,8 @@
         # https://docs.djangoproject.com/en/dev/ref/settings/#dirs
         "DIRS": [
             str(BASE_DIR / "templates"),
+            # Needed for django-machina to find the correct templates without cluttering the templates base directory
+            str(BASE_DIR / "templates/forum"),
```
giomhern: This looks like it's necessary and also good way to separate forum specific html files for the future too without clustering!

src/config/urls.py
```diff
@@ -21,6 +21,7 @@
     # Django-machina forum
     # https://django-machina.readthedocs.io/en/latest/getting_started.html#urls-configuration
     path("forum/", include(machina_urls)),
+    path("forum/", include("chigame.forums.urls")),
```
giomhern: Small issue of naming: should it be forums or just forum? As a user, somewhat confusing when I click on the "Forums" and then the naming isn't consistent. A very small thing, but could be of note in future.

src/templates/base.html
```diff
@@ -143,21 +152,14 @@
...
-              search_bar.submit();
-            }
-          });
 
           query_type.addEventListener("change", function() {
             if (query_type.value == "games") {
               search_bar.action = "{% url 'game-search-results' %}";
             } else if (query_type.value == "users") {
-              search_bar.action = "{% url 'users:user-search-results' %}";
+              search_bar.action = "{% url 'home' %}"; // placeholder
```
giomhern: Are the placeholders here necessary?

src/templates/forum/index.html
```diff
@@ -0,0 +1,6 @@
+{% extends 'machina/forum/index.html' %}
+
+{% block content %}
+  {{ block.super }}
+  <a class="btn btn-primary" href="{% url 'forums:forum-create' %}">Create new forum</a>
```
giomhern: The placement of the button is somewhat awkward here from a UX perspective. Perhaps putting it on the top right - ish area would make more sense.

src/templates/forum/board_base.html
```diff
@@ -0,0 +1,42 @@
...
+            {% if not request.user.is_anonymous %}
+              <a href="{% url 'forum_member:user_subscriptions' %}"
+                 class="d-inline-block ml-3"><i class="fas fa-bookmark">&nbsp;</i>{% trans "Subscriptions" %}</a>
+              <a href="{% url 'forum_member:user_posts' request.user.id %}"
+                 class="d-inline-block ml-3"><i class="fas fa-comments">&nbsp;</i>{% trans "View my posts" %}</a>
+            {% endif %}
+            {% get_permission 'can_access_moderation_queue' request.user as can_access_moderation_queue %}
+            {% if can_access_moderation_queue %}
+              <a href="{% url 'forum_moderation:queue' %}" class="d-inline-block ml-3"><i class="fas fa-gavel">&nbsp;</i>{% trans "Moderation queue" %}</a>
+            {% endif %}
```
giomhern: All the links are functional! Great stuff!

giomhern (commented):
I left a couple of comments on the model of the forums (about needing slug as a user input field) and some suggested UX changes to better accommodate a user. From a functionality perspective, it works great! Look over the suggestions and if any are necessary, let me know if you need assistance!

src/chigame/forums/views.py
bburgess19: The slug is from the machina implementation, and looking at their model doesn't have a `blank=null` field in the model, so I assumed they'd want that field to be non-optional on forms. Remember, the user here will always be an admin, so you can imagine they'd likely have some idea of how they'd want the url scheming/naming to work and thus customization over those urls.

Not fully sold on exposing it as a field, so let me know what you think!

src/config/urls.py
bburgess19: I agree, it should be called forums. I only kept it as that because the line before had "forum" as well. Will change.

src/templates/base.html
bburgess19: This is code from #309. I don't think the placeholder is necessary, so I will remove it

src/chigame/forums/views.py
giomhern: I think for the most part, if it's in the machina implementation, it doesn't cause any functional issues. We can leave as is and open an issue if we find that it is not necessary or we need a different way of employing slugs.

src/chigame/forums/views.py
```diff
@@ -0,0 +1,15 @@
...
+from django.views.generic import CreateView
+from machina.apps.forum.models import Forum
+
+
+class ForumCreateView(LoginRequiredMixin, SuccessMessageMixin, CreateView):
+    model = Forum
+
+    success_message = "Forum successfully created!"
+    template_name = "forum_form.html"
+    fields = ["name", "description", "image", "type"]
```
giomhern, line 13: This is good!

src/config/urls.py
```diff
@@ -20,7 +20,8 @@
     path("games/", include("chigame.games.urls")),
     # Django-machina forum
     # https://django-machina.readthedocs.io/en/latest/getting_started.html#urls-configuration
-    path("forum/", include(machina_urls)),
+    path("forums/", include(machina_urls)),
```
giomhern, line 24: Standardizing this is also good!

src/templates/forum/index.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'machina/forum/index.html' %}
+
+{% block content %}
+  <div class="d-flex justify-content-end">
+    <a class="btn btn-primary right-0"
+       href="{% url 'forums:forum-create' %}">Create new forum</a>
```
giomhern, line 6: This is much better placement!

giomhern (approved):
All of the suggested changes are well implemented! Looks very good to me after this second review.

src/config/urls.py
joshuamonroe: Be aware that making this change will require updating any existing references to the original URL scheme when  this is merged. As far as I can tell that is currently just [one line](https://github.com/uchicago-cs/chigame/blob/721ef602d22a7058f7f845665b2bb5759fd25307/src/templates/base.html#L72) though. 

Additionally, while it shouldn't be an issue, it's also worth keeping in mind that this scheme would no longer match the configuration given in the django-machina [installation instructions](https://django-machina.readthedocs.io/en/latest/getting_started.html#urls-configuration).

src/templates/forum/index.html
joshuamonroe: I think the placement of this button is fine, considering it's a separate admin function, but it's worth considering that it is inconsistent with the placement of similar buttons in the django-machina UI.
Screenshot 2023-11-30 at 11 07 36 AM
Screenshot 2023-11-30 at 11 05 55 AM

Also, the button is visible even when you are logged out. It should only be visible after you log in as an admin.
Screenshot 2023-11-30 at 11 09 01 AM

bburgess19: I could move it to the top left. That would make the most sense in my opinion. Also, thanks for pointing out the visibility issue. I'll change that.

src/config/urls.py
bburgess19: Thanks for the heads up. I was also concerned that the url change would cause downstream issues, but I haven't noticed any thus far.

majorsylvie: navbar. the navbar will no longer work if you change this URL since root forums URL in the `config/urls.py` is not namespaced:

```py
    # Django-machina forum
    # https://django-machina.readthedocs.io/en/latest/getting_started.html#urls-configuration
    path("forums/", include(machina_urls)),
    path("forums/", include("chigame.forums.urls")),
```

compared to: 
```py
    path("", TemplateView.as_view(template_name="pages/home.html"), name="home"),
    path("about/", TemplateView.as_view(template_name="pages/about.html"), name="about"),
```

src/chigame/forums/views.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def get_success_url(self):
+        return "/forums"
```
majorsylvie: Consider using `reverse("[url_name]")` such that if the forum URL changes later (eg: we add a prefix to all apps like `chigame.com/v2/forums`)

src/config/urls.py
```diff
@@ -20,7 +20,8 @@
...
+    path("forums/", include("chigame.forums.urls")),
```
majorsylvie, line 27: Why are there two lines for `path("forums/")`? Especially since the way django resolves URLs, anyone visiting `chigame/forum` will ALWAYS go to the top result.

To my understanding you did this such that you could have a `forums/create`. 
If this is how it you intend this URL scheme to work, please add verbose code comments explaining this, or else to someone who doesn't know about the setup of the forum it would seem redundant.

src/templates/base.html
majorsylvie: If this is unnecessary please remove!

src/templates/forum/index.html
majorsylvie: Two potential solutions could be emulating canvas:
image
Emulating canvas by placing it at the top near other management buttons
image

Or like reddit, adding a sidebar / moderation sidebar which has all moderation/admin specific actions. 
image

Thus you can keep the main page as your "for any user" and then all extra moderation functionality mentally and literally separate.

majorsylvie (changes requested):
Overall, happy to see this page made and it works fine.

I'd like responses to the comments I've put on the PR.

Then when it comes to improvements, they are UX related.
About button placement, I've already made a more detailed comment inside the PR.
Then i have another comment about this dropdown:
image
This is very terse. 
There is no information about what each of these categories or how they are different.
If a forum can only strictly be one type, it would be a lot better to have this as a single checkbox selection, where next to each selection there is a small blurb explaining what it is / why you'd use it:
image

That way it is immediately and consistently obvious without any previous knowledge or research necessary.

Overall, at least respond to the comments, and please make issues to:
- update the placement of the create button
- decide where to put moderation tools
- make forum type selection more readable

src/chigame/forums/views.py
bburgess19: I changed it to use `reverse` in the latest commit.

src/config/urls.py
bburgess19: I added comments to clarify that we need our own `forums/` urls/views.

bburgess19:
@majorsylvie Your suggestions were addressed in comments, and issues have been created as you've suggested:
- #378 
- #379 
- #382

majorsylvie (approved):
Thank you for your work, addressing the feedback (URLs comment being particularly important), and making the requested issues.

Keep moving forward and godspeed with your demo!

bburgess19:
# Closing comment
This branch added a button for admins to create new forums. The creation page redirects back to the forum landing page after successful completion.

There was an issue with the `static` tag not being loaded for `board_base.html` which prevented it from loading, but this has been addressed in #410

majorsylvie:
Issue Score: Excellent

PR #321, opened 2023-11-29
Added ability for `username` to act as identifier for API user detail endpoint in addition to primary key (`/api/users/<userID>` OR `/api/users/<username>`)

### LINKED ISSUE #320: Get user detail endpoint to use username as slug in addition to primary key (`/api/users/<userID>` OR `/api/users/<username>`)
**Goal:** In this issue, we want to allow users to access the Detail view of a `User` object using the user's `username` as a slug instead of always having to use the user's `id`. This way, it is easier for someone to use the API to get details of a specific user, as one does not need to know it's primary key in the model. 

We will support: `/api/users/<user_id>` AND `/api/users/<username>` to GET/PATCH/DELETE the Detail view of the `User` with the given `id` or `username`

abrahmasandra:
**Goal:** In this PR, we allow users of the API to access the Detail view of a given `User` object using the user's `username` as a slug instead of always having to use the user's `id`. For a user of the API to get a given `User's ID`, they need to go through the entire List view of user's and find the specific `User` they are looking for and then look at the `id` of that `User`. 

The `username` on the other hand, is a unique identifier for a `User` object, that is human readable and can be shared/accessed easily. This way, it is easier for someone that uses the API to get details of a specific user, as one does not need to know it's primary key in the model. 

We will support: `/api/users/<user_id>` AND `/api/users/<username>` to GET/PATCH/DELETE the Detail view of the `User` with the given `id` or `username`


src/chigame/api/views.py
cuyakonwu, line 1: What if a user goes by the name "1234"? That could cause an issue with searching for the correct user.   A better implementation could have you search through user name first and store usernames as "User#user_id" That way you can query id numbers by using the #, and we can exclude # from potential username characters.

cuyakonwu (changes requested):
Please see me comment on conflicting query options

src/chigame/api/views.py
abrahmasandra: I just made a commit that validates the User's `username`, and ensures that it cannot be fully numeric. This way, we will not have this problem, as a given URL parameter can only be interpreted as either a `username` OR an `id` but not BOTH. 

NOTE: This commit is in PR #291, because in that PR, I created the username field for the `User` model, so it is only consistent that I put restrictions on the `username` field in that PR.

cuyakonwu (approved):
Thanks for the changes!

src/chigame/api/views.py
majorsylvie: You must reach out to the users team before I will approve any changes to the `User` model.

src/chigame/api/views.py
```diff
@@ -47,6 +48,17 @@ class UserListView(generics.ListCreateAPIView):
 class UserDetailView(generics.RetrieveUpdateDestroyAPIView):
     queryset = User.objects.all()
     serializer_class = UserSerializer
+    lookup_field = "slug"
+
+    def get_object(self):
+        lookup_value = self.kwargs.get(self.lookup_field)
+
+        # If the lookup_value is an integer, use the id field
```
majorsylvie, line 69: Good code comment!

majorsylvie (commented):
per PR #291 , this PR involves changes that only fully work once the restriction on fully numeric usernames is merged to dev.

Until pr #291 is fully merged I will not merge this.

majorsylvie (approved):
This works as expected! 

I can POST users with unique usernames/email while having overlapping names, then go to `api/users/<username>` and get their detail view.

While still being able to access `api/users/<pk>`

Great work and keep moving forward! We're almost done!

majorsylvie:
Issue Score: Excellent

Comments:
Great job y'all! 
Appreciate the back and forth on review as well as linking relevant issues and problems as they come up.

abrahmasandra:
Closing Comment: Successfully, completed the `slug` functionality to get the detail view of a `User`. Can now access the detail view going to endpoint: `/api/users/<user_id>` OR `/api/users/<username>`.

PR #328, opened 2023-11-30
Create the Registration Deadline Feature for the Tournaments

### LINKED ISSUE #201: Create a registration deadline feature for tournament
Right now, players in a tournament are assigned to brackets right after the tournament is created. However, in a realistic setting, users should be allowed to sign up for the tournament and quit before the registration deadline. It is after the deadline that the brackets are created. We need to create this registration deadline feature accordingly. This change will affect the tournaments_brackets (# 189) and join_tournament (# 190) functionalities.

MikeChen012345:
This Pull Request addresses issue #201 to allow the registration deadline feature.
- The join_tournament functionality has been changed to only allow users to join and withdraw when the tournament status is "registration open". Also, the error message to reject join and withdraw actions after the status has changed is implemented.
- The tournament bracket is now created after the status is "registration close" and when the page is first accessed after the status has changed.
- A check_and_end_tournament function has been added to end the tournament when the tournament end date is reached, and matches that have not finished are automatically forfeited.
- "Join match" and "spectate" buttons have been added to the tournament detail view and are displayed when the status is "tournament in progress". However, since the related functionalities have not been implemented yet, the two buttons do nothing currently.


jb-01 (approved):
This worked great on my local machine! The Join and Withdraw buttons function without error and your `check_and_end_tournament` makes sense.

Great work!

src/chigame/games/views.py
```diff
@@ -231,6 +243,15 @@ class TournamentDetailView(DetailView):
     template_name = "tournaments/tournament_detail.html"
     context_object_name = "tournament"
 
+    def get(self, request, *args, **kwargs):
+        super().get(request, *args, **kwargs)
+        tournament = Tournament.objects.get(id=self.kwargs["pk"])
+        if tournament.matches.count() == 0 and tournament.status == "registration closed":
+            # if the tournament matches have not been created
+            tournament.create_tournaments_brackets()
```
elizabethli31, line 444: Something about this is messing up the ability to update. Because no matches are created in the registration period, I am unable to update the players in the tournament. Also, is it possible that the closing and tournament start happen at the same time? What would happen then?

src/chigame/games/views.py
```diff
@@ -257,8 +281,20 @@ def post(self, request, *args, **kwargs):
             elif success == 1:
                 messages.error(request, "You have not joined this tournament")
                 return redirect(reverse_lazy("tournament-detail", kwargs={"pk": tournament.pk}))
+            elif success == 3:
+                messages.error(request, "The registration period for this tournament has ended")
+                return redirect(reverse_lazy("tournament-detail", kwargs={"pk": tournament.pk}))
             else:
                 raise Exception("Invalid return value")
+
+        elif request.POST.get("action") == "join_match":
```
elizabethli31, line 483: I don't think it makes sense for this button to show up if the user is already in a match.

src/chigame/games/models.py
```diff
@@ -367,6 +371,24 @@ def set_archive(self, archive):
...
+        Any matches that have not finished are forfeited.
+        """
+        if self.status == "tournament ended":
+            brackets = self.matches.all()
+            for bracket in brackets:
+                assert isinstance(bracket, Match)
+                bracket_users = bracket.players.all()
+                bracket_players = [Player.objects.get(user=user, match=bracket) for user in bracket_users]
+                bracket_with_outcome = any(player.outcome is not None for player in bracket_players)
+                if not bracket_with_outcome:  # the match has not finished
```
elizabethli31, line 383: For forfeit option, maybe instead of just withdrawing from the match, you can also print something like "Match forfeited" because right now it just shows up as "No matches"

elizabethli31 (changes requested):
The functionalities listed in the description are all present and implemented very nicely! However, because the brackets aren't created until after the registration period is created, there are no matches in place, and this is impacting the update view. Please try to fix that in this PR.

src/chigame/games/views.py
MikeChen012345: Hi Elizabeth, for the closing and tournament start happening at the same time issue, lines 334, 338, 342 (in the clean() method of the Tournament model) have been changed to not allow these two to occur at the same time (which is allowed previously). Also, to avoid the situation in which no one accesses the tournament list view during the registration closed period, the line has been changed into `if tournament.matches.count() == 0 and (tournament.status == "registration closed" or tournament.status == "tournament in progress"):`, which will also create the matches during the tournament in progress stage.

For the issue with the update view, it seems that I am able to access the update view and change the attributes. Could you clone in a clean repository and try again? Thank you!

src/chigame/games/views.py
MikeChen012345: Hi Elizabeth, sorry for the ambiguity. This button on the tournament detail view is for the user to join a match (as opposed to joining a tournament during the registration open period), and the ideal behavior is to redirect the users to the pages where they can play a match with their opponents. The redirection part has not been implemented, so right now it is only a placeholder.

src/chigame/games/models.py
MikeChen012345: Hi Elizabeth, the matches field is implemented as such so that the behavior of the tournament detail view will not be too complicated. If all the matches that have been played in the tournament still need to be displayed, then it would be hard to keep track of the rounds in which the matches are played (currently the create_bracket and next_round methods remove all the matches played in the last round and create new matches for the next round). Therefore, I think that the matches field should be used to show all matches in that round instead of showing the status of some matches. 

Still, it is a great point that we should consider how to handle forfeited matches. An individual deadline for each round may be added later so that matches are forfeited in each round and do not impact the process of moving to the next round.

elizabethli31: Hey Mike, that logic makes sense to me. However, my point is mostly that the phrasing "no matches" is not intuitive and appears to be saying that this was a tournament without any matches at all (whether they were played or not). I think it should be changed to at least something like "matches forfeited" to indicate that there were in fact matches, but they were not played.

elizabethli31 (changes requested):
Now when I try to update the players, I am getting the 403 "You cannot add new players to the tournament after it has started." error even when the registration is open. This is because you are running `if len(form_players - current_players) ` in line 432 of TournamentUpdateView regardless of whether registration is open or not. To show you my testing: 
image
image
Looking at the tournament object for this specific pk also shows that registration is open, but I still cannot update.

MikeChen012345:
Hi Elizabeth, thanks for the review! I am not sure if it was me who wrote this part, but I have changed it to address the issue you mentioned. Also, matches are displayed as forfeited when no users are in the matches anymore (maybe because both have cheated). Thanks!

MikeChen012345:
Hi Elizabeth, the issue mentioned in the issue has been resolved. Thanks!

elizabethli31 (approved):
Thanks for fixing the update issue as well! Great work on this PR - the logic is sound and really elevates tournaments

MikeChen012345:
In this pull request, @MikeChen012345, @MuyanXie, and @rqueiroz20  worked on the code. The HTML parts and the views.py were mainly created by @MikeChen012345 and @MuyanXie, and @rqueiroz20  focused on different ways to implement the registration period feature.
@jb-01 helped us review the code and made sure that the code worked as expected. Thanks!

elizabethli31:
Issue Score: Excellent

Comments:
Nice work on this PR!

PR #329, opened 2023-11-30
Create the tournament Archived View Feature

### LINKED ISSUE #238: Add an archived view for archived tournaments
This issue would be implemented after issue #237 has been done. 
- Create an archived tournament list view to display all the archived tournaments. The information displayed on the archived tournament list view should be the same as that in the tournament list view, but the archived tournament list view should only display archived tournaments. 
- The `set_archive` function is called when an admin user clicks on an “archive” button on the tournament list view, detail view, and the archived tournament list view (displayed if 1. login as an admin and 2. Tournament has ended). (An unarchived button may be added if archived). (This button can be the "archive" star symbols.)
- Add a link of the archived tournament list view to the tournament list view, and vice versa (we can have a small toggle switch with labels showing “all” and “archived” to change the view (i.e., redirect to another view) displayed).

MikeChen012345:
These features have been implemented to address issue #238 :
- Create an archived tournament list view to display all the archived tournaments. The tournament information displayed on the archived tournament list view is the same as that in the tournament list view, but the archived tournament list view only displays archived tournaments. 
- The `set_archive` function is called when an admin user clicks on the “archive” star button on the tournament list view, detail view, and the archived tournament list view (displayed if 1. login as an admin and 2. Tournament has ended). If the tournament is archived, then the star is yellow, and clicking it results in unarchiving the tournament; if it is not archived, then the star is white, and clicking it results in archiving the tournament. If a non-admin user visits the page, the "archive" star icon is not a button, so it cannot be clicked.
- Add a button with the text "Switch to Archived Tournament List" to the tournament list view that, when clicked, redirects the user to the archived tournament list view, and vice versa "Switch to All Tournament List".


src/templates/tournaments/tournament_archived_list.html
```diff
@@ -0,0 +1,131 @@
...
+
+{% load static %}
+
+{% block extrahead %}
+  <link rel="stylesheet"
+        type="text/css"
+        href="{% static 'tournament_styles.css' %}" />
+{% endblock extrahead %}
+{% block content %}
+  <h1>Tournaments</h1>
```
rqueiroz20: I wonder if we should change this header to be "Archived Tournaments" or something similar. Right now, I almost got lost as to whether I was on the Tournaments list view or the Archived Tournaments list view.

src/templates/tournaments/tournament_list.html
rqueiroz20, line 1: This might be something to consider, but do we want archived tournaments to appear in the tournament list view as well as the archived tournament list view? Right now, they do, which I don't think is what we want.

rqueiroz20 (changes requested):
The archived tournaments feature looks good to me! I am able to archive a tournament that has ended and have it appear in the archived tournaments list view. Some things to consider:

- Do we want archived tournaments to be displayed on the tournament list view? I personally don't think so since the tournament would already be over by the time it is archived.
- The buttons to switch between the archived tournaments list view and the tournaments list view doesn't feel very good from a user perspective. The functionality works great, but I wonder if there's a better way to switch between the two views (perhaps some kind of filter feature?) more seamlessly.

Apart from that and some minor changes to the HTML files, great work! Re-assign me to the PR when the changes have been made (if any).

src/templates/tournaments/tournament_archived_list.html
MikeChen012345: Thanks for the comment! I will change this.

src/templates/tournaments/tournament_list.html
MikeChen012345: Hi Rafael, this is also something I am thinking about. I wonder whether the tournament list should display all tournaments that have been created or only tournaments that are ongoing or ended recently (meaning that they have a status other than "tournament ended" or the date is within a month after the tournament end date). If this is the case, all the tournaments that have ended a month before should be deleted unless they have been archived (and this is what "archive" means essentially). Currently, I don't know whether this would be a good idea. Therefore, I just leave it as it is and let the next generation decide. Thanks!

MikeChen012345:
Hi Rafael, for your second concern, I originally planned to use a nice-looking toggle switch to switch between these two views. However, it didn't work (and I don't know why), so I decided to implement the functionality first and let frontend people work on that later. Thanks!

rqueiroz20 (approved):
Nice job addressing my initial comments! Your explanation makes sense, and since I am also unsure of what we should do, I think we should document this question and leave it for future developers. The stylistic frontend design you mentioned also feels good to leave for future developers, but I'm glad that the functionality is there! Great work!

elizabethli31 (changes requested):
Eventually, I think it would make more sense to just use the same html base for both the list and archive, as they are virtually the same except the headers and buttons (which can be handled by an if statement). Also, I just realized that I am still able to update things (like the number of players) even though the tournament is over. Is it possible for you to add a check for that as part of this PR. Additionally, see my slack message for my other request.

MikeChen012345:
Hi Elizabeth, the archive view is implemented in this way in case we may want to add new features to the archive view. Therefore, if we use the same HTML file, then it would not be that flexible. For the second issue that you mentioned, I will add a check. Thanks!

MikeChen012345:
Hi Elizabeth, the check has been added to the update view. Thanks!

src/chigame/games/views.py
```diff
@@ -623,6 +623,10 @@ def form_valid(self, form):
         # Get the current tournament from the database
         current_tournament = get_object_or_404(Tournament, pk=self.kwargs["pk"])
 
+        # the tournament cannot be updated if it has ended
+        if current_tournament.status == "tournament ended":
+            raise PermissionDenied("You cannot update a tournament that has ended.")
```
elizabethli31: Is it possible for you to just do a redirect back to the detail view instead of the forbidden page and include this message?

elizabethli31 (changes requested):
See comment for minor additional change.

src/chigame/games/views.py
MikeChen012345: The issue has been addressed. Thanks!

elizabethli31 (approved):
LGTM!

MikeChen012345:
In this pull request, all the features mentioned in issue #238 have been implemented. However, there are some issues that need to be solved in the future.

Issues to be addressed in the future: 
Currently, I think this is part of a larger system that needs further consideration, so I put the issue into the next step part of the final report. Basically, the idea is that we need to design what the archive list represents. If it represents tournaments that are memorable, then all the tournaments that have ended shall be removed (and deleted) from the tournament list, and those that have been archived will be preserved and put into the archived list. If it represents other concepts, then a different action may need to be conducted.
I am not sure how to deal with tournaments that have ended, so I just keep all these ended tournaments, regardless of their archived status, in the tournament list and allow more flexibility later.

MikeChen012345:
In this pull request, @MikeChen012345 and @MuyanXie worked on creating the archived view and addressed feedback. @jb-01 helped with thinking about whether the archived view should be designed as a separate page or using the same template as the tournament list view.
@rqueiroz20 helped us with reviewing and testing the code. Thanks!

elizabethli31:
Issue Score: Excellent

Comments:
Sets the stage for future design choices of what an "archive view" looks like. Great work.

PR #331, opened 2023-11-30
Implement Chat Feed API and Frontend Chat Feed Features

### LINKED ISSUE #229: Add a update service to update the clients with latest changes in messages in a chat
In this issue, we wish to provide a way for the frontend to effectively receive updates from the backend and efficiently update the display to the players, with minimum requirements of reload

We are going to write a view, which upon query will give updates from the token_id held by the frontend, including all the updates to messages as well as new messages. Please notice once a message entry is posted into the database, it will not be modified in any way; Instead, we will use later messages to update or delete it if necessary

Token ID combined with Update On is a new design and should be updated fields to the model. Token ID is designed to reduce the payload in each query and to effectively track changes, with Update On is to effectively keep track of changes to the msg.

**When a msg is created**:
1. A unique PK is created
2. A unique Token ID is created, which is incremental to this specific chat
3. Update On is null, as we are not updating any previous code in this request

**When a msg is updated**:
1. A new msg is created with a new unique PK and unique and incremental Token ID
2. The Update_On will be set to the old message it is updating to

**When a msg is deleted**:
1. A new msg is created with a new unique PK and unique and incremental Token ID
2. The Update_On will be set to the old message it is updating to
3. Content is set to null

**[IMPORTANT] When the user request to read the messages**:
1. The user will send the latest token ID it has from the front end
2. The server will return it with the messages having the token ID greater than the current token ID the frontend has back
3. The frontend can then update the views based on the results from the payload - it can update previous histories of messages if necessary and can add new messages, all dependent on the Update_On field and Content field of the payload entries

By doing so, we will save a lot of bandwidth and provide a feasible way to real-time update and send messages. To be honest, this is pretty complex and need to be reviewed heavily.

References: Adopted some [log update logic](https://levelup.gitconnected.com/logs-and-distributed-systems-ebd3902732bd) from the design of distributed database systems

MuyanXie:
This PR is the final in the Chat feature series. Huge appreciation here for those who generously helped realize this feature! In this PR, we will address Issue #229. 

TODO:
1. Implement the update feed backend API for message
2. Implement an API test feature
3. Implement a fixture for testing purposes
4. Furnish the frontend to enable the send message logical flow
5. Furnish the frontend to enable the delete message logical flow


MuyanXie:
## Testing Suggestions to Reviewers
1. Use pytest to see if the update feed API is working properly
2. Play with the frontend with send/delete feature to see it works
3. You can actually open one tab in chrome and one in safari to chat with yourself

## Using fixture

if you wish to use fixture to test this implementation [__highly recommended__], you should first run:

```
python3 manage.py loaddata tournaments-chat-general.json
```

Note: To load a fixture, the best practice is to delete your current db, remigrate, and then load fixture

After this, go to :

```
http://127.0.0.1:8000/games/tournaments/1/chat/
```

And then play with it! In this chat, there are two players in the chat; You can login using admin in one browser and using test1 in another browser to play it now [I recommend using Safari and Chrome to avoid messing the cookies up]:

Credentials:
email: admin@example.com PWD: pass0000
email: test1@test.com PWD: pass0000

MikeChen012345:
Everything looks good on my end! I tried to use two browsers and tested the chat. One time I used the fixture with the admin and test1 accounts provided above, and I could send messages, receive messages from the other user, and delete my own messages (which was reflected on the end of the other user). The other time I created my own tournament through the create view and signed up two users, and it seemed that it worked fine as well. The following image was the chat history that I used in my first test with fixture:
2699b64e2cc77fd6e933174f936eb59

MikeChen012345 (approved):
Everything looks good on my end! I tried to use two browsers and tested the chat. One time I used the fixture with the admin and test1 accounts provided above, and I could send messages, receive messages from the other user, and delete my own messages (which was reflected on the end of the other user). The other time I created my own tournament through the create view and signed up two users, and it seemed that it worked fine as well. The following image was the chat history that I used in my first test with the fixture:
2699b64e2cc77fd6e933174f936eb59

src/static/js/getChatUpdate.js
```diff
@@ -71,36 +60,36 @@ function updateView(tokenId) {
...
 }
 
 function getMessageText(message) {
   if (message.content !== null) {
     return message.sender + ": " + message.content;
   } else {
-    return (
-      message.sender + ": " + `Message ${message.tokenId} deleted`
-    );
+    return message.sender + ": " + `Message ${message.token_id} deleted`;
```
elizabethli31: From a user perspective, it's a little confusing to have the token_id here. Deleting out of order will give something like this which is not intuitive as it seems like the "message 9 deleted" should be message 2
image

elizabethli31 (changes requested):
This is so cool, and the live chat works very nicely!! I left a minor comment on a stylistic choice. Also, should the chat be visible to someone who isn't logged in? Please also resolve merge conflicts for the next commit

src/static/js/getChatUpdate.js
MuyanXie: Thanks for pointing that out! I've added made the delete message like "Admin: Message Deleted" without the token id here in the newest PR

MuyanXie:
Hi Elizabeth! Thanks for reviewing. I have addressed the concerns you pointed out in the newest commit. Please let me know if you more questions!

elizabethli31 (changes requested):
Thanks for fixing the errors I pointed out! One more minor change I'd like is for there to be a redirection page if a chat object has not yet been created. I know you have a PR open to automatically create one when a tournament is created, but it's still good practice to handle the error in case something goes wrong with the creation. Right now, I'm able to go to tournaments/pk/chat even if that tournament does not have a chat object. It just doesn't allow me to send any messages. Please redirect the user back to the tournamnent details page with an error saying something like "chat does not exist" when this happens, so they aren't able to access that page.

MuyanXie:
Thanks for pointing that out! I have made the changes accordingly, and please let us know if there are other problems!

elizabethli31 (approved):
Great work finishing up the last PR in the chat series! This was a huge endeavor to create a live chat, and it's working so well. This PR looks good to go, so I will approve and merge. But please make note in your final report and some Wiki documentation about the mock chat that is currently still in place (for tournament 1). I'm fine with it being there for the purposes of your demo, but it shouldn't in the future.

elizabethli31:
Actually sorry ignore my comment about the mock messages - I just realized those were part of your fixture. Great work!

MuyanXie:
Closing Statement:

@MuyanXie worked on coding the frontend and API. @MikeChen012345 worked on reviewing the code and pointing out inefficiencies. @rqueiroz20 worked on peer programming with Sam on the API.

elizabethli31:
Issue Score: Excellent

Comments:
Nice work again on this massive feature. Super cool seeing it work - you guys did a great job!

PR #77, opened 2023-10-30
Created API endpoint for GET request (list view) for USERS with proper JWT authentication

Esterello2:
Modified views.py, urls.py, and serialization.py to establish endpoint for GET User (List View) with JWT authentication. This included adding the simpleJWT module from rest framework, along with JWT token built-in API views to allow the for the generation and refresh of tokens.
A POST User request API endpoint was also created here, which was also originally meant to deal with JWT authentication, however, team decided to make it an open endpoint.

PR #335, opened 2023-11-30
Implementing test cases to GET and DELETE user models at the api/users endpoint

cuyakonwu:
This pull request addresses the need for testing of the User API endpoints, specifically focusing on the GET and DELETE functionalities. The introduced tests ensure the reliability and correctness of these critical features, contributing to the overall stability and maintainability of Chigame.


src/chigame/api/tests/test_api.py
```diff
@@ -282,3 +282,26 @@ def test_delete_message(self):
...
+    def test_user_get(self):
+        user = UserFactory()
+
+        list_url = reverse("api-user-list")
+        detail_url = reverse("api-user-detail", kwargs={"pk": user.id})
+
+        list_response = self.client.get(list_url)
+        assert list_response.status_code == 200
+
+        detail_response = self.client.get(detail_url)
```
giomhern, line 393: Perhaps compare fields here to ensure that the response and the user (from the factory) match!

src/chigame/api/tests/test_api.py
```diff
@@ -282,3 +282,26 @@ def test_delete_message(self):
...
+        assert detail_response.status_code == 200
+
+    def test_user_delete(self):
+        user = UserFactory()
+
+        url = reverse("api-user-detail", args=[user.id])
+        response = self.client.delete(url, format="json")
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
```
giomhern, line 403: This looks good to me! Perhaps in this case, have an assert to ensure that a user was made in the first place and then the count goes to 0 after its deletion!

giomhern (changes requested):
I left some comments for what would make the two tests more "punishing" and strict to make sure we don't presume anything!

src/chigame/api/tests/test_api.py
cuyakonwu: There would be no simple way to do that as the faker is random, I will fix that once I implement the patch test!

src/chigame/api/tests/test_api.py
```diff
@@ -299,6 +299,7 @@ def test_user_get(self):
 
     def test_user_delete(self):
         user = UserFactory()
+        self.assertEqual(User.objects.count(), 1)
```
giomhern, line 398: This looks good to me!

giomhern (approved):
I am approving this! Looks good to me to ensure that its created and then deleted!

majorsylvie:
I will hold off on merging this until the JWT user `GET`ing PR gets merged #77 . Since that will change how `GET`ing users behaves.

PR #337, opened 2023-11-30
Fixed the bug in NotificationQuerySet.get_by_actor() and added tests

### LINKED ISSUE #333: NotificationQuerySet.get_by_actor() calls a non-existent method
While working on #332, we discovered that  in NotificationQuerySet.get_by_actor(),  get() is used to retrieve a Notification and then is_not_deleted() is called on the return value of the get() which is a Notification object while is_not_deleted() is a method of NotificationQuerySet not Notification.

omarelkashef:
Fixed the bug where in NotificationQuerySet.get_by_actor(), get() is used to retrieve a Notification and then is_not_deleted() is called on the return value of the get() which is a Notification object while is_not_deleted() is a method of NotificationQuerySet not Notification. I added conditional logic that raises an exception if no notification matching criteria was found and also added tests that cover the fixed bug.

### DIFF
--- a/src/chigame/users/models.py
+++ b/src/chigame/users/models.py
@@ -147,10 +147,10 @@ class NotificationQuerySet(models.QuerySet):
         try:
             actor_content_type = ContentType.objects.get(model=actor._meta.model_name)
             actor_object_id = actor.pk
-            queryset = self.get(actor_content_type=actor_content_type, actor_object_id=actor_object_id, **kwargs)
-            if not include_deleted:
-                queryset = queryset.is_not_deleted()
-            return queryset
+            notification = self.get(actor_content_type=actor_content_type, actor_object_id=actor_object_id, **kwargs)
+            if not include_deleted and not notification.visible:
+                raise Notification.DoesNotExist
+            return notification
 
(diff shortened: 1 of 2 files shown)


24raniwalar (approved):
LGTM! Nice catch. I like how you added not notification.visible too. Since it is a get(), I also like how the return type has changed to just one notification rather than a queryset.

elizabethli31 (changes requested):
When I create a notification through the server and manually test with include_deleted=False and visible=True, I get the DNE error along with an attribute error. 
image
It works fine to actually get the notification when both fields are true, but I can't find where this error is coming from or if it matters. If it's just a case that shows up when manually testing, I can approve this PR.

elizabethli31 (approved):
Nice job fixing the tests as well. LGTM!

omarelkashef:
Closing Statement: the bug is fixed by using the appropriate method and now it raises an exception if no notification matches the criteria. Tests were added to cover this logic.

elizabethli31:
Issue Score: Excellent

Comments:
Same as issue comment - I'm marking this as minor, but great job!

PR #341, opened 2023-12-01
Add 'Like' Functionality to Forum Posts

### LINKED ISSUE #253: Create Voting Functionality On Posts With Cumulative Counter
# Goal
Implement a voting feature for posts in our application, allowing users to express appreciation or disapproval for content. 

## Notes
Each post should display a cumulative count of likes and dislikes it has received. The resulting displayed number will be the sum total of the likes and dislikes.

giomhern:
# Purpose

This pull request seeks to add a voting features (like/dislike) to posts. Upon completion, this resolves #253.


bburgess19:
I had to force push a rollback onto the previous merge with `dev` because ed13665 deleted the code which solved issue #292.

src/chigame/forums/forum_conversation/models.py
```diff
@@ -0,0 +1,32 @@
...
+
+    poster = models.ForeignKey(User, on_delete=models.CASCADE, null=True)
+    post = models.ForeignKey(Post, on_delete=models.CASCADE, null=True)
+    rating = models.IntegerField(choices=Rating.choices)
+
+    class Meta:
+        constraints = [models.UniqueConstraint(fields=["poster", "post"], name="unique_rating")]
+
+
+from machina.apps.forum_conversation.models import *  # noqa
```
joshuamonroe, line 35: This import statement probably shouldn't be at the end of the file like this. Also, I have no idea what 'noqa' means, so that comment should probably be made clearer if it's important.

src/templates/forum_conversation/topic_detail.html
joshuamonroe, line 1: Would be good to include a comment at the top of this file like we did in board_base.html to link to the original template for reference purposes.
```
<!--  This file was derived from the following django-machina template.
https://github.com/ellmetha/django-machina/blob/main/machina/templates/machina/forum_conversation/topic_detail.html-->
```

src/chigame/forums/forum_conversation/views.py
```diff
@@ -0,0 +1,71 @@
...
+from .models import Vote
+
+Post = get_model("forum_conversation", "Post")
+
+
+class TopicView(BaseTopicView):
+    def post(self, request, **kwargs):
+        """Handles POST requests."""
+
+        # Retrieve the liked post
```
joshuamonroe, line 18: Appreciate these code comments for giving a high-level indicator of what is going on.

src/config/settings/base.py
```diff
@@ -109,6 +108,9 @@
     "chigame.games",
     # Additional apps go here
     "chigame.api",
+    "chigame.forums.base",
+    # Overridden django machina apps
```
joshuamonroe: Very minor point, almost not worth mentioning, but should be django-machina.

joshuamonroe (changes requested):
Good work getting the backend in place for being able to like/dislike posts. I made comments on a few things that I think need to get resolved before merging, mostly relating to code comments and an import, so please check those out.

I think we've still got a lot of work to do on the UI though. First off all, the implementation is looking downright unresolved right now, no two ways around that. But I realize that you already have another [issue](https://github.com/uchicago-cs/chigame/issues/360) open to address this, so I won't worry too much about it. One thing that I am concerned about from a functionality perspective though, is that it is not clear at all whether you have already liked or disliked a post.

For example, in this screenshot, it appears that I've neither liked nor disliked the post. However, it is actually the case that another user had already liked the post and I then disliked it, cancelling out the other vote.
Screenshot 2023-12-05 at 11 18 37 AM

Whereas, in this example, I've actually not voted.
Screenshot 2023-12-05 at 11 18 04 AM

And in this one I've liked the post.
Screenshot 2023-12-05 at 11 18 47 AM

So, I hope you can see that it's really not very clear which option has been chosen from the total vote count alone, except after trial and error. I think if there was some sort of highlighting going on to indicate which option has been selected, things would be much clearer.

Another area that I think could be improved in a future issue is the addition of like counts to the topics view, something like this.
Screenshot 2023-12-05 at 11 33 04 AM

src/templates/forum_conversation/topic_detail.html
bburgess19: Great idea, this is now included in the newest commit.

src/chigame/forums/forum_conversation/models.py
bburgess19: I added documentation for clarification.

TLDR: 'noqa' turns of the linter for the line since the import is at the bottom. It has to be at the bottom per django-machina instructions on overriding models.

src/config/settings/base.py
bburgess19: Good catch. Fixed in the latest commit.

src/chigame/forums/forum_conversation/models.py
joshuamonroe: Ok, that makes sense and I appreciate the clarifying comment with link to the relevant documentation.

bburgess19:
Thank you for reviewing thus far! I've created #415 to address the like counter being included in the forum list bar. Otherwise, see the comment threads for specifics.

joshuamonroe (approved):
Looking good now, thanks for addressing everything in the previous review. Approved!

majorsylvie (approved):
I can successfully like and dislike comments only once, and they're saved across reloads and server restarts! 

Great functionality, happy to see it implemented.

Thank you for adding a good amount of comments, particularly in your templates.

bburgess19:
# Closing Comment

This PR implemented voting functionality for posts

## Detailed Notes

This involved creating a new `Vote` model and extending the `django-machina` `Post` model to include a through table between `User` and `Post`.

Extending `Post` involved creating a new `forum_conversation` app and overriding the `django-machina` implementation. We didn't want to inundate the `/src/chigame/` directory with many `django-machina` app overrides, so we decided to move the previous `forum` app we had into a new `/src/chigame/forum/` directory which will contain our own forum code along with any `django-machina` related overrides. We also renamed the `forum` app we had previously to `base` to make it clear which code was entirely ours vs `django-machina`'s.

### Errors found during review

#### No visual indicator of which vote button was clicked _(credits: @joshuamonroe)_

Added a highlight to the button to make the selection clear

#### Lack of documentation for template and model code _(credits: @joshuamonroe)_

Included links to the relevant `django-machina` code and documentation.

Thank you to @joshuamonroe and @majorsylvie for reviewing this PR!

majorsylvie:
Issue Score: Excellent

PR #343, opened 2023-12-01
Add Factory for Lobby Model in Django Tests

cuyakonwu:
This PR introduces a factory for the Lobby model in our Django tests using the factory_boy library. The factory includes default values and support for creating instances with associated users as members.


src/chigame/api/tests/factories.py
```diff
@@ -106,3 +106,18 @@ class Meta:
...
+class LobbyFactory(DjangoModelFactory):
+    class Meta:
+        model = Lobby
+
+    match_status = Lobby.Lobbied
+    name = Faker("word")
+    game = SubFactory(GameFactory)
+    game_mod_status = Lobby.Default_game
+    created_by = SubFactory(UserFactory)
+    min_players = 2
```
giomhern: I'm not fully convinced that we have set values for these attributes. What if we wanted to test the creation of other factories that don't fit into this. An alternate could be employing lazy attributes as below:

```
min_players = factory.LazyAttribute(lambda x: random.randint(2, 6))
```

src/chigame/api/tests/factories.py
```diff
@@ -106,3 +106,18 @@ class Meta:
...
 
     tournament = SubFactory(TournamentFactory)
+
+
+class LobbyFactory(DjangoModelFactory):
+    class Meta:
+        model = Lobby
+
+    match_status = Lobby.Lobbied
+    name = Faker("word")
```
giomhern: We could use a sequence here from factory boy to denote different lobbies to not run into uniqueness issues from another PR.

giomhern (changes requested):
I suggest changing the set values of this factory in the case that we want to test a varying number of these attributes to make sure it is universally functional.

src/chigame/api/tests/factories.py
```diff
@@ -106,3 +106,18 @@ class Meta:
         model = Chat
 
     tournament = SubFactory(TournamentFactory)
+
+
+class LobbyFactory(DjangoModelFactory):
+    class Meta:
+        model = Lobby
+
+    match_status = Lobby.Lobbied
```
majorsylvie: Same logic that Gio mentions extends here, it should be able to make lobbies with one of the three status' in the lobby model:

```py
class Lobby(models.Model):
    """
    A lobby that users can join before starting a match.
    The match can start as soon as min_players join the lobby.
    """

    Lobbied = 1
    Viewable = 2
    Finished = 3
    STATUS = ((Lobbied, "Lobbied"), (Viewable, "In-Progress"), (Finished, "Finished"))

    Default_game = 1
    Modified_game = 2
    MODS = ((Default_game, "Default Game"), (Modified_game, "Modified Game"))

    match_status = models.PositiveSmallIntegerField(choices=STATUS, default=1)
```

src/chigame/api/tests/factories.py
majorsylvie: Yes exactly thank you for catching this Gio

src/chigame/api/tests/factories.py
```diff
@@ -106,3 +106,18 @@ class Meta:
...
+    name = Faker("word")
+    game = SubFactory(GameFactory)
+    game_mod_status = Lobby.Default_game
```
majorsylvie: This should also be allowed to be one of the two game options (default / unique)

```py
class Lobby(models.Model):
    """
    A lobby that users can join before starting a match.
    The match can start as soon as min_players join the lobby.
    """

    Lobbied = 1
    Viewable = 2
    Finished = 3
    STATUS = ((Lobbied, "Lobbied"), (Viewable, "In-Progress"), (Finished, "Finished"))

    Default_game = 1
    Modified_game = 2
    MODS = ((Default_game, "Default Game"), (Modified_game, "Modified Game"))

    match_status = models.PositiveSmallIntegerField(choices=STATUS, default=1)
```

src/chigame/api/tests/factories.py
```diff
@@ -106,3 +106,18 @@ class Meta:
...
+    created_by = SubFactory(UserFactory)
+    min_players = 2
+    max_players = 4
```
majorsylvie: When converting `min_players` to a `LazyAttribute`, change this to another `LazyAttribute` which uses the values of `min_players` such that there will never be a sitation where `min_players > max_players`

src/chigame/api/tests/factories.py
```diff
@@ -106,3 +106,18 @@ class Meta:
...
+    time_constraint = 300
```
majorsylvie: This also should vary, with 300 being an acceptable value.

majorsylvie (changes requested):
Overall this is a good idea but needs to be fleshed out.
Follow the comments Gio and I have made and it should result in a significantly more generalizable model that is appropriate as the basis for tests.

src/chigame/api/tests/factories.py
```diff
@@ -1,17 +1,18 @@
 import random
 
-from factory import Faker, LazyAttribute, Sequence, SubFactory, post_generation
+from django.utils import timezone
+from factory import Faker, Iterator, LazyAttribute, LazyFunction, Sequence, SubFactory, post_generation
 from factory.django import DjangoModelFactory
 
-from chigame.games.models import Category, Chat, Game, Mechanic, Tournament
+from chigame.games.models import Category, Chat, Game, Lobby, Mechanic, Tournament
```
giomhern, line 7: Good catch on the missing imports!

src/chigame/api/tests/factories.py
```diff
@@ -106,3 +107,23 @@ class Meta:
...
+    match_status = Iterator([Lobby.Lobbied, Lobby.Viewable, Lobby.Finished])
+
+    name = Sequence(lambda n: f"lobby_{n}")
+    game = SubFactory(GameFactory)
+
+    game_mod_status = Iterator([Lobby.Default_game, Lobby.Modified_game])
+
+    created_by = SubFactory(UserFactory)
+    min_players = LazyAttribute(lambda x: random.randint(2, 6))
+    max_players = LazyAttribute(lambda o: random.randint(o.min_players, 10))
```
giomhern, line 125: This is good to ensure that the min_players doesn't exceed the max_players.

giomhern (approved):
This factory looks great now! The sequences and random ints are necessary! Good job Conrad!

cuyakonwu:
Included lazy attributes to resolve the issue of randomness with min and max players, randomized the selection of lobby types, added an iterator to move through different lobby statuses, randomized date created by in this decade, along with a randomizer within a range for minimum play time.

src/chigame/api/tests/factories.py
```diff
@@ -35,8 +36,8 @@ class Meta:
 
     min_players = Faker("pyint", min_value=1, max_value=10)
 
-    # LazyAttribute allows setting a field's value based on other fields at runtime.
-    # In this case, we want max_players to be at least min_players, but no more than 10.
+    # LazyAttribute allows setting a field's value based on other fields at runtime
+    # In this case, we want max_players to be at least min_players, but no more than 10
     max_players = LazyAttribute(lambda x: random.randint(x.min_players, 10))
```
majorsylvie, line 41: Nice code comment!

src/chigame/api/tests/factories.py
```diff
@@ -106,3 +107,23 @@ class Meta:
...
+class LobbyFactory(DjangoModelFactory):
+    class Meta:
+        model = Lobby
+
+    match_status = Iterator([Lobby.Lobbied, Lobby.Viewable, Lobby.Finished])
+
+    name = Sequence(lambda n: f"lobby_{n}")
+    game = SubFactory(GameFactory)
+
+    game_mod_status = Iterator([Lobby.Default_game, Lobby.Modified_game])
```
majorsylvie, line 121: Great use of iterators! Thank you for taking the feedback

majorsylvie (approved):
Great job generalizing and adding code comments! This is now much more representative of the range of factories.

I've reached out to the games team on slack if there is a reasonable min or max for min/max players. 

Based on that you should update the ranges for `min_players` and `max_players` to be representative of the limits of the field instead of chosen arbitrarily. 

However that can be a separate minor PR, this code is good to merge!

giomhern:
Closing this issue! Possible issue and minor PR to be opened in the future in regard to the min and max players. This is perfect to test the Lobby models. Thanks to @cuyakonwu for coding up a storm and implementation and to @giomhern for reviewing and suggesting changes.

majorsylvie:
Issue Score: Excellent

Comments:
Great work y'all!

Happy to see this functionality, good PR reviews, then taking and implementing the feedback!

Keep moving forward!

PR #344, opened 2023-12-01
Created a deleted notifications page where users can look at deleted notifications and move them back to the inbox 

### LINKED ISSUE #339: Add Junk box for deleted Notifications
Currently you can delete notifications on the admin page. They're marked as deleted but they're not actually removed (visible attribute is set to false). Add a filtering feature where the visibility is set to false.

henrietta-k:
Addresses issue #339 

In the user's inbox, they can now click on a button called "Show Deleted Notifications" which will take them to a page displaying all their deleted notifications. On this page, for each notification, they have the option to move it back to their main inbox, which will mark the notification as unread. A new HTML template page was created to show the deleted notifications, and it can be accessed from the user's main inbox. On the deleted notifications page, the user also has the option to return to their main inbox through a "Back" button.


24raniwalar (approved):
The addition of self.visible makes sense as unread notifications need to be visible again. Nice protection against viewing notifications that are not yours in notification_detail and act_on_inbox notification. My favorite part is the JS script as the end of user_inbox.html to change the style of the buttons. Tested in local environment and works well.

24raniwalar (changes requested):
Sorry, mistaken earlier - not sure deleted_notifications.html has been added so I think this needs to be added as the delete button does not currently lead to a valid page.

henrietta-k:
> Sorry, mistaken earlier - not sure deleted_notifications.html has been added so I think this needs to be added as the delete button does not currently lead to a valid page.

Forgot to git add that earlier on. Just pushed the deleted_notifications.html file. Thanks!

24raniwalar (changes requested):
I think the template is in the wrong directory. The deleted_notifications.html template should be in src/templates/users as user_inbox.html is.

henrietta-k:
> I think the template is in the wrong directory. The deleted_notifications.html template should be in src/templates/users as user_inbox.html is.

Sorry about that! For some reason there were two copies of the template in my local repository and I pushed the wrong one. I moved the deleted_notifications.html file to the correct folder now.

24raniwalar (approved):
LGTM! Nice patch on the Move to Inbox button as well, I wasn't sure if we were just going to address that in another PR so way to handle it here.

elizabethli31 (approved):
Ideally the other commits from dev should not be in your PR, but it doesn't seem like that's broken anything. Junk box looks good to go!

henrietta-k:
Closing statement:
The issue has been addressed. In the user's inbox, they can now choose to see all their deleted notifications and move it back into their main inbox (marked as unread) if they want. All the assignees worked together to write the code but I was in charge of making the GitHub PR.

elizabethli31:
Issue Score: Excellent

Comments:
Nice work on this PR!

PR #345, opened 2023-12-01
Integration of Placeholder Game + Match Start Conditions Implemented

### LINKED ISSUE #222: Add functionality to keep track of changing attributes (Lobby timer and number of players) in real time
**Goal:** To have a transition from the lobby page to the match page, we can either start a match automatically once the timer runs out (given the lobby has the minimum number of people), start a match automatically once the lobby reaches the maximum number of people, or give users to a button to start a match once the lobby has the minimum number of people. We want to implement functions that allow us to keep track of the time and keep track of the number of people in the lobby in real-time so that users do not have to refresh the page to see the changes.

### LINKED ISSUE #224: Add functionality to trigger the start of a match once conditions are met for a lobby
**Goal:** To have a transition from the lobby page to the match page, we can either start a match automatically once the timer runs out (given the lobby has the minimum number of people), start a match automatically once the lobby reaches the maximum number of people, or give users to a button to start a match once the lobby has the minimum number of people. We want to implement functions that allow us to automatically redirect players when a condition is reached and enable a was-greyed-out "Start" button when a condition is reached.

### LINKED ISSUE #268: create a dummy game

### LINKED ISSUE #269: Integrate Placeholder Game
This task is concerned with getting the placeholder game to occur in a lobby that meets start conditions. This relies most directly on #266, our placeholder game, and #224, match start conditions. Some match behavior regarding how multiple user lobbies are created, needs to be ironed out as well.

nugerhard:
This is a 2 (probably 3) issue pr: Essentially, what I've done is I have made the placeholder game accessible from any particular lobby page. But, it only becomes accessible once the lobby conditions have been met. To this end, I've integrated a timer related to the lobby model onto the lobby detail page using an AJAX function: once this timer has expired, and sufficient users are in the lobby, the lobby allows the user to play the game. More notes to come, likely on the relevant issues.


nugerhard:
Apologies, added some small last-minute stuff: when the timer ran out, page did not reflect changes until it was reloaded. So, when the update function is now called, the page is refreshed. Also, did a bit of shuffling of the html on the lobby details page to make it look nicer.

jiyeongchloe (approved):
Everything works as expected! Good job implementing the match status change that we talked about and integrating the dummy game! We might want to make a new issue for future teams for making sure that the game link corresponds with the user's game of choice.

src/templates/games/game_coinresult.html
```diff
@@ -13,4 +13,7 @@ <h1>Coin Flip Result</h1>
   {% else %}
     <p>Sorry, your guess was incorrect.</p>
   {% endif %}
+  <p>
+    <a href="{% url 'lobby-list' %}">Back to Lobbies</a>
```
elizabethli31: Intuitively, it makes more sense for this to redirect back to the lobby detail page that the game was started from. I think this would be easier for your demo too, so that you can see the match status after the game has been played.

src/chigame/games/views.py
```diff
@@ -407,6 +423,11 @@ def check_guess(request, pk):
         player.outcome = Player.LOSE
     player.save()
 
+    # Checks if everyone has played
+    if match.players.all().count() == lobby.members.all().count():
```
elizabethli31, line 602: Logic is a little too simple right now, and I don't love that this automatically leads to the "match is finished or cancelled" option. I don't think it would be too difficult to check separately if the lobby members count is less than min players just to distinguish the match status comment a little.

src/templates/games/lobby_details.html
```diff
@@ -1,34 +1,132 @@
...
+    // Hey! Datetime is in milliseconds so everything p much
+    // needs to be divided by 1000.
+    document.addEventListener("DOMContentLoaded", function() {
+      const lobbyCreated = new Date("{{ lobby_detail.lobby_created|date:'c' }}");
+      const elapsedTimeElement = document.getElementById("elapsed-time");
+      const timeConstraint = parseInt("{{ lobby_detail.time_constraint }}", 10);
+
+
+      function updateTime() {
+        //console.log("lobby_created:", "{{ lobby_detail.lobby_created }}");
```
elizabethli31: Remove these print statements!

elizabethli31 (changes requested):
Timer is really really neat and works beautifully! Also, the real time change to the lobby closed message is so sick. I left a few minor change requests. It's a bit beyond the scope of this PR, so if it proves to be too much to add the different match finished outcomes, just let me know. Really great work Nat!

src/templates/games/game_coinresult.html
nugerhard: Implemented

src/chigame/games/views.py
nugerhard: I definitely agree, I think the match_status field should maybe eventually get a fourth–match cancelled–option. For the time being, I am running a frontend check on whether a match met conditions to start or not, on the lonny_details.html page.

src/templates/games/lobby_details.html
nugerhard: Fixed.

src/templates/games/lobby_details.html
```diff
@@ -1,34 +1,134 @@
...
+          }
+        }
+      }
+
+      function updateMatchStatus(lobbyId) {
+        // Eugh. AJAX request to the Django view
+        fetch(`/games/lobby/${lobbyId}/update_match_status/`)
+          .then(response => response.json())
+          .then(data => {
+            console.log(data);
```
elizabethli31: Can you remove this print too

elizabethli31 (changes requested):
Thanks for addressing my changes! One last thing I didn't catch earlier, but currently only once the match has been invoked will the status be updated on the lobby list page. If the match was cancelled due to insufficient players the match status will still show up as lobbied from the list page. For example, Dragon's Lair
image
image
The status updates fine from the list view when the match is actually completed.

elizabethli31 (approved):
LGTM! Again, great work on this PR and thanks for addressing my comments.

nugerhard:
Closing Statement: This PR was definitely down to the wire, apologies. But the placeholder game and the lobby_details page are now connected. And, the timer is as robust as it can be (so long as it can only exist on one page, say lobby_details). 

Here's a bit more color on the sqlite issue, from a conversation on Slack between me, Elizabeth, and Chloe: The database can't calculate the values independently. So, the script at the bottom of our lobbies page has to run in order for the match_status to update. I figured, rather than adding a script at the bottom of the lobby list page which runs a timer check for each of the lobbies on the page (and reloads the when any one runs out), it would be better to have the non-updated lobby status displayed on the page.

elizabethli31:
Issue Score: Excellent

Comments:
Very large PR that provided much more functionality to matches - great work!

PR #346, opened 2023-12-01
Add functionality to create a new chat when a tournament is created

### LINKED ISSUE #226: Add auto-create chat when a tournament is generated

rqueiroz20:
This PR addresses #226 and creates and saves a new instance of a tournament chat object when a tournament is created. 

Specific modifications include the following:
- Importing the `Chat` model into the `src/chigame/games/views.py` file.
- Creating a new function `create_chat()` within the `TournamentCreateView` class that automatically creates a tournament chat for that specific tournament.
- Add a button in the tournament list view and tournament detail view to directly link to the chat.


Riyush (approved):
This addition to the view properly imports the Chat model and creates an instance of the chat with reference to the tournament.

elizabethli31 (changes requested):
Good addition to the model! However, I would like to see at least a button that leads to the newly created chat. This should be a quick change, and I believe it is necessary because typing in chat to the url is not the most accessible from a user perspective.

rqueiroz20:
Hi Elizabeth, I have added your suggested change and added a button linking to the tournament chat in both the tournament list view and the tournament detail view. It only appears if a user is logged in to ChiGame.

src/chigame/games/views.py
```diff
@@ -357,6 +357,11 @@ def form_valid(self, form):
     def get_success_url(self):
         return reverse_lazy("tournament-detail", kwargs={"pk": self.object.pk})
 
+    def create_chat(self):
+        # Auto-create a chat for this respective tournament
+        chat = Chat(tournament=self)
```
elizabethli31: This function isn't working right now, and no chat is being created. You need to create the chat inside the form_valid function. Please make sure to test that this function is actually creating a chat before you push again!

elizabethli31 (changes requested):
Buttons look good, but the functionality isn't working. See my comment on how to fix it!

src/chigame/games/views.py
rqueiroz20: This function was creating a new chat each time on my end, but I will double check to see what is going on.

Edit: Turns out it was not creating a new chat, but somehow linking to an already existing chat I had somewhere? I'm a bit confused on the specifics, but now it creates a new Chat object.

elizabethli31 (changes requested):
This look good now! Can you make one more really minor change (not related to your PR, but this is the last one open for the chat functionality and I missed it earlier. On line 6 of tournament_chat.html, it says "Chat for Match:". Since we've adjusted to have chats for tournaments, could you just change that to tournament? After that, I'll approve and merge!

elizabethli31 (approved):
Thanks for fixing the last minute change and updating the conditions on which Chat is shown. Good work!

rqueiroz20:
In this PR, a new chat associated to a tournament is automatically generated whenever that tournament is created. @MikeChen012345 and @MuyanXie talked over how such a feature would be implemented and the pseudo-code necessary. @rqueiroz20 worked on implementing the code in the branch itself, and @Riyush reviewed the code and made sure it aligned with the pseudo-code.

elizabethli31:
Issue Score: Excellent

Comments:
Nice work implementing the chat auto-creation!

Issue #316, opened 2023-11-28
(tournaments) implement frontend tournament design


Riyush:
This issue involves developing an html and css file for displaying a tournament to a user who has joined a tournament. The page will display other users, the tournament name, and the game associated with the tournament.

Riyush:
With the join tourmanent button merged to dev, we want to make that button redirect to the html page created in the implement-frontend-tournament-design branch.

PR #349, opened 2023-12-01
Tournaments/implement frontend tournament design

Riyush:
The following pull request address issue #316. When a user presses the button to join a tournament, they will be redirected to the tournament bracket url. This url shows a visual representation of the tournament. It displays the tournament name, game and all joined users in a visually appealing format. The bracket is built for tournaments of 8 people. Create a tournament and join as a user to test this functionality. There is a seperate page for displaying a winner. Modifications are made to many files and this functionality relies on the join button which is currently in the dev branch. Rafael helped conceptualise the design and Riyush implemented the code.


src/templates/tournaments/tournament_winner.html
rqueiroz20, line 1: Based on my testing and this PR, I don't see how this file is related to pressing the Join button for a tournament. This could perhaps be included in another PR.

rqueiroz20 (commented):
The tournament bracket frontend display looks amazing! I was able to successfully view the tournament bracket when I pressed the join button when logged in as a user. Some things we should address before i approve the PR:

1. Do we want the tournament to add brackets dynamically based on the maximum number of players in a tournament? I don't think your code currently does that, but we can always leave that feature for future classes.
2. I don't see the point of the tournament winner HTML and CSS pages since I can't interact with them in this PR. Perhaps moving them to a different PR would be more useful?

Besides those things, this looks great!

src/templates/tournaments/tournament_winner.html
Riyush: Got it! I will delete the tournament winner html file as that is not directly related to this branch.

rqueiroz20 (approved):
Everything looks good to go! I was successfully able to view the tournament bracket structure and was able to see multiple people in the bracket view as well. Nicely done!

src/chigame/games/urls.py
```diff
@@ -27,6 +27,8 @@
     # placeholder game
     path("lobby/<int:pk>/coinflip", views.coin_flip_game, name="placeholder-game"),
     path("lobby/<int:pk>/flipresult", views.check_guess, name="flip-result"),
+    # tournament bracket display
+    path("tournament-bracket/<int:tournament_id>/", views.TournamentBracketView.as_view(), name="tournament-bracket"),
```
elizabethli31, line 31: URL should be something like tournament/id/bracket to match the other paths

src/chigame/games/views.py
```diff
@@ -414,6 +421,30 @@ def check_guess(request, pk):
     )
 
 
+class TournamentBracketView(TemplateView):
```
elizabethli31, line 424: Would like to see the bracket view as a button somewhere on the detail view, so users can access it

src/templates/tournaments/tournament_bracket.html
```diff
@@ -0,0 +1,184 @@
...
+      <div class="frame">
+        <div class="div">
+          <div class="group">
+            <div class="overlap-group">
+              <div class="rectangle"></div>
+              <div class="rectangle-2"></div>
+              <div class="text-wrapper">
+                {% if players.1 %}{{ players.1.username }}{% endif %}
+              </div>
+              <div class="text-wrapper-2">0</div>
```
elizabethli31, line 24: 0 should not be hardcoded. Do something to get whether that player has won the match.

elizabethli31 (changes requested):
Sizing of the page looks weird because you are hardcoding by pixels for everything in the CSS which will not accommodate to other computers. This is what I see: 
image
My main issue with this PR right now is that it requires 8 players exactly. If it's possible to make a dynamic brackets feature, that would be ideal. But for the purposes of the demo that's okay, just make sure to document that. The brackets themselves look very good - color choice and gradient are nice touches! See my comments for more detailed changes.

PR #356, opened 2023-12-01
Fixed search bar by renaming query to query-input

### LINKED ISSUE #355: Fix search bar so that it correctly submits
Currently, the search bar is broken; when a user tries to submit a query in the search bar, the search bar does nothing. This is due to the text input field being renamed from "query" to "query-input" and then back to "query" while the javascript uses "query-input". In order to fix this bug, the html element will have the id "query-input". This also necessitates changing the user search results view and the games search results view to use "query-input".

ptlin25:
The currently search bar does not work because the search bar input has an id of "query-input" while some other code uses "query". To fix this bug, all references to the "query" HTML element have been changed to "query-input". The games search results view and the users search results view have been changed to reflect this change.

### DIFF
--- a/src/chigame/games/views.py
+++ b/src/chigame/games/views.py
@@ -139,7 +139,7 @@ class GameEditView(UserPassesTestMixin, UpdateView):
 
 
 def search_results(request):
-    query = request.GET.get("query")
+    query_input = request.GET.get("query-input")
 
     """
     The Q object is an object used to encapsulate a collection of keyword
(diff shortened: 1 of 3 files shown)


omarelkashef (approved):
Looks good to me! and nice job catching and fixing all references to that variable.

rakrusch (approved):
This all looks good to me as well. Nice job fixing this bug Patrick!

majorsylvie (approved):
Thank you for catching, documenting, and fixing this bug!

ptlin25:
In this pull request, references to the search bar's text input field as "query" were changed to "query-input" for consistency and clarity. This fixed the issue of not being able to submit a search request.

majorsylvie:
Issue Score: Excellent

Comments:
Minor because it's a smaller bugfix.

Thank you for your effort and bringing in folks on the users team to review!

PR #358, opened 2023-12-01
Forum Advanced Search Integration

### LINKED ISSUE #310: Integrate the forums advanced search option into the nav bar
The aim of this issue is to incorporate the advance search option where the forum search option is available.

Benjoman12:
Per issue #310, the forum's advanced search bar capability has been integrated into the search bar.


Benjoman12:
This commit is based on the changes that will be made through PR #309. So once that PR is merged the changes that this makes will only be relevant to base.html making this much easier to review.

src/chigame/games/views.py
```diff
@@ -181,7 +181,6 @@ def search_results(request):
     sort = request.GET.get("sort_by", "name-asc")
     players = request.GET.get("players", "")
     page_number = request.GET.get("page")
-
```
joshuamonroe: Although this doesn't appear to do any harm, it also isn't necessary to modify this file. Let's keep the PR clean by removing this change.

joshuamonroe (changes requested):
One small change requested. Otherwise, this looks good to me from a code perspective, although I'm not certain I'm in love with it from a UI perspective. It works for now, but the way the button appears and disappears, I could see it becoming a pain point later on when the nav bar layout starts to become more resolved. However, once that change is made, I'll approve it and we'll let @majorsylvie weigh in.

joshuamonroe (approved):
Thanks for cleaning that up, approved.

src/templates/base.html
```diff
@@ -250,6 +256,17 @@
                 source: []  // placeholder; replace with list of forums
               });
             }
+
+            if (query_type.value === "forums") {
```
majorsylvie, line 260: Please add a code comment explaining the purpose of this `classList` editing.

majorsylvie (changes requested):
There is a section that I would love a code comment on!

I appreciate how you can still use the general search bar as well as the Advanced Search UI!

I was able to search and get matches from post descriptions and titles!

This is a great feature to bring in! Thank you for your work.

Please add the code comment then this is good to merge.

majorsylvie (approved):
Thank you for adding the code comment! Now that code is much easier to immediately understand.

Let's merge this cool searching!

Benjoman12:
The ability to use the Django-Machina "advanced search" function when searching the forums has been implemented. This current "advanced search" function is decently limited in its intelligent functionality as it is the one provided by Django-Machina. This lays the groundwork for future implementation of a more robust advanced search system across the project. Thanks to @joshuamonroe for reviewing the PR.

majorsylvie:
Issue Score: Excellent

Issue #125, opened 2023-11-02
Adding directing from friend requests in inbox to profile page and add option to read and delete notifications


omarelkashef:
Clicking on a Friend Invitation in the inbox should direct to the sender profile page and that page should display options to accept or reject friendship. This builds on the implementing an inbox in #124. There should be buttons to read, unread and delete notifications and that should make them look different in the inbox.

omarelkashef:
While working on redirection from friend requests, we added some bootstrap and JS to the inbox page. Also, we create another issue to add the option to select multiple notifications to do some action (e.g. deleting). we are trying to decide whether that is better done through view function or an API POST request.

omarelkashef:
The redirection work as well as the default message was done based on Notification that belong to FriendInvitations. Other types of Notifications can be implemented once we discuss with respective team what the behavior of clocking on a their Notification should be.

omarelkashef:
Closing Statement: We handled clicks on Notification of type FriendInvitation where they redirect to the sender's profile page. We also added buttons to read, unread and delete notifications in the inbox page as well as default message for friend request notifications.
@omarelkashef worked on the backend logic and view functions
@24raniwalar worked on the front end html, bootstrap and JS

elizabethli31:
Issue Score: Excellent

Comments:
Nice usage of the notification model and also well written updates as the issue progressed. This issue spanned two sprints but looks good otherwise.

PR #363, opened 2023-12-01
Allows users to perform bulk actions on notifications via checkboxes

### LINKED ISSUE #353: Add the option for bulk actions on Notifications
In the current inbox page, users can only do an action one notification at a time and each time the page will refresh. We need to allow users to select multiple notifications and do the same action on all of them. Notifications should also be marked as read once they are clicked by a user.

24raniwalar:
In this PR, we add functionality such that in a user's inbox, they may check multiple notifications using checkboxes and then perform mass delete or read operations by pressing a delete or read button. To do so, we add a view function that handles actions when a POST request is sent from the buttons and use a form to track the status of checkboxes and buttons. We are also marking a notification as read once it is clicked by the user.


24raniwalar:
The PR currently has many files that will eventually be merged with PR 344 so this PR is essentially an extension. We branched off of origin/notifications-junk-box, the branch associated from PR 344. Waiting to request review until PR 344 is passed.

omarelkashef:
We will also be adding some logic to mark a notification as read once it is clicked which was missed in #125  per a suggestion by @elizabethli31. The Issue and PR descriptions has been edited to reflect that.

ishiii-s (approved):
Code looks good to me. URLs.py sets the url paths, views method addresses the ability for users to check multiple notifications using checkboxes in their inbox and then perform mass delete or read operations by pressing corresponding buttons, as well as marking as read. Template file logic for html also doesn’t seem to have any erroneous logic. 

Only thought is if get_object_or_404 would be helpful instead of …objects.get to handle exceptions. Not requesting a change because code still seems to address PR objectively accurately!

elizabethli31 (approved):
PR looks good to go! Thanks for the additions of marking as read when clicked on and also fixing the remove friend edge case.

24raniwalar:
Closing comment: In this PR, users may now mark all notifications as read or delete all notifications using a checkbox next to each notification and buttons for the actions. Both developers worked on all parts of the PR.

elizabethli31:
Issue Score: Excellent

Comments:
Like my comment on the associated issue, this PR added some nice new functionality for notifications!

PR #365, opened 2023-12-01
Displaying Historical Information on a user view

### LINKED ISSUE #300: Implement historical information from view
This will allow for the historical information of a user to be displayed on a page. For example, how many games a user has won, and achievements of that degree.

AndrewM131:
This pull request addresses issue #300 where it aims to display the correct historical information such as games won and such to a view where any view can access it.

### DIFF
--- a/src/chigame/users/views.py
+++ b/src/chigame/users/views.py
@@ -12,6 +12,8 @@ from django.views.generic import DetailView, RedirectView, UpdateView
 from rest_framework import status
 from rest_framework.response import Response
 
+from chigame.games.models import Player
+
 from .models import FriendInvitation, Notification, UserProfile
 from .tables import FriendsTable, UserTable
 
@@ -75,8 +77,9 @@ def user_list(request):
 def user_history(request, pk):
     try:
         user = User.objects.get(pk=pk)
+        player = Player.objects.get(pk=pk)
 
-        return render(request, "users/user_history.html", {"user": user})
+        return render(request, "users/user_history.html", {"user": user, "player": player})
     except User.DoesNotExist:
         return Response(status=status.HTTP_404_NOT_FOUND)
 
--- a/src/chigame/users/views.py
+++ b/src/chigame/users/views.py
@@ -7,12 +7,13 @@ from django.db.models import Q
 from django.http import HttpResponseNotFound
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
+from django.utils import timezone
 from django.utils.translation import gettext_lazy as _
 from django.views.generic import DetailView, RedirectView, UpdateView
 from rest_framework import status
 from rest_framework.response import Response
 
-from chigame.games.models import Match, Player, Tournament
+from chigame.games.models import Lobby, Player, Tournament
 
 from .models import FriendInvitation, Notification, UserProfile
 from .tables import FriendsTable, UserTable
@@ -78,10 +79,12 @@ def user_history(request, pk):
     try:
         user = User.objects.get(pk=pk)
 
-        match_count = Match.objects.filter(players__in=[user]).count()
+        match_count = Lobby.objects.filter(match_status=3, members__in=[user]).count()
         match_wins = Player.objects.filter(Q(user=user, outcome=Player.WIN) | Q(team=user, outcome=Player.WIN)).count()
 
-        tournament_count = Tournament.objects.filter(players__in=[user]).count()
+        tournament_count = Tournament.objects.filter(
+            tournament_end_date__lt=timezone.now(), players__in=[user]
+        ).count()
         tournament_wins = Tournament.objects.filter(winners__in=[user]).count()
 
         return render(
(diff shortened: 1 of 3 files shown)


AndrewM131:
Update: I implemented the code such that there is statistics for matches played, matches won, tournaments played, tournaments won. I added the attributes in views.py in the users folder. I also went into userprofile_detail.html and added a link to the user's stats. The button works and goes to the user's stats. This can be accessed by any user if they go on the user's profile.

src/templates/users/userprofile_detail.html
```diff
@@ -34,6 +34,8 @@
                   <a href="{% url 'users:add-friend' object.user.pk %}"
                      class="btn btn-primary">Add Friend</a>
                 {% endif %}
+                <a href="{% url 'users:user-history' object.user.pk %}"
+                   class="btn btn-primary">Stats</a>
```
24raniwalar: Once you merge dev, you will see that the Edit Profile and Account buttons are sectioned off separately from the general profile buttons such as the Friend Request buttons and the Stats button. The Friend Request and stats buttons should be wrapped such that they either appear alongside the other buttons or such that there is space between the two rows of buttons.

24raniwalar (changes requested):
Changes to views and user_history align. The logic for displaying user stats in the user_history page works well, but the button placement is a little odd in userprofile_detail.html. I have added specific comments there to address the button and where it should be placed; other than that this PR LGTM!

src/templates/users/userprofile_detail.html
AndrewM131: Got it, I'll add them into the if statements just like I did with the account button so it looks cleaner

24raniwalar (approved):
LGTM! I think the if/else logic in the userprofile_detail.html page can be abbreviated since Stats shows up for all users, but that can be left for future developers to clean up in my opinion. Nice work!

src/chigame/users/views.py
```diff
@@ -72,7 +74,23 @@ def user_history(request, pk):
     try:
         user = User.objects.get(pk=pk)
 
-        return render(request, "users/user_history.html", {"user": user})
+        match_count = Match.objects.filter(players__in=[user]).count()
+        match_wins = Player.objects.filter(Q(user=user, outcome=Player.WIN) | Q(team=user, outcome=Player.WIN)).count()
+
+        tournament_count = Tournament.objects.filter(players__in=[user]).count()
```
elizabethli31: Hm this isn't quite right because it's just counting if the player is currently in a tournament (could be on going). If you could add another filter where the tournament_end_date < timezone.now() that should make this work.

src/chigame/users/views.py
```diff
@@ -72,7 +74,23 @@ def user_history(request, pk):
     try:
         user = User.objects.get(pk=pk)
 
-        return render(request, "users/user_history.html", {"user": user})
+        match_count = Match.objects.filter(players__in=[user]).count()
```
elizabethli31: Same comment as for tournaments. I think you might want to use Lobby here instead of match and filter for match_status = 3.

elizabethli31 (changes requested):
Thanks for adding back the account info and the stats page display looks really nice! A couple changes related to the way matches and tournaments played is being counted right now.

src/chigame/users/views.py
AndrewM131: That's a good filter I wasn't aware of, I'll use that

src/chigame/users/views.py
AndrewM131: I'll implement it since it has a match_status I can take advantage of

AndrewM131:
I implemented the requested changes, but my only comment is that a user can be in a match if assigned by an admin, and not be in a lobby. I think there should be some functionality in the future to require a lobby for a match just so a person can't have matches won, but no matches played due to it being tied to lobbies.

24raniwalar:
Added back remove friend functionality; got removed elsewhere in another PR.

elizabethli31 (changes requested):
Remove tournaments related files (just to avoid any accidental merge issues), and I'll approve.

elizabethli31 (approved):
Looks good now! Good work fixing up the matches/tournaments so they are taking in the proper info

AndrewM131:
Closing comment: In this PR, we created a view that can be accessed in the profile of a user that displays the stats of a user, so for example, how many matches a user has been in, how many matches they've won, how many tournaments they've participated in and how many have they won. This is visible to any user when looking at another user's profile. One thing is that match count is linked to lobby, and since an admin can manually add a user to a match without a lobby, there should be something in the future that mandates a lobby for a match such that the numbers stay consistent. Also in the future, there should be more stats, such as rank in Chigame website, and maybe win rate, and things like that.
Team contributions can be found in issue closing statement.

elizabethli31:
Issue Score: Excellent

Comments:
Great work using models from other teams to complete this PR!

PR #366, opened 2023-12-01
Using the current search bar implementation to search notifications based on message

### LINKED ISSUE #351: Allow for searching notifications and filtering by type, sender and sorting by date in inbox
In the inbox page, we need to allow users to search through their notification using words that match the messages body of the notification. They should also be able to filter their inbox according to sender and type as well as sort by date from most to least recent and vice versa

24raniwalar:
In this PR, we will add searching functionality for notifications. Users will be able search for notifications based on their message attribute in the universal search bar.


henrietta-k (approved):
Tested it out on my local environment and it works well. LGTM!

elizabethli31 (approved):
Working now  with the auto added message! LGTM

24raniwalar:
Closing Comment: Users may now use the search bar to search for Notifications. This is done by adding a query option for the universal search bar in base, creating a view function for the results of the query, and displaying the results as search_results.html. We also had to modify part of user_inbox as a bug fix as message attributes were not being linked properly. Both developers worked together on all parts of the PR.

elizabethli31:
Issue Score: Excellent

Comments:
Nice work!

PR #370, opened 2023-12-02
POST API Endpoints for Sending and Accepting Friend invitations between two Users; GET request endpoint for listview for Friend Invites

### LINKED ISSUE #367: [API] Creating API endpoint to send/accept friend requests
Summary:

In order for two users to become friends, one user needs to send a friend request to the other user. In ChiGame, this is handled specifically by the `FriendInvitation` model. A user can create a `FriendInvitation` and another user can then accept that invitation. 

In this issue, we aim to allow a user to make a friend request from the API endpoint: `/api/friend-invitations/send/` and accept a specific friend request at `friend-invitations/accept/<int:pk>/'` where the `pk` is the `id` of the specific request.

We will know when this issue is completed when these endpoints work as described above.

Esterello2:
Creating two separate API endpoints that deal with POST requests for Friend Invitations. To send an invitation, a Friend Invitation object is posted at URL 
`friend-invitations/send/sender<pk>/receiver<pk>/ `
and to accept, the endpoint is run at: 
`friend-invitations/accept/invite<pk>/`
Then a third endpoint that deals with a GET request for a ListView of all Friend Invitations (pending and accepted) is created at:
`/friend-invitations/`


Esterello2:
As of right now, the POST endpoint for "sending invitations" and the GET endpoint for the LISTview of invitations are working. There is an issue with accepting invitations that stems from the model itself (accept_invitation method is not working properly)

Esterello2:
The Accept Invite POST Endpoint works now, meaning that all three endpoints are successful. I also had to make endpoints for the UserProfile objects since it is a separate model from Users. I introduced three endpoints for this, each handling POST, PATCH, and GET (listview) requests. This had to be done to ensure that the Friend Invitation endpoint functioned properly.

bsokolow10 (approved):
Looks good!

src/chigame/api/views.py
```diff
@@ -106,6 +102,76 @@ class MessageView(generics.CreateAPIView):
...
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+class UserProfileListView(generics.ListAPIView):
+    queryset = UserProfile.objects.all()
+    serializer_class = UserProfileSerializer
+    permission_classes = [IsAuthenticated]
+
+
+class UserProfileUpdateView(APIView):
```
majorsylvie, line 165: Both this update and user profile list view are relevant for JWT and are related to authentication.

src/chigame/api/views.py
```diff
@@ -106,6 +102,76 @@ class MessageView(generics.CreateAPIView):
...
+        serializer = FriendInvitationSerializer(invitation)
+        return Response(serializer.data, status=status.HTTP_200_OK)
+
+
+class FriendInvitationList(generics.ListAPIView):
+    queryset = FriendInvitation.objects.all()
+    serializer_class = FriendInvitationSerializer
+
+
+class UserProfileCreateView(APIView):
```
majorsylvie, line 144: What does this class have to do with friend invitations?

majorsylvie (changes requested):
This is looking better now! There are more places to polish up.

Immediate problem is that, with the current version of the code, upon trying to access the sending enpoint, I get an error message which asks me to add `fields = "__all__" to the `UserProfileSerializer`. 

After doing so locally, I could get from `http://localhost:8000/api/friend-invitations/` the list. In each entry seeing the `id` `accepted` status, `timestamp`, `sender`, and `receiver`:
image

I was then able to navigate to multiple different
```
http://localhost:8000/api/friend-invitations/send/<sender pk>/<receiver pk>
```
endpoints.

then upon sending an empty post request (via just clicking post on an empty message via the DRF gui:
image

image

I get feedback that my POST is successful

Then if I give user primary keys that are invalid, I am returned an unsuccessful response to my POST request:
image

image

I can then go to: http://localhost:8000/api/friend-invitations/ and see a lit of all the POST'ed friend invitation 

```py
 {
            "id": 14,
            "accepted": false,
            "timestamp": "2023-12-07T16:49:39.098944-06:00",
            "sender": 1,
            "receiver": 2
        },
        {
            "id": 15,
            "accepted": false,
            "timestamp": "2023-12-07T16:49:43.343774-06:00",
            "sender": 1,
            "receiver": 3
        },
        {
            "id": 16,
            "accepted": false,
            "timestamp": "2023-12-07T16:49:47.416511-06:00",
            "sender": 1,
            "receiver": 4
        },
        {
            "id": 17,
            "accepted": false,
            "timestamp": "2023-12-07T16:49:55.486581-06:00",
            "sender": 2,
            "receiver": 3
        },
        {
            "id": 18,
            "accepted": false,
            "timestamp": "2023-12-07T16:52:01.030614-06:00",
            "sender": 2,
            "receiver": 7
        }
```

**However, there is a problem here**
Once I try to accept an invitation by POST'ing at the `friend-invitation/accept` endpoint (such as at URL: http://localhost:8000/api/friend-invitations/accept/18/ ) I get the following error:

image

**Then there also seems to be some JWT user authentication code still in this PR**

Overall, please:
[] add the `fields = "__all__" to the `UserProfileSerializer`
[] update your code and test if accepting inivtations actually works
[] remove the extraneuous code as it's been commented upon
[] generally answer all comments

src/chigame/api/views.py
Esterello2: OK, I am removing the permission classes

src/chigame/api/views.py
Esterello2: IT is because Friend Invitations use User Profile pks rather than User class, so I had to create a way to make these UserProfile objects

majorsylvie (changes requested):
Even after creating user profiles through the administrator site, I am still not able to actually accept friend invitations. 

I specifically created 4 user profiles via the `admin` page:
image

Then I made have a request with `pk=1` from profile id=1 to profile id=2:
image

Thus to my understand of how your endpoints are supposed to work, I should then be able to go to:
```
http://localhost:8000/api/friend-invitations/accept/1/
```
_which corresponds to friend invitation pk=1_

Then by posting **nothing** to this endpoint, it should then cause this friend invitation to be accepted, in my example making profiles 1 and 2 friends.

However when I post with the following:
image

I get this error:
image

Which upon further inspection, happens in the `.accept_invitation()` call on line 134:
```py
        invitation.accept_invitation()  # Error with this method in the users.models
```
Where in my case, it is failing when trying to run the line:
```py
        receiver_profile = UserProfile.objects.get(user__pk=receiver.pk)
```
where the `reciever.pk` We can create a bugfix issue to ask the users team to look at this method, since we don't think it should be failing.is 2.

Also, unless I need to create UserProfiles through the API endpoint specifically _instead of the admin site_, I don't understand what the problem is.

We can create a bugfix issue to ask the users team to look at this method, since we don't think it should be failing.

For now, if I've properly understood how these endpoints are supposed to work, **your coded functionality fails due to a function outside of your team.**

Overall, this is the current state of trying to run this code, please do the following:
- _if you want_, try to see if the reason I can't accept friend invitations is because of something in your code
- **make a bugfix issue for the users team which details AND LINKS TO what's going wrong in this PR with `accept_invitation()`**

PR #375, opened 2023-12-02
Created API Endpoints to GET List of Users from User Group and vice versa

### LINKED ISSUE #373: Create API Endpoints to GET List of Users from User Group and vice versa
Summary: The goal of this issue is to create an endpoint that will GET all of the groups that a given user is in. In addition, we want to create another endpoint that can GET all of the members of a given group. These endpoints will only support GET requests. They are clearly important, as an API user may want to get information relating `Users` to `Groups`

We will support a `GET` request to the endpoint: `/api/users/<user_id>/groups` to get the groups from a given user. This endpoint allows us to specify the user through the `user_id`, and then we can look up the groups of this user. 

We will also support a `GET` request to the endpoint: `/api/groups/<group_id>/members` to get all the users/members of a given group. This endpoint allows us to specify the group through the `group_id`, and then we can look up the members of the group. 

We will know that this issue is completed when the GET requests to the URLs above is functional

abrahmasandra:
In this PR, we created an endpoint that will GET all of the groups that a given user is in. In addition, another endpoint was created that can GET all of the members of a given group. These endpoints will only support GET requests. They are clearly important, as an API user may want to get information relating `Users` to `Groups`

We support a `GET` request to the endpoint: `/api/users/<user_id>/groups` to get the groups from a given user. This endpoint allows us to specify the user through the `user_id`, and then we can look up the groups of this user. 

We also support a `GET` request to the endpoint: `/api/groups/<group_id>/members` to get all the users/members of a given group. This endpoint allows us to specify the group through the `group_id`, and then we can look up the members of the group.


src/chigame/api/urls.py
cuyakonwu, line 1: It might be worth putting the URLs in a variable and then putting them in the path like the other URLs

cuyakonwu (approved):
Looks Great! I love the consistency and organization

majorsylvie (approved):
Looks great and both endpoints work! 

Please add support for slugs when doing user lookup! 
This should also be done for the friends endpoint, though that should have another issue / PR opened for in in the `2023/wrap-up` milestone.

Please at least add support for username slugs when going to `users/<username_slug>/groups`

Then this will be good to merge!

majorsylvie (changes requested):
Meant to request changes!

---

Looks great and both endpoints work! 

Please add support for slugs when doing user lookup! 
This should also be done for the friends endpoint, though that should have another issue / PR opened for in in the `2023/wrap-up` milestone.

Please at least add support for username slugs when going to `users/<username_slug>/groups`

Then this will be good to merge!

abrahmasandra:
Slug functionality for `users/<username_slug>/groups` implemented. Also refactored the `slug` checking/lookup into a separate helper function called `get_user`

src/chigame/api/views.py
```diff
@@ -9,13 +9,24 @@
...
     MessageSerializer,
     UserSerializer,
 )
 from chigame.games.models import Game, Lobby, Message, User
-from chigame.users.models import UserProfile
+from chigame.users.models import Group, UserProfile
+
+
+# Helper function to get user from slug
+def get_user(lookup_value):
```
majorsylvie, line 23: Good job modularizing this into a helper function

majorsylvie (approved):
Thank you for adding the slug functionality for the `/api/users/<username>/groups` endpoint!

Given that you've addressed the feedback, this is good to merge now.

abrahmasandra:
Closing Comment:

In this PR, we implemented the endpoint `/api/users/<slug>/groups`, which supports `GET` requests and returns back all of the groups that the `user` corresponding with the given `slug` is a part of. The `slug` can either be the `user_id` or the `username` of the `User`. We also implemented the endpoint `/api/groups/<group_id>/members` in order to get all the `User` objects that are a part of a certain `Group`, specified by the `group_id`

majorsylvie:
Issue Score: Excellent

Comments:
Nice work! Good closing comment.

PR #376, opened 2023-12-02
Admin user and regular user view tests

### LINKED ISSUE #371: Create Tests for Admin view vs regular user view

AndrewM131:
This PR addresses issue #371 and has tests what an admin user would see and a regular user would see on different views.


AndrewM131:
Update: Created two tests for admin vs users. Also created GameFactory and TournamentFactory for future implementation. updated the view to only give admins access to user_list.

AndrewM131:
I also added a link in admin account pages where they can view all the users if they click a button

henrietta-k (approved):
I've tested out the two new Pytests in my local environment and it works as expected. LGTM!

src/chigame/users/views.py
```diff
@@ -59,13 +60,16 @@ def get_redirect_url(self):
...
+        table = UserTable(users)
+        context = {"users": users, "table": table}
 
-    # Add information about top ranking users, total points collected, etc.
+        # Add information about top ranking users, total points collected, etc.
 
-    return render(request, "users/user_list.html", context)
+        return render(request, "users/user_list.html", context)
+    else:
+        return HttpResponseNotFound("Access to link is restricted to admins")
```
elizabethli31, line 72: This is ok for now, but it's worth noting that this should redirect back to the user page or something instead.

elizabethli31 (approved):
LGTM! Minor comment that could be worth making a note of for future groups (making sure everything redirects instead of taking you to a blank page).

src/chigame/users/views.py
AndrewM131: Got it, will including in closing statement!

AndrewM131:
Closing statement: In this PR, we implemented tests such that we can test that an admin has a different view than a user. This was made possible by making users_list an admin exclusive page, and this is linked in the admin account page, so only admins can access this page. If a user tried to access this page via a link, they are given a 404 page. In the future, this should redirect them somewhere else, such as their account, and a future team can implement that. 
Team contributions can be found in the issue.

elizabethli31:
Issue Score: Excellent

Comments:
Thanks for adding the users_list as part of this PR as well!

PR #383, opened 2023-12-02
Integrate Notifications Into Game and Tournament Invites

### LINKED ISSUE #301: Integrate notifications for game and tournament
Description: This issue will address integrating notifications into previous functionality of allowing friends to invite other people to games and tournaments (handled by issue #194)

ishiii-s:
This PR addresses #301 to integrate notifications into the functionality to invite users to games and tournaments


AndrewM131 (changes requested):
When you use game invitation, do you mean match invitation as that's what you're using in game invitation. Also, when sending the game invitation or tournament invitation in /admin, it doesn't show up in my inbox unless I manually create a notification. Should this be the case?
Also, your models look correct to me, as well as the code you implemented in views.py, just want some clarification on above.

ishiii-s:
Re: game vs match invitation - everything that refers to a game invitation is an invitation for a match (since a match is a specific instance of playing a game). Sorry if that was confusing
Re: showing up in inbox - this code should create a Notification object (as per the Notification class in models.py) automatically after creating an invitation. For now, I'm not concerned with how that is handled by the inbox. LMK if you're still finding an issue considering that. Sorry for any confusion!

ishiii-s:
> Re: game vs match invitation - everything that refers to a game invitation is an invitation for a match (since a match is a specific instance of playing a game). Sorry if that was confusing Re: showing up in inbox - this code should create a Notification object (as per the Notification class in models.py) automatically after creating an invitation. For now, I'm not concerned with how that is handled by the inbox. LMK if you're still finding an issue considering that. Sorry for any confusion!



> When you use game invitation, do you mean match invitation as that's what you're using in game invitation. Also, when sending the game invitation or tournament invitation in /admin, it doesn't show up in my inbox unless I manually create a notification. Should this be the case? Also, your models look correct to me, as well as the code you implemented in views.py, just want some clarification on above.

Re: game vs match invitation - everything that refers to a game invitation is an invitation for a match (since a match is a specific instance of playing a game). Sorry if that was confusing
Re: showing up in inbox - this code should create a Notification object (as per the Notification class in models.py) automatically after creating an invitation. For now, I'm not concerned with how that is handled by the inbox. LMK if you're still finding an issue considering that. Sorry for any confusion!

AndrewM131:
In the event that it isn't in the inbox, then I'm okay with the current code as I can use the tournament invite and game invite. Thank you for clarifying!

AndrewM131 (approved):
Questions were answered, and in the future, it would be a good idea to integrate the notifications into the inbox, but for now, I can access them through /admin

src/chigame/users/urls.py
```diff
@@ -31,8 +32,10 @@
     path("add_friend/<int:pk>", view=send_friend_invitation, name="add-friend"),
     path("remove_friend/<int:pk>", view=remove_friend, name="remove-friend"),
     path("cancel_friend_invitation/<int:pk>", view=cancel_friend_invitation, name="cancel-friend-invitation"),
+    path("invite_to_game/<int:pk>", view=invite_to_game, name="invite-to-game"),
+    path("invite_to_tournament/<int:pk>", view=invite_to_tournament, name="invite-to-tournament"),
```
jonathanbacon22, line 36: for paths here, do we also need / require more information like the tournament id, in this example?

src/chigame/users/views.py
```diff
@@ -184,6 +185,53 @@ def cancel_friend_invitation(request, pk):
...
+    else:
+        messages.error(request, "Something went wrong please try again later!")
+    return redirect(reverse("users:user-profile", kwargs={"pk": request.user.pk}))
+
+
+@login_required
+def accept_game_invitation(request, pk, match_id):
+    match_invite = GameInvitation.objects.get(pk=pk)
+    match_invite.accept_invitation()
+    lobby_join(request, pk=match_invite.lobby.pk)
```
jonathanbacon22, line 231: do we have to also reference the match? I see we have match_invite.lobby.pk, but do we need to adjust and have match_invite.match.lobby.pk? Depending on how lobby follows from match.

src/chigame/users/migrations/0006_merge_20231205_1656.py
jonathanbacon22, line 1: It appears the empty migration is created from a clash. It likely depends on both 0005_ migrations. I don't think it detriments from anything, but is it better to remove this or make it so that this empty operations migration isn't generated?

src/chigame/users/migrations/0005_alter_notification_type_alter_user_username_and_more.py
jonathanbacon22, line 1: Should there be some sort of logic to prevent repeated invited? like making the sender, receiver, and match unique? Sending multiples invites for the same match can be ok, maybe with a time buffer between when each request can be sent.

jonathanbacon22 (commented):
I don't see any significant issues in this code, but please take a look at my comments for some suggestions and points where I would like some more clarity.

PR #388, opened 2023-12-03
Styling the Forum Vote Buttons 

### LINKED ISSUE #360: Fix CSS for Like and Dislike Buttons in Forums
# Problem
The like and dislike buttons in our forums are currently experiencing styling issues that affect their appearance. This issue needs to be addressed to ensure a better user experience.

## Goals ##

The implementation for this should in some way address styling in these respects
- [x] Color of the buttons
- [x] Shape of the buttons
- [ ] Have the vote count have different styling based on the user's vote
- [x] Placement of the buttons

giomhern:
# Purpose
Fix the rather clunky looking visuals of the forum like and dislike buttons. The intended result is to have buttons that fit the theme of the machina forum.


src/templates/forum_conversation/topic_detail.html
```diff
@@ -7,14 +7,6 @@
...
-    .like-button,
-    .dislike-button {
-      width: 80px;
-      height: 80px;
-      transition: background 0.3s;
-    }
-  </style>
 {% endblock css %}
 {% block sub_title %}
   {{ topic.subject }}
```
bburgess19: If we're removing the CSS from here, we could delete

```
{% block css %}
  {{ block.super }}
{% endblock css %}
```
as well.

src/templates/forum_conversation/topic_detail.html
```diff
@@ -248,4 +204,12 @@ <h4 class="m-0 subject">
...
     });
   </script>
+  <style>
+    .vote-container {
+      display: flex;
+      flex-direction: row;
+      align-items: center;
+      gap: 1rem;
+    }
+  </style>
```
bburgess19: We could put these stylings on the `.vote-container` itself to make the code more succinct. In bootstrap, the classes would be:

`d-flex align-items-center gap-1`

src/templates/forum_conversation/topic_detail.html
```diff
@@ -88,19 +80,23 @@ <h4 class="m-0 subject">
...
+                  </button>
                   <span class="like-count fw-bold fs-3">{{ post.rating }}</span>
-                  <button class="dislike-button rounded-circle"
+                  <button class="btn btn-danger btn-sm dislike-button"
                           type="submit"
                           name="rate"
-                          value="dislike">Dislike</button>
+                          value="dislike">
+                    <i class="fas fa-thumbs-down"></i> Dislike
+                  </button>
```
bburgess19: I think these stylings look awesome! Great choice with the colors and icons!

bburgess19 (commented):
# Review
Great work overall! 

I like the placement of the buttons, and it looks great even on large bodies of text:
image

## Suggestion
Let me know what you think about adding a larger gap between the items in the flex box.

Here's the current `gap-1` on the `.vote-container`:
image

and here's `gap-3`:
image

src/templates/forum_conversation/topic_detail.html
giomhern: This sounds good, but just wasn't sure if this had any additional dependencies. Will update.

src/templates/forum_conversation/topic_detail.html
giomhern: Considering all of the other stylings are inline -- for uniform sake, it is good to maintain homogeneity.

bburgess19 (approved):
These are changes bringing machina changes into our templates. The code and the buttons looks good to me!

src/templates/forum_conversation/topic_detail.html
```diff
@@ -0,0 +1,204 @@
...
+                     data-post-id="{{ post.id }}"
+                     data-user-rating="{{ post.user_rating }}"
+                     data-rating="{{ post.rating }}">
+                  <button class="like-button btn btn-primary btn-sm"
+                          type="submit"
+                          name="rate"
+                          value="like">
+                    <i class="fas fa-thumbs-up"></i> Like
+                  </button>
+                  <span class="like-count fas">{{ post.rating }}</span>
```
bburgess19: # Color like button text
 As noted by @joshuamonroe mentioned [here](https://github.com/uchicago-cs/chigame/pull/341#pullrequestreview-1765656393), we should have some color indicator based the user's vote.

e.g. the user disliked:
image

the user liked:
image

The colors used here are the same colors from the buttons:

```
#3579F6 // blue
#CB444A // red
```

bburgess19 (changes requested):
# Report
I like the stylings! Just implement the styles based on the user vote then resubmit!

src/templates/forum_conversation/topic_detail.html
giomhern: I have instead just changed the color of the icon as opposed to the number because I figured it would be somewhat convulated for sake of people like me who are colorblind LOL. Attached are photos of the changes instead. 
Screenshot 2023-12-05 at 2 55 09 PM
Screenshot 2023-12-05 at 2 55 15 PM

giomhern: To add onto this, the deselecting when the other one is selected is not completely working. It works from like --> unlike but not the opposite way, but this can be a minor PR.

majorsylvie (changes requested):
Running on firefox I do not see the graphics: 

image

image

Even after force refreshing the cache (mac: `cmd + shift + r`), the thumbs up/down don't appear.

would be good to have this working in a way that is supported on other/older browsers!

majorsylvie (approved):
Looks better! 

FYI, the placement of the buttons shifts depending on the width of the current vote count: 

https://github.com/uchicago-cs/chigame/assets/86740254/d57744d0-c3e7-4dec-bfb7-51b707d50e58

This is not a problem for this PR, something you might want to make a separate issue if this behaviour is not desired.

src/templates/forum/board_base.html
```diff
@@ -19,7 +19,7 @@
   <link rel="stylesheet" href="{% static 'machina/custom-styling.css' %}" />
 {% endblock css %}
 {% block body %}
-  <div class="my-5 container" id="main_container">
+  <div class="my-5 container " id="main_container">
```
joshuamonroe: Very minor, but lets get rid of this extra space to keep the PR clean.

joshuamonroe (changes requested):
This looks great and everything seems to be working as intended on my end. One very small housekeeping request in the code comments and then I'll be happy to approve it.

joshuamonroe (approved):
Thanks for cleaning that up, good to go now. Approved!

bburgess19 (approved):
The buttons look great to me! I added a `min-width` to the buttons at the end to make sure changing the vote count doesn't move the buttons around until the vote count is ridiculously large (>100,000)

bburgess19:
# Closing comment

We added styles to the barebones like buttons, and ensured they behaved properly when toggling and switching between button clicks.

## Detailed notes

We faced some issues with the voting icons not appearing for Firefox, but that was solved by using FA5 class names for those icons, i.e. `fas` instead of `fa-solid`. _(credits: @majorsylvie)_

We made sure the user's vote was clear by choosing to fill an icon instead of solely changing the icon's color. This helped with accessibility. _(credits: @giomhern )_

We added a bit of extra space around the vote counter so the button placement would shift so much with every vote. _(credits: @majorsylvie)_

Thank you to @joshuamonroe and @majorsylvie for reviewing, and @giomhern for spear-heading this!

majorsylvie:
Issue Score: Excellent

PR #395, opened 2023-12-04
Add new styles for forum creation page

### LINKED ISSUE #382: Improve the styling on the forum creation page
# Goal
Improve the styling of the forum creation page

## Suggested Edits
Here is the current page:
image

- [x] Header text that says "Create new forum"
- [x] Fields with the text box beneath it
- [ ] Radio buttons for the "Forum type" field (see issue #379)
- [x] Change "Save" to "Create"
- [x] Add styling to "Create" button
- [x] Shrink the size of the "Description" field

Here is an example of a reference form:
image

bburgess19:
# Purpose
This PR seeks to add new styling for the forum creation page. Merging of this PR will close #382.

## Notes
Here is an example of how the page currently looks:
image

Also implemented is more styles for the image add/remove button:
image


src/config/urls.py
```diff
@@ -21,7 +21,10 @@
     path("account/", include("django.contrib.auth.urls")),
     # Django-machina forum
     # https://django-machina.readthedocs.io/en/latest/getting_started.html#urls-configuration
-    path("forum/", include(machina_urls)),
+    path("forums/", include(machina_urls)),
+    # This is necessary in order to add our own pages under `/forums`
```
giomhern: Thanks for commenting the clarification here!

src/templates/forum/forum_form.html
```diff
@@ -0,0 +1,78 @@
...
+        // Clear the images
+        if (imageField && imageField.files[0]) {
+          imageField.value = "";
+          event.currentTarget.removeEventListener('click', removeSelectedImage);
+          event.currentTarget.remove();
+
+          // Remove the styles from the upload button
+          imageLabel.classList.remove('btn-warning');
+          imageLabel.classList.add('btn-secondary');
+          imageLabel.innerText = "Upload forum image";
```
giomhern, line 72: One extra comment here is that I cannot really tell if the file I uploaded was the correct one I wanted. I love the change in colors, but in the previous iteration, we got a little preview of the image rather than just the colors changing to not only confirm that it was uploaded, but that the correct one was uploaded. Let me know your thoughts!

giomhern (commented):
Overall, this looks really good. It fits more into the machina implementation and really well done styling. My only concern would be with the image, but I would love to hear your thoughts @bburgess19.

src/templates/forum/forum_form.html
bburgess19: Great point, I've added a little preview to the currently displayed image:
image

and here's how it displays on the main page:
image

src/static/css/forums/forum_form.css
```diff
@@ -0,0 +1,30 @@
...
+label {
+    display: block;
+    font-weight: 600;
+}
+
+.container form {
+    display: inline-block;
+    width: 40%;
+    box-sizing: border-box;
+    background: #eaeaea;
```
giomhern, line 14: Actually like the abstraction here with the new file dedicated to CSS.

src/templates/forum/forum_form.html
```diff
@@ -1,14 +1,96 @@
...
+
+          // Change the text and styling when an image is uploaded...
+          imageLabel.classList.remove('btn-secondary');
+          imageLabel.classList.add('btn-warning');
+          imageLabel.innerText = "Change forum image";
+
+          // and add an image preview and trash can to remove the image
+          imageField.parentElement.appendChild(trashButton);
+          trashButton.addEventListener('click', removeSelectedImage);
+          trashButton.imgPreview = imgPreview;
```
giomhern, line 51: Addition of the preview image was well needed and nicely implemented here!

giomhern (approved):
This is looking a lot better now with all of the prime functionality from before. The colors and the UI feedback when creating a forum are really nice touches. Love what you did here Ben! Approving! Thanks for incorporating my feedback!

majorsylvie (approved):
This looks so sick!

I'm successfully able to make forums, with and without images. 

The only thing that is odd is after I make a forum, I get two different "forum successfully created" popups:
image

This is not an problem for this PR, **please add an issue to the backlog about addressing this bug**

bburgess19:
# Closing comments
We added new styles to the forum creation page. We had to add extra logic to ensure the optional added image is displayed in the forum list.

> The only thing that is odd is after I make a forum, I get two different "forum successfully created" popups: image
> 
> This is not an problem for this PR, **please add an issue to the backlog about addressing this bug**

Issue #418 was created to address this.

majorsylvie:
Issue Score: Excellent

PR #414, opened 2023-12-05
Added default forum user permissions

### LINKED ISSUE #412: Add default forum permissions setting
During testing I noticed that regular user accounts do not have permissions to view the forums.

This is the forum viewed as an admin. Thus far, most testing has been done from an admin account.
Image

And this is the same forum viewed as a regular user.
Image

A fix would be pretty easy, we would just need to set django-machina's default authenticated user permissions setting according to the defaults provided [here](https://django-machina.readthedocs.io/en/latest/settings.html#machina-default-authenticated-user-forum-permissions). This would apply to all authenticated users across all forums that don't have more restrictive permissions.
```
MACHINA_DEFAULT_AUTHENTICATED_USER_FORUM_PERMISSIONS = [
    'can_see_forum',
    'can_read_forum',
    'can_start_new_topics',
    'can_reply_to_topics',
    'can_edit_own_posts',
    'can_post_without_approval',
    'can_create_polls',
    'can_vote_in_polls',
    'can_download_file',
]
```
Alternatively, we could consider setting permissions manually on a per forum basis, more info on what that would entail and a full list of possible permissions can also be found [here](https://django-machina.readthedocs.io/en/latest/forum_permissions.html).

joshuamonroe:
To address issue #412, I've added the necessary django-machina setting to define default user permissions on the forum.


bburgess19:
Thank you for reviewing thus far! I've created #415 to address the like counter being included in the forum list bar. Otherwise, see the comment threads for specifics.

src/config/settings/base.py
```diff
@@ -357,3 +357,25 @@
...
+    "can_see_forum",
+    "can_read_forum",
+    "can_start_new_topics",
+    "can_reply_to_topics",
+    "can_edit_own_posts",
+    "can_post_without_approval",
+    "can_create_polls",
+    "can_vote_in_polls",
+    "can_download_file",
+]
```
bburgess19, line 383: These permissions look correct, and thank you for the documentation

bburgess19 (approved):
# Review
The permissions are reasonable, and the documentation is a great inclusion. LGTM.

## Proof of functionality

https://github.com/uchicago-cs/chigame/assets/55334507/00bb58a1-0dea-43b7-9da9-06928e00c72e


*squirrel_lover15@example.com is unauthenticated*

majorsylvie (approved):
I can successfully sign in as a non-admin user and see the posts!
Then when I'm completely signed out, I cannot see any posts in the forum!

A good next issue to come from this would be a landing page of some sorts that recognizes if the user is not signed in and tells them they need to sign in to view the forum + a link to sign in.

This way someone who's not signed in will know that "yes, things actually have been posted to the forum, I just can't see them". Instead of "this forum looks empty, I have no idea if this is because there are no posts or if I can't see them.

Thank you for your documention!! Particularly for future implementation. This is phenomenal to put in code as well as github issues.

Thank you for your work!

joshuamonroe:
Closed without significant revision. Thanks to @bburgess19 for reviewing the PR and for the video documenting the result.

@majorsylvie We already had an issue open to create a landing page, although it was recently closed due to inactivity and lack of a clear purpose. However, I've gone ahead and reopened it and clarified that it should cater to unauthenticated users, referencing your comment here.
- #252

majorsylvie:
Issue Score: Excellent

PR #434, opened 2025-04-08
Updated developer guidelines

borjasotomayor:
I've updated the developer guidelines as follows:

- Expanded the Style section to specify what checks we perform, and instructions on how to perform the checks locally.
- Added a Commit Message section
- Updates the branching section to refer to GitHub Flow instead of git flow.

### DIFF
--- a/docs/developer/guidelines.md
+++ b/docs/developer/guidelines.md
@@ -2,53 +2,107 @@
 
 ## Style 
 
-Your code must follow the [Python Style Guide](https://uchicago-cs.github.io/student-resource-guide/style-guide/python.html) 
-in the [UChicago CS Student Resource Guide](https://uchicago-cs.github.io/student-resource-guide/) 
+ChiGame is a large codebase, and it is important that all code
+is written in a consistent style. This is enforced via a series of
+automated style checks, which will detect (and often automatically fix)
+any style issues in the ChiGame codebase.
 
-## Pre-Commit Hooks
+These checks include the following:
 
-When setting up a [local development environment](local.md), you must make sure to
-install the pre-commit hooks as instructed in that page. This will run a series of
-code quality checks before committing any code, and will ensure your code is
-following proper style.
+* [Black](https://black.readthedocs.io/en/stable/) - a code formatter that
+  automatically formats your code to follow the [PEP 8](https://peps.python.org/pep-0008/)
+  style guide.
+* [Flake8](https://flake8.pycqa.org/en/latest/) - a linter that checks
+  for common style issues in Python code.
+* [isort](https://pycqa.github.io/isort/) - a tool that automatically sorts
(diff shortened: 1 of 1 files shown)


docs/developer/guidelines.md
```diff
@@ -2,53 +2,111 @@
...
+* Miscellaneous checks for trailing whitespace, missing newlines, etc.
 
-Often, the pre-commit hooks will automatically fix any style issues in your code,
+These checks are enforced via GitHub Actions any time you push code to the
+repository, and pull requests will not be approved if they don't pass
+these checks.
+
+To ensure your code already passes these checks, we suggest you get into
+the habit of running the `black` command to automatically reformat your code
+(or use a code editor plugin to do so automatically). For example:
```
abargher: I think we should provide a quick link to instructions on how to set this up (if not here, at least somewhere). I think this link should be plenty for VSCode: https://code.visualstudio.com/docs/python/formatting

docs/developer/guidelines.md
```diff
@@ -2,53 +2,111 @@
...
+- If the commit represents a work-in-progress, you can use the `WIP` prefix (e.g., `[api] WIP: add new tournaments endpoint`).
+- Commit message titles must not exceed 50 characters. The lines of the body of the commit message, if any,
+  must not exceed 72 characters (see the [Git 50/72 rule](https://www.midori-global.com/blog/2018/04/02/git-50-72-rule))
+- If the commit relates to an issue or PR, you must include the issue/PR number
+  in the commit message body, using the format `#<issue number>` (e.g., `#42`). This
+  ensures the commit is linked to the issue/PR in GitHub.
+
+We also recommend reading [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/) for more
+details on how to compose the title and body of the commit message.
+
```
abargher, line 87: I like these commit message guidelines a lot, although I think it would help to include an example message (a made-up one is fine) that follows all of the guidelines.

abargher (changes requested):
I like the guidelines overall, just a couple small places for some extra useful reference information.

cmarshall282:
I agree with Alec's comments, but other than that it looks good to me!

jacqsmei:
sorry got to looking at this a bit later, but i agree w alec and chris!

MuyanXie (approved):
the current version already looks great to me! I agree with the other comments and I also think it might be a good idea adding the setup guide for python formatting

docs/developer/guidelines.md
borjasotomayor: Sounds good, I've added a link to the VS Code guide.

docs/developer/guidelines.md
borjasotomayor: Good idea, I've included a few examples.

Issue #450, opened 2025-04-16
Expand notification model for types of notifications


hollyfwei:
We can add a notification class that expand  Notification model for each model. This will facilitate dealing with each different type of notification depending on the actor (model). For example, notifications from FriendInvitation directs to profile page of the sender while notifications from MatchProposal directs to a page to accept or deny. These new sets of classes will not be tables in the databases. They will just be interfaces of Notification that provide custom methods such as redirect_to_actor()

davdma:
@hollyfwei and I have implemented `BaseNotificationHandler` and subclasses of that handler with a method that allows for separate redirects. The code is in the recent PR. I think we can close this issue for now.

cmarshall282:
Issue Score: Satisfactory

Comments:
Great job on this issue! Here are some thing to keep in mind:

- Your issue did not include any status updates between the opening of the issue and the closing of the issue (and this issue was not small enough to justify having no status updates at all). If you make any intermediate progress before finishing the task, then it should be documented on the GitHub issue via a comment.

PR #455, opened 2025-04-18
[User Profiles + Friending] Added notification handler to models.py and adjusted views.py

hollyfwei:
To address issue #450, I added notification subclasses for FriendRequestNotification, ReminderNotification, UpcomingMatchNotification, MatchProposalNotification, and GroupInvitationNotification in users>models.py.


hollyfwei:
@davdma and I added BaseNotificationHandler and associated subclasses FriendRequestNotification, MatchProposalNotification, GroupInvitationNotification, ReminderNotification, UpcomingMatchNotification. Adjusted views.py to handle these changes accordingly and tested in browser.

davdma:
Currently notifications can be seen in the inbox and works for friend requests (it will redirect to the user profile with accept or decline invitation buttons). However currently @hollyfwei and I have decided to hold off for implementing where redirects will go for the other subclasses of notification types (e.g. group invites, match proposal, upcoming match, reminders). That should be work for a future issue.

harriet-fisher (approved):
All looks good! I have a couple of comments we may want to keep in mind as we continue dev:
- Maybe add validation to FriendRequestNotification class? self.notification.actor.sender.pk is a long attribute chain and if any of the attributes are null, it'll throw an error, so we should maybe consider implementing some checks before returning?
- for future devs/long-term maintenance, should we make a separate file for notifications, since they aren't explicitly models? This may be a question for @uchicago-cs/chiventure-senior-developers ?

Other than that 👍

src/chigame/users/views.py
```diff
@@ -4,7 +4,7 @@
 from django.contrib.auth.mixins import LoginRequiredMixin
```
cmarshall282, line 4: Right now, views.py has a mix of using the raw Notification class and our interfaces. I think we want to design this in such a way that we can remove the use of some sort of mixed approach (and I believe the original intention of the issue was to be able to do so).  

I do think that is enough work to warrant another task. So we have a few options that we can discuss on slack or at the Sprint 2 meeting
1. Turning this issue into a interface designing issue (without touching views.py or anywhere that notifications are actually used) and continue on this work in another issue
2. Extend this issue into the next sprint to remove this mixed reference.
3. Finish this issue where we are and start a new issue (which is similar to 1)

I personally would lean towards 1 or 2 but am open to your suggestions.

jonathanbacon22 (commented):
I agree with Chris, and in my opinion we should do option 1 or 2 of what he suggested. I also agree with what Harriet stated regarding validation. Other than that, this looks good to me! Nice work.

hossainmd-uc (approved):
I reviewed the Notification code additions in the context of a `FriendInvitation`. I was able to test the code using the web interface and django shell to inspect the database. It seems that the code is pretty robust and functional. See below for more details:

I verified the following parts from the `NotificationQuerySet`/`Notification `class:
- Actor correctly set based on Notification type: Working
- Filter_By_X Commands: Working
- Bulk Commands: Working
- is_read/unread/deleted, etc: Working

Next Steps:
- Needs Merge with dev to integrate soft-deletion of invitation, after which an update needs to be made to preserve Notifications
- Interface needs a restore all deleted button

cmarshall282 (approved):
Looks good

cmarshall282:
Issue Score: Needs Improvement

Comments:
This was a great early PR. Here is some feedback:

- A branch should represent work on a concrete feature that is being added to the mainline branch, instead of a collection of many different changes.
- This PR represents a large amount of work that could've been integrated into `dev` through smaller PRs earlier in the quarter. Remember that you should not treat PRs as a singular event to merge all or most of your work into `dev`. Integrating incrementally can help identify integration issues early, as well as ensure that you are not rushing to merge all your code towards the end of a sprint.

PR #462, opened 2025-04-20
Implement mutual friendship relationship between two User instances

### LINKED ISSUE #456: Implement mutual friendship relationship between two User instances
The current implementation defines friends as an attribute of UserProfile, which points to a User instance. This setup complicates the data model and creates a mismatch in abstraction by spreading user-to-user relationships across different models.
This issue proposes simplifying the architecture by defining the friendship relationship as a self-referential many-to-many field within the User model, allowing users to directly connect with each other.

This task is complete when:
- The User model allows symmetric friendships
- Legacy friend-related code in UserProfile is removed
- Existing app functionality that uses friendship still works after the update

harriet-fisher:
To address issue #456 we removed legacy friend attribute from UserProfile and made the friendship relationship symmetrical and self-referential as a User attribute. We updated FriendInvitation to comply with this new model and added unique together constraints to prohibit users from sending duplicate friend requests. Finally, we updated api/views.py, users/views.py to function with the update.

- The User model now allows symmetric friendships and functionality has been moved from UserProfile to User model
- Legacy friend-related code in UserProfile has been removed
- Existing app functionality that uses friendship still works after the update


hollyfwei (approved):
Looks great!

src/chigame/users/models.py
```diff
@@ -96,13 +96,13 @@ class FriendInvitation(models.Model):
...
+        unique_together = ("sender", "receiver")
+
     def accept_invitation(self):
         sender = self.sender
-        sender_profile = UserProfile.objects.get(user__pk=sender.pk)
         receiver = self.receiver
-        receiver_profile = UserProfile.objects.get(user__pk=receiver.pk)
-        sender_profile.friends.add(receiver)
-        receiver_profile.friends.add(sender)
+        sender.friends.add(receiver)
```
cmarshall282: Do we need to add the friendship going both ways, or does Django take care of that for us?

src/chigame/users/views.py
```diff
@@ -104,18 +104,18 @@ def user_history(request, pk):
 
 def user_profile_detail_view(request, pk):
     try:
```
cmarshall282: In general, can we try documenting some of the more complicated pieces of code (like this one). It seems as though we are lacking some decent documentation for what we are currently doing, and whenever we make some changes or updates, I'd like to add some in-code documentation as to what this version of the code does.

cmarshall282 (changes requested):
This is a great start! The code looks good. I just had one question that should be easy to test out.

If you could add a little more documentation to what we are doing that would be great!

src/chigame/users/models.py
jonathanbacon22: I'm also curious about this.

jonathanbacon22 (commented):
Looks good to me!

src/chigame/users/models.py
davdma: @cmarshall282 and @jonathanbacon22 from reading Django documentation I assume with `ManyToManyField` and `symmetrical=True`, doing `sender.friends.add(receiver)` also will do `receiver.friends.add(sender)` since it goes both ways.

davdma (approved):
Code looks good! We only need to add some documentation and should be good to go. Also, correct me if I am wrong, but it seems we are still keeping the `UserProfile` model? We just moved the friending mechanism over to `Users`. In that case @hollyfwei and I's code can be merged with only minor changes once you merge with dev

hossainmd-uc:
> Code looks good! We only need to add some documentation and should be good to go. Also, correct me if I am wrong, but it seems we are still keeping the `UserProfile` model? We just moved the friending mechanism over to `Users`. In that case @hollyfwei and I's code can be merged with only minor changes once you merge with dev

That's correct; we are still keeping the `UserProfile` model and moving the friending mechanism into the `Users` class.

src/chigame/users/views.py
hossainmd-uc: Yes, I completely agree with this. When I first started modifying this function, it took a second for me to understand how it worked since there was no documentation (at all in views.py for Users app). Moving forward, adding documentation would greatly help current and future team members by saving time and confusion.

cmarshall282 (approved):
Looks good! Thanks for adding documentation!

harriet-fisher:
Tests pass now that we have resolved some migration conflicts. Should be ready to merge, @cmarshall282!

hossainmd-uc:
User-User friendship implementation is completed, so I am closing this issue.

cmarshall282:
Issue Score: Excellent

Comments:

PR #522, opened 2025-04-23
[livechat] Added LiveChat models to models.py

### LINKED ISSUE #491: Implement live chat tables
To begin working on chat features we need to work on live chat by adding the tables, we need to add the relationships and the ORM models so we can work with it! I will provide the info and @pmdepaulo will actually implement it.

pmdepaulo:
Added our outlined tables to the `games/models.py` file. These models can be imported to the users/models.py file if needed. They include the basic attributes for each of the tables, which map the different live chats, messages, and users that use the live chat feature.


aflekkas (changes requested):
Channel needs to be unique for live chat, other than that looks good to me

pmdepaulo:
Just fixed, thanks for the feedback!

aflekkas (approved):
Looks good to me!

abargher (approved):
Looks good to me as well!

abargher:
Issue Score: Excellent

Comments:
Great work!

Issue #413, opened 2023-12-05
Fix logic bug in removing friends


omarelkashef:
Currently, when a friend is removed, the corresponding friend invitation is deleted from the database but the notification assigned to that invitation is only marked as deleted and a user can retrieve it. However, it will not have an assigned "actor" because the FriendInvitation is deleted. One solution is that FriendInvitation is be treated as and is only marked as deleted and can be retrieved with the notification. Now, we only work around this and display a message that this notification is invalid.

hossainmd-uc:
So far: Jonathan added an is_deleted field to FriendInvitation & readjusted the delete function to circumvent the issue of FriendshipInvitation not having an assigned actor. I created the appropriate migrations file since he was having some error on his end.

**Potential bug due to persistent invitation:** Upon removing a friend, you are given the option to accept/decline friend request once again without having being sent another friend request. (I can't assess if this is a logic bug because issue #455 has major updates to `Users` app that were not merged into dev). When I accept the friend request and then remove friend, there's an error (attached).  Next steps are to resolve issue #455, test the features to assess what doesn't work, then make appropriate changes.

![Image](https://github.com/user-attachments/assets/c897f7b5-0da6-4e30-ae06-3a1baebec957)

hossainmd-uc:
Upon further testing and using the shell to inspect the tables, once a friend is removed, the corresponding invitation remains in `FriendInvitation` table and is only marked as deleted. I am unable to test whether the notification has an assigned actor now because the previous team has implemented a solution where it displays the attached message, preventing access to the notification.

![Image](https://github.com/user-attachments/assets/88dcb87a-6e9b-4aa9-8a55-57215521f9dd)


Moving forward, we should assess if this error should be removed and whether users should be able to view the invitation via the notification. I believe the point of this issue is to allow users to have access to the notification that displays the corresponding friend invitation.

jonathanbacon22:
Rain, do you know where this error is coming from? I do not see an error message with that text in the users views.py file.

hossainmd-uc:
Yes this is the old implementation `views.py`:

![Image](https://github.com/user-attachments/assets/1b989ac6-fdba-4bdd-bcca-981116039529)

This is the one in branch users/remove-friend-logic-bug:

![Image](https://github.com/user-attachments/assets/939cc1ac-ee1b-41db-b676-bc32c086fe30)

Let me double check accessing the notification in the users/remove-friend-logic-bug

[...]

PR #540, opened 2025-04-26
[Users Profiles + Friending] Fixed bug that deleted a FriendInvitation if a friend was removed.

jonathanbacon22:
This pull request references Issue #413. Previously, when a friend was removed, the corresponding FriendInvitation was also deleted. This broke the notification logic, as we no longer had an assigned actor. So, I changed the deletion method to be a "soft" delete, in which we still have the invitation record in the database, meaning we can still view deleted invitation and notifications will still have their "actor"

This was done by simply adding a is_deleted flag, and creating a delete method that doesn't actually do a hard delete.


jonathanbacon22:
Did not account

jonathanbacon22:
Did not account for making the migration (I made changes to models.py)

PR #546, opened 2025-04-27
[knowledge-base] startapp & add data models

### LINKED ISSUE #514: Design markdown file storage
The knowledge base stores markdown content to render as game guides. Some platforms like GitHub and BitBucket (sources of inspiration for our markdown guides) store this data as actual files, and others like Slack embed them in database objects. We will need to determine which approach we should use. If we opt for file storage, we need to demonstrate working Django file storage. If we opt for database storage, we should implement the models in ChiGame.

yizou275:
This PR mainly implements the data models according to https://github.com/uchicago-cs/chigame/wiki/Knowledge-Base-~-Tables.md. Details include:
- startapp knowledge_base, update INSTALLED_APPS in src/config/settings/base.py
- update knowledge_base/models.py, including `Guide`, `ReviewFeedback`, `Favorite`, `Like`
- add `published_guide_id` field to `Games` object in games/models.py
- add `moderator` boolfield to `User` object in users/models.py

Closes: #514


Luca-W-D (changes requested):
Really great work! Super detailed – I love the behavior that calling .save() on a Guide object has conditional logic for updating the timestamp. A ton of work clearly went into this PR.

I cloned the branch locally and looked into the test debug log. I think the tests are failing because the knowledge-base DB migrations weren't pushed to the remote repo. After `make-migrations` and `migrate`, this branch works as expected! Happy to look again after that file/those files are pushed.

yizou275:
Thanks Luca for your insight! I just pushed the migration files. The check seems to pass now:)

Luca-W-D:
Great! I have only one mechanical note:
 * Right now, you have likes and favorites implemented as explicit tables. I think that it may be easier for us to implement these as a many-to-many relationship. This creates a table under the hood anyway, and it allows for easier interactions in code. Specifically, this would allow us to call something like `guide.likes.count()` so we don't have to maintain `n_likes` as a separate field. Does this sound right to you?

Also, this is just nit-picky, but the examples we've covered in class have defined integer choices to be a subclass of the model they're used in. I'd recommend moving GuideStatus to be under Guide (so that if we ever call `Guide` in code we can immediately reference `Guide.GuideStatus.PENDING` or similar). Of course, your call on whether to change it!

yizou275:
These are great points! I addressed both in my new commit. Thank you so much:)

Luca-W-D (approved):
Looks good to me! I was going to comment that the explicit integer choices under ReviewFeedback may be verbose, but I realized that it allows us to avoid a "pending" response to a review. Again, great detail! :)

jacqsmei (approved):
Great job on this! I agree with the notes Luca left.

jacqsmei:
Issue Score: Excellent

Comments:
Nice job on this one + great feedback/review

PR #557, opened 2025-04-28
[User Profiles + Friending] Fixed Removing Friend Logic Bug

jonathanbacon22:
This pull request references Issue https://github.com/uchicago-cs/chigame/issues/413. Previously, when a friend was removed, the corresponding FriendInvitation was also deleted. This broke the notification logic, as we no longer had an assigned actor. So, I changed the deletion method to be a "soft" delete, in which we still have the invitation record in the database, meaning we can still view deleted invitation and notifications will still have their "actor"

This was done by simply adding a is_deleted flag, and creating a delete method that doesn't actually do a hard delete.


hossainmd-uc (approved):
LGTM! 

For future we might want to think about the following:

- Interface to display the notifications and use the is_deleted field to invalidate a pending friend request (when you remove friend, you can still accept the friend request).

- Interactive element for users to access the notifications pertaining to previous FriendInvitations after it is moved to the deleted notifications section.

cmarshall282 (approved):
Looks great to me! Looks like we have some merge conflicts to resolve

cmarshall282:
Issue Score: Excellent

Comments:
Great job! Here are some things to consider:

- Make sure to select your team's identifier under "Labels"

PR #524, opened 2025-04-23
[User Profiles + Friending] Add uniqueness constraint for notifications

jonathanbacon22:
This resolves issue #489, by added uniqueness constraints to the users model. I create Meta classes for each relevant model, in order to ensure duplicate invitations are not possible.

PR #562, opened 2025-04-28
Mechanism to remove addressed notifications

### LINKED ISSUE #470: Add a mechanism to remove notifications from user inbox that have been addressed
We would like to add a mechanism to remove notifications from the user's inbox that have been addressed. For example, if a friend request was sent to a user and then accepted or deleted, then the corresponding notification should be removed from the inbox automatically. For notifications that do not need an action such as a general reminder notification or an upcoming match notification, the notification does not need to be removed automatically and would just be optionally cleared by the user manually.

harriet-fisher:
To address issue #470:
- Added an “is_addressed” function to Notification class to validate auto removal
- Updated friendship acceptance to delete friend request after acceptance
- Updated inbox view to check for addressed notifications before re-filtering
- Updated unfriending to appropriately handle friendship removal and notification removal
- Updated to completely delete the notification, rather than just marking it as deleted

Friendship notifications are now deleted upon acceptance (not just marked as deleted). I updated the notification model to automatically check for addressed notifications and delete them as well as added automatic checks in the views to ensure when users check their inbox, the appropriate notifications are deleted.


src/chigame/users/models.py
```diff
@@ -96,16 +97,28 @@ class FriendInvitation(models.Model):
     timestamp = models.DateTimeField(auto_now_add=True)
     objects = FriendInvitationManager()
 
+    class Meta:
```
jonathanbacon22, line 100: This was also resolved in pr #524

src/chigame/users/views.py
```diff
@@ -159,24 +203,29 @@ def send_friend_invitation(request, pk):
...
             actor=friendship,
             receiver=receiver,
             type=Notification.FRIEND_REQUEST,
         )
-        notification.mark_as_deleted()
+        notification.delete()
+
     num, _ = friendship.delete()
-    notification.mark_as_deleted()
     if num:
```
jonathanbacon22, line 229: I believe my bug fix can be integrated well into this.

jonathanbacon22 (commented):
I believe that the changes you made integrate well into the bug fix that I also submitted.

cmarshall282 (approved):
This looks great! If we can implement Jonathan's soft-delete mechanism (that resolves a bug), that would be great!

PR #568, opened 2025-04-28
Resolving 'Make the admin page for games more robust #135' via Games/admin page

### LINKED ISSUE #135: Make the admin page for games more robust
The admin page for the Game model needs considerable enhancements. Most notably, the admin search bar should allow for partial matches or substring matches i.e. matching in the middle. The admin search bar currently only supports searching for a game by game name. Some additional options that can be implemented are searching or filtering by a game's producer/artists/designers, by category, or by mechanic. A sort feature may also be necessary. A smaller task is to customize the plural names of the models. For example, the model Category is displayed as "Categorys" by default; this should be fixed.
Edited by ptlin25

Abebes01:
Resolving #135 by allowing partial searching and added Meta class to some objects for plural conjugations.


MuyanXie:
@ekabaeva we need your review before we can proceed

ekabaeva (approved):
The list of sorting metrics is sufficient, and it works correctly. Looks good to me!

ekabaeva (changes requested):
The original implementation works perfectly. I would just print the string that is used for the search somewhere after the search is performed because it might be good for a user to see the string they used the most recently (it can be either in a search bar or next to the Games Title). It might be also useful to add a button somewhere that can redirect the user to the original games grid because if they want to sort the whole game grid after they perform the search, they have to go all the way back to the original grid because sorting allows sorting among the search results.

The task itself asks for the feature to be implemented for the admin. Currently, the feature is available for all registered users and I think we should leave it as it is, but we might need to reconsider the task.

Abebes01:
Hello @ekabaeva I have added the requested additions. Let me know if there are any other issues.

ekabaeva (approved):
I tested the implementation, and all the comments were addressed and implemented. Great job!

MuyanXie:
@Abebes01 I not able to verify the changes. When I load the fixtures games-fixture-500.json and the test_players.json, I can login to the games page. However, when selecting sort by rating lowest first, I encountered this error:

FieldError at /games/
Cannot resolve keyword 'review' into field. Choices are: BGG_id, achievement, avg_rating, categories, complexity, description, expected_playtime, game_lists, guide, id, image, interactivefictiongame, leaderboardprivacysetting, leaderboards, lobby, match, matchproposal, max_players, max_playtime, mechanics, metrics, min_players, min_playtime, name, people, popularity, published_guide_id, published_guide_id_id, publishers, rating_percentage, reviews, rules, suggested_age, tournament, twine_file, users, year_published

MuyanXie (changes requested):
see above error messages

Issue #496, opened 2025-04-21
Add checkers board demo to Sandbox


jacobwolf-1:

vliao1223:
Objectives: The goal is to create a functional checkers game board that can serve as a starting point. Afterwards, the code will be added to Sandbox so others will be able to experiment with the code.

vliao1223:
Created and pushed the branch 'adding-games/checkers' to github. I've just created a pull request, but for now, you can find the codes to the checkers game board in the adding-games/checkers branch (in checkers-prototype directory inside sandbox).

jacobwolf-1:
Made PR to add the 2d and 3d boards from testing js libraries to sandbox

abargher:
Issue Score: Excellent

Comments:
Nice job!

PR #576, opened 2025-04-28
[Adding Games] Adding simple boards to Sandbox

jacobwolf-1:
Added simple boards created when testing JS libraries for adding games as per issue #496


vliao1223 (changes requested):
Please fix the linter issues

Issue #535, opened 2025-04-25
Creating Initial Draft of Twine Importation and Dummy Game


zaynacheema:
For this issue we want to figure out how to embed a twine game by creatings a dummy game and then figuring out how to import it onto the site.

Goals:
- Import Twine stories into our platform cleanly 
- Minimize changes required to existing Twine story files during the import process.

Open Questions:
- What format(s) of Twine are easiest to support — HTML files? Source code?
- Can we display the game on the site or would it be better to link and it opens on the browser?

zaynacheema:
Have a dummy project to use in implementation trial:
file:///Users/zaynacheema/Downloads/Midterm%20Project.html


Going to work on altering codebase to allow for html file upload, so the idea is a Twine game developer can export their file as html (which is default and easy to do) and then choose that file in the interactive fiction page to upload it and then can be clicked to go to that game.

zaynacheema:
Note: This importation will be for testing and then will be implemented into the frontend mockup we have created

We should also try to test a game in SugarCube format for save states

zaynacheema:
<img width="827" alt="Image" src="https://github.com/user-attachments/assets/7cfb1bb4-e1bc-4bcd-b00b-26e422e1db30" />

included an option to upload files to the interactive fiction page, accessible by the server at http://127.0.0.1:8000/games/interactive-fiction/

zaynacheema:
once an html file is uploaded successfully (a message will appear at the top as shown below:

<img width="798" alt="Image" src="https://github.com/user-attachments/assets/f3b20516-0159-46d7-9ce4-8e24720a5209" />

now, once it is uploaded there is a button to click to play the uploaded story:
<img width="513" alt="Image" src="https://github.com/user-attachments/assets/1541bec6-fd46-4c2a-a57f-3649bedcb8ef" />

which successfully opens another tab to play the twine game in-browser as normal and it works perfectly

zaynacheema:
The code edits have been pushed to the "interactive-fiction/twine-intergration" branch

 https://github.com/uchicago-cs/chigame/blob/interactive-fiction/twine-integration/docs/developer/local.md

edits were made to:
-  the game-detail.html (under an if block so it will only show up for the interactive fiction page)
-  urls.py 
-  views.py -> edits were made to the InteractionFictionView class display and I added a class for UploadFileView for my edits

zaynacheema:
[...]

Issue #507, opened 2025-04-21
Adding page to host Interactive Fiction Implementation


gkharrell01:
Objective:
Create a page within Chigame that hosts interactive fiction games. This page will serve as a platform for embedding or integrating playable narrative games ( like Twine) to expand what Chigame offers for users.

Potential Steps and Directions:
- Initial Setup:  Add a new route and page component (e.g. /play), create an accessible UI for this page
- Embedding/Other Options: Support .html stories via <iframe>, track progress for Users, allow for admin control

gkharrell01:
Pulled branch from main to explore format and function of existing code base.

gkharrell01:
Pushed changes to git that creates new page under http://127.0.0.1:8000/games/interactive-fiction/ 
This is based on the existing game_detail.html template.

gkharrell01:
Closing comment: 
IF Page has been implemented using the existing game_detail.html template. After discussion, this will be one page of two for IF: the detail view(this) and the create view (a different issue). The current page can be accessed at this tag: http://127.0.0.1:8000/games/interactive-fiction/

Future direction: this template might be changed to create a specific IF template because "matches" are not necessary for IF implementation.

Grace worked solely on this issue.

jacqsmei:
Issue Score: Excellent

Comments:
Great job! Thank you for the super clear closing statement! Next time link the PR before closing

PR #638, opened 2025-04-29
[Interactive Fiction] Adding page to host Interactive fiction games and Twine importation

gkharrell01:
#507 
PR to add page/view to host IF games in the future. This implements the game_detail template, which will either need to be changed or a separate IF template will need to be create to best fit IF games... the current page is accessible under http://127.0.0.1:8000/games/interactive-fiction/ , so whether or not the template is used in the future, the url and page is set up and ready for use.

Also #535 
Twine importation capability with an upload file link was added as well to the host page, as well as the ability to click the upload and open the game in a new tab


zaynacheema:
Note: the last commits in the past 30 minutes starting with "cleaning up views.py" were because the PR were not passing the linter checks, they were not substantial code edits but solely for the purpose of passing the linter.

src/chigame/games/views.py
```diff
@@ -379,7 +383,50 @@ def search_results(request):
...
-# Tournaments
+# =============== Interactive Fiction Views ===============
+class InteractiveFictionView(TemplateView):
+    template_name = "games/game_detail.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        # creates a fake game object
+        fake_game = Game(
+            pk=9999,
```
smagdalen, line 406: This default value of 9999 is good for testing in beginning, assuming that we are going to replace this with real game data in the future. Seems to be no issues here, good work!

src/chigame/games/views.py
```diff
@@ -379,7 +383,50 @@ def search_results(request):
...
+            fs = FileSystemStorage(location=upload_path)
+            safe_filename = uploaded_file.name.replace(" ", "_")
+            fs.save(safe_filename, uploaded_file)
+
+            request.session["uploaded_interactive_file"] = f"interactive_uploads/{safe_filename}"
+
+            messages.success(request, "File uploaded successfully!")
+            return redirect("interactive-fiction")
+
+        messages.error(request, "No file selected.")
```
smagdalen, line 437: This error message here is a little vague and doesn't distinguish the cases that no file is selected, a file exceeds storage requirement or a file is not of a valid type. The rest of UploadFileView is pretty solid.

src/templates/games/game_detail.html
```diff
@@ -90,6 +90,22 @@ <h3 id="new-review">Write a Review</h3>
...
             <button type="submit" class="btn btn-primary">Submit Review</button>
           </form>
+          {% if game.pk == 9999 %}
+            {# for interactive fiction games #}
+            <hr />
+            <h3>Upload a File</h3>
+            <form method="post"
+                  action="{% url 'upload-file' pk=game.pk %}"
+                  enctype="multipart/form-data">
+              {% csrf_token %}
```
smagdalen, line 133: Good use of this token. Code is also easy to read and follow

src/templates/games/game_detail.html
```diff
@@ -90,6 +90,22 @@ <h3 id="new-review">Write a Review</h3>
...
+              <input type="file" id="myFile" name="uploaded_file" />
+              <input type="submit" class="btn btn-primary" value="Upload" />
+            </form>
+            <h3>Uploaded Story</h3>
+            <a href="{{ uploaded_file_url }}"
+               target="_blank"
+               class="btn btn-success">Play Uploaded Game</a>
```
smagdalen, line 140: Might be a good idea to have more checks for safety such as checking the file type/extension (.jpg, .zip, etc) but this is a good start

smagdalen (approved):
Solid work! There are some areas for improvement, but considering that we are establishing a working base, I think this is a good direction.

jacqsmei (approved):
looks good to me! i agree w the previous comments -- things to consider as you continue tweaking the views, but for now this is a good first setup.

jacqsmei:
Issue Score: Satisfactory

Comments:
Nice job! We already talked about this but next time close each issue with its own commit. Also, make sure your commits follow the developer guide!

PR #646, opened 2025-04-30
[Forums] Implement a forum landing page for unauthenticated users

Luca-W-D:
Currently, an unauthenticated user who navigates to the forum page will see an empty forum even if content exists on the backend. This is because the security rules deny logged-out users read permissions. While secure, this leads to a confusing user experience. Therefore, as discussed in #252, it would be better to have a landing page that instructs unauthenticated users to log in. 

This PR (1) implements a landing page for unauthenticated users and (2) redirects unauthenticated redirects forum sub-pages via middleware. An unauthenticated user now receives the following view:
image


src/templates/forum/index.html
```diff
@@ -5,5 +5,9 @@
     <a class="btn btn-primary right-0"
        href="{% url 'forums:forum-create' %}">Create new forum</a>
   {% endif %}
-  {{ block.super }}
+  {% if user.is_authenticated %}
+    {{ block.super }}
+  {% else % }
```
yizou275: Here there is an extra space between "%" and "}"? The page mistakenly shows "  {% else % } / Not logged in" explicitly when I access the page.
image

yizou275 (changes requested):
Hi Luca, thanks for this great work! It seems that the file changes in games folder (removal of favorite_list) are also included in this pr, but not directly related to the issue? It might be a good idea to recover these in this pr and open a separate pr for these unrelated file changes?

Luca-W-D:
That information is super helpful– thank you! I agree that something weird is going on with the Git history. I think that – because this PR touches the /template folder, it won't pull in the more recent dev changes that added these files. I went ahead and manually pulled them in.

I also fixed that typo, I'm hoping that fixes your rendering issue!

yizou275 (approved):
Yes this looks good to me now! Thank you so much for the great work!

Luca-W-D:
Thank you for the in-depth feedback throughout this PR, Yi! Especially catching that last misconfiguration– I would not have seen that. :)

Luca-W-D:
Heart dropped for a second– if you look at the workflow history, the automated tests after the merge failed on the first iteration. I looked at the error logs, and it turns out that one of the automated tests generated two games as fake data with the same name, causing an integrity issue. Fortunately just bad luck rather than a code issue!

jacqsmei (approved):
nice job yall! and great catch Yi! lgtm

jacqsmei:
Issue Score: Excellent

Comments:
This was a great extra task to pick up! Nice teamwork figuring out the review. Great that it finally got done

PR #649, opened 2025-04-30
[User Profiles + Friending] Fix pending friend request status after removing friend

### LINKED ISSUE #601: Fix pending friend request status after removing friend
Currently, when a user removes another as a friend, the interface incorrectly shows options to accept or remove the friend again. This happens because the FriendInvitation object still exists and is misinterpreted as a pending request.
We need to ensure that once a FriendInvitation is marked as deleted, it is no longer treated as pending, and the UI updates accordingly.

- [x]  Confirm deletion status is correctly stored in the FriendInvitation object
- [x] Update logic that checks for pending invitations to ignore deleted ones
- [x] Test UI behavior for both users after unfriending

hossainmd-uc:
This pull request references Issue #601 .

Previously, after unfriending someone, the interface still showed friend request options due to a lingering `FriendInvitation` object that wasn’t properly excluded from logic checking for pending requests. To fix this, I updated the logic so that once a `FriendInvitation` is marked as deleted, it is no longer interpreted as a pending request. Now, when a user removes a friend, the system ensures the old invitation does not surface again in the UI.

While addressing this, I discovered a side issue: if two FriendInvitation objects exist between the same pair of users, removing a friend could throw a `MultipleObjectsReturned` error. This happened because the removal logic did not account for multiple invitations. I resolved this by modifying the query to reference only the latest `FriendInvitation` object when removing a friend, avoiding the error and ensuring consistent behavior.


src/chigame/users/models.py
```diff
@@ -85,7 +85,11 @@ class FriendInvitationManager(models.Manager):
     def get_by_users(self, user1, user2, **kwargs):
         """Gets a friend invitation given two user, which can be a sender
         or a receiver"""
-        return self.get(Q(sender=user1, receiver=user2) | Q(sender=user2, receiver=user1), **kwargs)
+        return (
+            self.filter(Q(sender=user1, receiver=user2) | Q(sender=user2, receiver=user1), **kwargs)
+            .order_by("-timestamp")
```
jonathanbacon22, line 97: Nice idea with the timestamp filtering!

src/chigame/users/views.py
```diff
@@ -131,10 +137,20 @@ def send_friend_invitation(request, pk):
...
+            Q(sender=curr_user, receiver=other_user) | Q(sender=other_user, receiver=curr_user), is_deleted=False
+        )
+        .order_by("-timestamp")
+        .first()
+    )
+    new = False
+    if invitation is None:
+        invitation = FriendInvitation.objects.create(sender=curr_user, receiver=other_user)
+        new = True
+    else:
```
jonathanbacon22: is this else necessary? If new is already set to False, won't it remain false if invitation is not None?

jonathanbacon22 (approved):
I have a comment regarding a minor style change, but besides that this looks great to me! Nice work.

src/chigame/users/views.py
hossainmd-uc: Good catch. Since new is initialized as False it's safe to remove that conditional. Fixed ✅

jonathanbacon22 (approved):
LGTM! Great work and nice fix.

cmarshall282 (approved):
Looks great! Thanks for fixing this!

hossainmd-uc:
I resolved the merge conflicts, so the code changes can now be properly integrated into the dev branch. The recent changes to dev implemented an almost identical solution except for not filtering FriendInvitations in the user_profile_detail_view by timestamp. After merge, issue is ready to be closed.

cmarshall282:
Issue Score: Excellent

Comments:
Great job on this! Here is some feedback:

Issue #520, opened 2025-04-23
Make visual mockups for achievements and achievement views


jiayuew1:
Create a visual mockup for what the achievement page will look like using HTML and CSS. The mockup should show the page that users will see when they want to check which achievements they have gotten for each game they have played.

emilyzen:
Update: we have created an HTML page mockup. The scrum board doesn't allow us to upload HTML files, so I will include a screenshot of the page.

![Image](https://github.com/user-attachments/assets/72845a07-702e-49b3-bf9a-5166dac81bce)

emilyzen:
Closing the issue because the visual mockup is sufficient

abargher:
Issue Score: Excellent

Comments:
Great work!

PR #650, opened 2025-05-01
[achivements] Adding visual-mockup HTMLand trophy icons to sandbox

jiayuew1:
All files are located under chigame/src/sandbox/achievement-mockup.

This mockup uses simple, hard-coded HTML to demonstrate the structure and logic of the achievement page, aligned with our current user stories. The user achievement database is still under development and will be integrated at a later stage.

For now, this mockup serves as a reference for the Frontend and other teams to illustrate the expected layout and basic logic of the achievement page we aim to implement.

To address issue #520


mhenry111 (approved):
Super cool! I don't have any design critiques, but maybe frontend will since they are responsible for the website aesthetic. I think this is good to merge so that they can see it and build on it.

src/sandbox/achievement-mockup/user-achievements-page.html
```diff
@@ -0,0 +1,567 @@
...
+    <style>
+      :root {
+        --bg-light: #f8f9fa;
+        --text-dark: #212529;
+        --card-bg: #ffffff;
+        --header-bg: #f8f9fa;
+        --accent: #0d6efd;
+      }
+
+      body {
```
abargher, line 29: This looks great! It's fine as-is for this mockup, but going forward, I think the CSS, and HTML (and probably JS too) should be separated into their own files.

abargher (approved):
Looks good, great work!

abargher:
Issue Score: Excellent

Comments:
Excellent work!

Issue #582, opened 2025-04-28
Perform chat app channels configuration


aflekkas:
This relies on #512 as well as #488. Once those issues are resolved I can begin configuring the new `chat` app to support channels etc and its ASGI so that we can begin implementing all of the chat features.

aflekkas:
Hi, I'll be doing the basic setup now - will create a branch for this - also pruned some old branches

abargher:
Issue Score: Excellent

Comments:
Good job!

PR #652, opened 2025-05-01
[LiveChat] Migrated LiveChat models from games to chat

### LINKED ISSUE #583: Migrate models from games to chat
Models were previously created and added to games by issue #491

If we are to migrate to `chat` these models should be then moved there

pmdepaulo:
Added str() representations for the models and added a ManyToMany field to access users from a LiveChat model


aflekkas (approved):
Looks good to me. Let's ship it. ￼￼￼

pmdepaulo:
Issue #583

aflekkas:
@pmdepaulo I made a couple changes

image

I would say this is the most important, turns out we don't need `channel` - we can just use the ID of the chat and provide a display name for the chat visually

The channel should just be unique - we can just assign a channel based on the ID

Also changed the `message_content` field to just `content` - minor fixes

aflekkas:
Other changes included like setting up the app properly (something I had not done), this PR combined with #582 should ensure that the app is fully configured. I also made migrations

aflekkas:
@pmdepaulo Also code I pushed was not passing some CICD checks (just fixed)

abargher (changes requested):
Please fix the CI check-- looks like an issue with outdated migrations again.

abargher (approved):
Looks like that fixed it, thanks!

abargher:
Issue Score: Excellent

Comments:
Great job!

Issue #631, opened 2025-04-28
migrate models


mhenry111:
make migrations once models are finalized

PR #660, opened 2025-05-02
Add necessary files to achievements directory

bensim0305:
To address issue ~~#631 , I made migrations to the src/achievements folder for the achievements model. Also included chigame.achievements in the config's base.py LOCAL_APPS and added a __init__.py and app.py for setting up the achievements directory.~~ #666, added a __init__.py, app.py, tests.py, views.py, admin.py, and urls.py.  Updated PR with correct branch name can be found at #667


mhenry111:
everything looks good but for tracking purposes i feel like adding the app.py and init.py could have been part of a separate issue. can we actually make a new issue like 'create necessary files for achievements directory' and then you can link this PR to that issue instead? and then also add files like admin.py, tests.py, urls.py, views.py...whatever we need

mhenry111:
i created an issue for this. #666

bensim0305:
Updated PR to address issue #666 and added the necessary files

mhenry111 (approved):
Looks great! Thanks for the update.

PR #674, opened 2025-05-04
Added checkers model to models.py

### LINKED ISSUE #647: Build data models for checkers according to ERD
Create Django models for the checkers ERD on the wiki. Once completed these models should go in the models.py file in src/chigame/games.

enzeeo:
Added three different entities for the checkers model based mostly on the ERD although some things were changed to provide further clarity (e.g. timestamps became start_time and end_time, and the turn now tracks which player's turn it is) 

The Checkers entity tracks the players, winner, start time, and end time. I have also commented out a _bot_ as we have not created a model for this yet but theoretically the Checkers entity would also keep track of this. _Player_2_ is nullable as we can have a single player game. 

The CheckersBoard tracks the state of the board. As of right now this is tracked using a list of list. Whether that be integer values or dict values at each index is still up for discussion. I have also commented out a bit representation called _state_bits_ in the case we want to optimize the storage space in the future. 

The CheckersTurn stores the game from _Checkers_, the board from _CheckersBoard_, the turn (an incrementing integer each time either player makes a move) and a _Player_ which tracks which player is currently making the move. 

Addresses issue #647


nneidlinger (approved):
Looks good to me! I think the names for the models and attributes are clear. The models also seem to accurately reflect the original ERD with any deviations from the original ERD having good reason.

abargher (changes requested):
Please fix the failing CI check. Otherwise, the models look good!

brianHarder:
I fixed the CI check by using the related_name field in the Checkers model.

abargher (changes requested):
Looks like one more migration merge to do, hopefully should be a quick fix. Sorry this has happened more than once here!

brianHarder:
Migration merge resolved.

brianHarder:
Migrations should be set now.

abargher (approved):
Looks good! Models are well thought out.

abargher:
Issue Score: Excellent

Comments:
Great work!

Issue #630, opened 2025-04-28
Finalize achievement models


mhenry111:
The plan is to add a couple new attributes to our models and think of any additional entities and/or functions that belong in the achievements/models.py file. The end goal of this issue is to commit the final achievement models

mhenry111:
thinking about making an AchievementType class/list with types like **progress** for time-based achievements, **multiplayer** for multiplayer only achievements, **quantity** for things like 5 wins, etc. could be useful if most games plan to have achievements for things like 10 day streak, 5 wins, etc.

mhenry111:
might make slight model updates later in a separate issue; for now, Achievements have thresholds and UserAchievements have a progress tracking attribute

abargher:
Issue Score: Excellent

Comments:
Great work!

PR #682, opened 2025-05-04
[achievements] added new models and updated existing ones

mhenry111:
to address issue #630 I updated existing achievements models to account for multi-player logic from our user stories. I also introduced some new models intended to account for stats tracking. These models will be helpful for tracking a user's progress towards an achievement if applicable, as well as for tracking general game stats


bensim0305:
How would the userAchievementsStat be tracked exactly? Would the game like provide it's own metric of tracking user progress?

mhenry111:
will probably remove the stats models for now and discuss tracking progress/metrics with Games team

mhenry111:
got rid of all the stats stuff so the only thing that is really new is that now Achievements have an optional threshold attribute and UserAchievements have an optional progress attribute so that if it is a progress-based achievement, the game can update the user's progress incrementally. Because of this, I also got rid of auto_now_add in the date_earned attribute of UserAchievements because if it is a progress-based achievement, then the date that the UserAchievement is created isn't necessarily the date that it is earned.

mhenry111:
got rid of all the stats stuff so the only thing that is really new is that now Achievements have an optional threshold attribute and UserAchievements have an optional progress attribute so that if it is a progress-based achievement, the game can update the user's progress incrementally. Because of this, I also got rid of auto_now_add in the date_earned attribute of UserAchievements because if it is a progress-based achievement, then the date that the UserAchievement is created isn't necessarily the date that it is earned.

bensim0305 (approved):
Looks good to me!

abargher (approved):
Looks good, I think these additional fields should be enough to support the use cases we discussed in the last meeting, although feel free to add more of them if it would make things easier-- I'm open to either approach.

abargher:
Issue Score: Excellent

Comments:
Nice work!

PR #689, opened 2025-05-04
[User Profiles + Friending] Allow users to bookmark notification 

### LINKED ISSUE #600: Allow notifications to be bookmarked
Currently, notifications can not be bookmarked. The plan is to add a new feature that allows notifications to be bookmarked.

We know that users may be getting lots of notifications, and allowing them to bookmark important ones is something we believe will serve users well. There are several different types of notifications, and for users, some notifications may be more important by others. We want users to be able to then have a separate place in which they can view their bookmarked notifications.

jonathanbacon22:
This pull request references Issue #600. The changes include adding a new field to the Notification model to toggle whether or not a notification was bookmarked. I also added two new views, to support users seeing their bookmarked notifications and bookmark their notifications. I also added two new field to urls.py after creating the views.


src/chigame/users/models.py
hossainmd-uc, line 1: Bookmark field looks good!

src/chigame/users/models.py
```diff
@@ -247,6 +247,7 @@ class Notification(models.Model):
     actor = GenericForeignKey("actor_content_type", "actor_object_id")
     message = models.CharField(max_length=255, blank=True, null=True)
     objects = NotificationQuerySet.as_manager()
+    bookmarked = models.BooleanField(default=False)
```
hossainmd-uc, line 293: Bookmark field looks good!

src/chigame/users/views.py
```diff
@@ -364,3 +364,38 @@ def bulk_inbox(request):
...
+        "pk": pk,
+        "user": user,
+        "notifications": notifications,
+        "default_notification_messages": default_notification_messages,
+    }
+    if pk == user.id:
+        return render(request, "users/bookmarked_notifications.html", context)
+    else:
+        messages.error(request, "Not your inbox")
+        return redirect(reverse("users:user-profile", kwargs={"pk": request.user.pk}))
```
hossainmd-uc: Looks great!

hossainmd-uc (approved):
Overall, it looks like this should work. 

Next Step: Make template to integrate the view and test it out

src/chigame/users/views.py
```diff
@@ -364,3 +364,38 @@ def bulk_inbox(request):
...
+        messages.error(request, "Notification does not exist.")
+
+    return redirect(reverse("users:user-inbox", kwargs={"pk": request.user.pk}))
+
+
+@login_required
+def view_bookmarked_notifications(request, pk):
+    user = request.user
+    notifications = Notification.objects.filter_by_receiver(user).filter(bookmarked=True)
+    default_notification_messages = Notification.DEFAULT_MESSAGES
```
cmarshall282, line 666: What's the point of default messages here?

cmarshall282 (approved):
Had one question, but otherwise looks great!

src/chigame/users/views.py
jonathanbacon22: I used the deleted_notifications_view function as a sort of template here. In lines 397 and 398, I use the context, which default_notification_messages is part of, to render the request. Is that not needed? I can remove.

cmarshall282: That should be fine I think!

cmarshall282:
Issue Score: Excellent

Comments:
Great job on this PR!

PR #691, opened 2025-05-05
[Users Profiles + Friending] Updated Tables to display friends with display names

### LINKED ISSUE #560: [Users+Profiles] Friend List Displaying User Emails
The friend list display shows users by email, which is not only a potential privacy violation, but also not the most intuitive friendship display. This should maybe be changed to Display Name for clarity and privacy

harriet-fisher:
To address issue #560 :

- Replaced the email column with a new display_name column to show friend names instead of emails
- Updated the column’s accessor from "email" to "userprofile.display_name" to fetch the name from the related UserProfile.
- Adjusted linkify to still link to the user’s profile using pk to preserve navigation
- Removed the "Email" header by setting verbose_name="" and hid the table header with thead: { "class": "d-none" } for cleaner UI since it's currently a single column table
- Ensured the fields list matches the new column ("display_name" instead of "email")


hollyfwei (approved):
Looks great - thanks, Harriet!

jonathanbacon22 (approved):
I am a little confused by all the changes happening with the migrations. But overall these changes look got to me! Nice work!

src/chigame/users/models.py
```diff
@@ -108,12 +108,16 @@ class FriendInvitation(models.Model):
     objects = FriendInvitationManager()
     is_deleted = models.BooleanField(default=False)
 
+    class Meta:
+        unique_together = ("sender", "receiver")
+
```
davdma: @harriet-fisher The uniqueness constraint here is problematic with the current dev implementation which only does soft deletes for invitations (which we agreed was the best way to go). There can exist multiple invitations from the same sender to receiver with one `is_deleted = True` and many possible `is_deleted = False` (if there have been declined invitations in the past). The solution we have instead of using uniqueness constraint is to verify that there is only one active friend invitation from the sender to receiver at one time. This is checked in the `views.py` `send_friend_inviation` function, so that if there already is an active invitation, a new invitation cannot be created.

davdma (changes requested):
Nice work Harriet! Most of the code looks good to me except for changes to the invitation model that conflicts with the dev implementation of soft deletes. If you just remove the uniqueness constraints then it should be fixed. Also, I am still thinking about how username plays a role in all this. Is the username redundant now because of the display name?

cmarshall282 (changes requested):
Thanks for doing this! Can you address david's comment and remove all of the changes to the migrations (since they don't seem to impact anything)?

harriet-fisher:
The uniqueness constraint was addressed and migration errors fixed.

hossainmd-uc (approved):
LGTM, thanks for addressing the uniqueness constraint issue!

cmarshall282 (changes requested):
In the last few weeks, we removed display_name. Can we update this PR accordingly?

harriet-fisher:
Updated table to use User.username and integrated both table and card display for backward integration.

cmarshall282 (approved):
Looks great!

src/chigame/users/models.py
davdma: Thank you for resolving the constraint harriet!

davdma (approved):
Everything looks good to me now Harriet! Thanks for making the requested changes.

cmarshall282:
Issue Score: Excellent

Comments:

PR #692, opened 2025-05-05
[User Profiles + Friending] Implemented Profile Display Name and Bio Editing

### LINKED ISSUE #559: [Users+Profiles] Profile Updating Debug: Display Name and Bio
Right now, users can update their name and email via profile editing, but not their bio. Users should be able to edit and save their bios (fyi, bios are attributes of UserProfile). Additionally, there is a display_name attribute of the UserProfile that also has no updating field, this one is especially important as their display name is their primary name on the platform.

harriet-fisher:
To address issue #559:

User Account Editing (user_detail.html):
- Shows basic user information (name, email, display name, bio) added display name 
- Has an embedded form to update the user's name (instead of redirecting to ~update page)
- Uses UserUpdateView to handle name updates
Profile Editing (user_form.html):
- Shows a form with fields for display name and bio
- Uses the same UserUpdateView but handles profile fields separately
- Redirects back to profile page after update
View Logic (UserUpdateView):
- Handles both user and profile updates in one view
- Uses fields = ["name"] for the user model form
- Manually handles profile fields in form_valid
- Handles cases where profile doesn't exist (API
URL Structure:
- ~update/ - Now Handles both user and profile updates
- account_email - Handles email updates
- user-profile/<int:pk>/ - Shows profile page


src/chigame/users/tables.py
```diff
@@ -4,16 +4,26 @@
...
+        accessor="userprofile.display_name",
         linkify=("users:user-profile", {"pk": tables.A("pk")}),
     )
 
     class Meta:
-        model = User  # Referencing the UserProfile model
+        model = User
         template_name = "django_tables2/bootstrap.html"
-        fields = ["email"]  # Adjust fields to show relevant information from the UserProfile model
+        fields = ["display_name"]
```
jonathanbacon22, line 20: nice change! I agree it makes more sense to show the display_name and not email.

jonathanbacon22 (approved):
Changes look good to me. Great work!

src/chigame/users/models.py
```diff
@@ -108,13 +108,18 @@ class FriendInvitation(models.Model):
     objects = FriendInvitationManager()
     is_deleted = models.BooleanField(default=False)
 
+    class Meta:
+        unique_together = ("sender", "receiver")
+
```
davdma: As I also mentioned in a separate PR, to work with the soft delete dev implementation, we need to remove the uniqueness constraint of sender and receiver in the table.

src/templates/users/user_form.html
```diff
@@ -3,18 +3,30 @@
 {% load crispy_forms_tags %}
 
 {% block title %}
-  {{ user.name }}
+  Update Profile
 {% endblock title %}
 {% block content %}
-  <h1>{{ user.name }}</h1>
+  <h1>Update Profile</h1>
```
davdma, line 9: I like this template format better - "update profile" instead of just the user's name.

src/templates/users/user_detail.html
```diff
@@ -28,23 +37,30 @@ <h2>My Account:</h2>
...
             <li class="list-item">
-              Email: {{ object.email }}
-              <a class="btn btn-primary"
-                 href="{% url 'account_email' %}"
-                 role="button">Update E-Mail</a>
+              Email: {{ object.email }}<a class="btn btn-primary btn-sm update-button"
+   href="{% url 'account_email' %}"
+   role="button">Update E-Mail</a>
             </li>
+            <li class="list-item">Display Name: {{ object.userprofile.display_name }}</li>
```
davdma: This is just something I am thinking of so no need to act on this, but we could also have some placeholder text for updating the bio if they have empty bios (but this might require custom submission forms).

davdma (changes requested):

src/chigame/users/models.py
davdma: Everything else looks good though!

cmarshall282 (changes requested):
Can we make this PR only contain relevant changes instead of having changes in both #691 and this pr?

src/templates/users/user_form.html
```diff
@@ -3,18 +3,30 @@
...
         action="{% url 'users:update' %}">
     {% csrf_token %}
     {{ form|crispy }}
+    <div class="form-group">
+      <label for="display_name">Display Name</label>
+      <input type="text"
+             class="form-control"
+             id="display_name"
+             name="display_name"
+             value="{{ profile.display_name }}" />
```
davdma, line 21: Just a note but my recent PR that merged into dev removed the redundant display name so during the merge I would make sure to get rid of references to it like here.

davdma (approved):
Looks good to me @harriet-fisher! Thanks for fixing the uniqueness constraint

harriet-fisher:
This PR is stale, as conflicting implementations have been merged.

PR #698, opened 2025-05-05
Userprof/notifications

hollyfwei:
Added handlers for achievement and tournament notifications & added user path for usergroupsview


src/chigame/api/serializers.py
```diff
@@ -84,7 +84,8 @@ def create(self, validated_data):
 class GroupSerializer(serializers.ModelSerializer):
     class Meta:
         model = Group
-        fields = "__all__"
+        fields = ["id", "name", "description", "members", "group_admin_permissions"]
+        read_only_fields = ["created_by", "date_created"]
```
jonathanbacon22, line 90: nice change here!

src/chigame/api/views.py
```diff
@@ -102,6 +102,14 @@ def get_object(self):
         return get_user(lookup_value)
 
 
+# Custom permission class for authentification
+class IsAuthenticatedOrReadOnly(generics.BasePermission):
+    def has_permission(self, request, view):
+        if request.method in ["GET", "HEAD", "OPTIONS"]:
```
jonathanbacon22, line 138: Nice thinking! I like how you are checking and returning true if a read operation is being carried out.

jonathanbacon22 (approved):
This looks great to me! Nice creativity and effective changes!!

davdma (approved):
LGTM Holly! Thanks for expanding on the notifications for tournament handling.

cmarshall282 (approved):
Thanks for adding these types

cmarshall282:
Issue Score: Needs Improvement

Comments:
Here is some feedback for this PR:

- The title of an issue has to convey what the issue will be producing. In the future, try using simple titles with a "Verb-Object" structure that clearly states what you will be doing in the issue (e.g., “Design and implement a ‘circle’ module”, “Research libraries for drawing simple geometric shapes”, etc.)
- Even though you can add labels to issues to categorize them, you should avoid titles that could plausibly apply to multiple teams
- The summary for this issue is too terse. A good rule of thumb is that you should be writing at least one full paragraph in your summary (see https://uchicago-cs.github.io/cmsc22000/project/issue-pr-grading.html#descriptive-summary for examples). If this issue was generated from a Scrum board card, please remember that the summary you write in the Sprint meeting is meant to be a quick draft; it needs to be fleshed out when you convert the card to an issue.

PR #744, opened 2025-05-06
Live chat/component base

### LINKED ISSUE #527: Convert chat page into basic component & add channel recognition
Once issue #526 is resolved, creating a chat component that is styled the same and depends on which channel is passed in to it

aflekkas:
This is to close issue #527. In addition to this, we added the ability to save messages to chat.


pmdepaulo (approved):
this looks great! love the demo as well— only thing is that copilot has a message about the save_message function not actually returning to the db. it looks like you guys do create an instance but it’s worth double checking.

aflekkas:
> this looks great! love the demo as well— only thing is that copilot has a message about the save_message function not actually returning to the db. it looks like you guys do create an instance but it’s worth double checking.

I think I addressed the issue in my latest commit! If not I'll initialise another PR to fix. TY for your review!

aflekkas:
Just fixed a typo, should be good for @abargher to review!

src/templates/chat/index.html
```diff
@@ -1,63 +1,7 @@
...
-        document.querySelector("#id_message_send_input").value = "";
-        document.querySelector("#id_chat_item_container").appendChild(div);
-      };
-    </script>
-  </body>
-</html>
+{% extends "base.html" %}
+
+{% block content %}
+  {% include "chat/components/_header.html" %}
```
abargher, line 4: I like the split here into separate files, makes it much easier to read.

abargher (approved):
Looks good, nice work!

abargher:
Issue Score: Excellent

Comments:
Great job!

Issue #640, opened 2025-04-29
Design white-label knowledge base interface


Luca-W-D:
While most users will interact with the ChiGame knowledge base through its primary interface, some game developers may eventually want to use the platform as a general wiki for their game. A small game studio, for example, may not be able to self-host its wiki elsewhere. To support this use case, we should allow contributors to generate a URL to their guide that, when opened, contains minimal ChiGame UI. This can then be distributed for broader use outside the ChiGame ecosystem. While there may still be links / references to ChiGame, the interface should present almost exclusively their game guide.

davidmakoyo:
I will use a query parameter (?minimal=true) which will allow any existing URL to be converted to minimal mode

davidmakoyo:
added new template minimal_content.html, it is a minimal version of the regular chigame UI. 

Modified views.py, added minimal UI mode detection via query parameter

Updated urls.py, added new URL pattern for minimal mode

I pushed these changes to the knowledge-base/white-label branch

davidmakoyo:
<img width="1224" alt="Image" src="https://github.com/user-attachments/assets/d8c1ed85-bbde-49b6-94b4-30dcb1818903" />

davidmakoyo:
<img width="1445" alt="Image" src="https://github.com/user-attachments/assets/c8935be6-f701-45c1-864b-86321bd536a2" />

davidmakoyo:
We could also retain the current UI — including the multi-page navigation and table of contents — and simply remove all chigame branding in the final implementation of the minimal UI. The core change I introduced was leveraging the query parameter, and the only difference between the regular and minimal UI can be the absence of chigame branding if we'd like. I will make that change in a future issue after we've integrated our code into chigame

davidmakoyo:
Closed as minimal mode has been fully implemented as described above and in PR #751 which has been merged

jacqsmei:
Issue Score: Excellent

Comments:
Great work on this! Good considerations to have

PR #570, opened 2025-04-28
[knowledge-base] multi-page markdown rendering

davidmakoyo:
Implemented initial multi-page markdown rendering with section navigation and created an example guide for testing. All changes included in this PR are part of issue #513
image

PR #751, opened 2025-05-06
[Knowledge Base] Created white-label mode for guides

davidmakoyo:
These change accomplish the goal of issue #640. You can flip any knowledge-base page into a “white-label” view by adding ?minimal=true (or by hitting the /minimal/ route). Under the hood the view detects that flag and swaps out your normal template for minimal_content.html, which extends your base layout but strips out all ChiGame navigation, headers, footers, and branding—leaving only your guide’s HTML wrapped in a simple, unbranded container with light styling. This gives you a clean, embeddable wiki page without the full app chrome

You can test this by pulling knowledge-base/white-label, starting a server, and adding ?minimal=true to the url. it should then look like this:

image

### New Files

- **`src/sandbox/markdown-demo/md_app/templates/md_app/minimal_content.html`**
  - A stripped-down template that extends the normal base layout.
  - Removes navigation, sidebars, footers, and branding.
  - Renders `{{ html_content|safe }}` inside a centered, lightly-styled container.

- **`src/sandbox/markdown-demo/md_app/urls.py`**
  - Added route:
    ```python
    path("minimal/", views.markdown_content_view, {"minimal": True})
    ```
  - Allows access to `/minimal/` (optionally with `?section=...`) for minimal mode without needing to manually add query parameters.

### Modified Files

- **`src/sandbox/markdown-demo/md_app/views.py`**
  - Added minimal mode detection:
    ```python
    is_minimal = request.GET.get("minimal", "false").lower() == "true"
    ```
  - Updated template selection logic:
    ```python
    template_name = "md_app/minimal_content.html" if is_minimal else "md_app/standard_content.html"
    ```
  - All other markdown-rendering logic remains unchanged.


Luca-W-D:
Looks really cool, David! I ran the tests that you described, and everything worked as expected. To anyone replicating this in the future, make sure to switch `?minimal=true` to `&minimal=true` if you already have GET parameters included in the URL. 

One thing I did notice is that the minimal version of the UI lacks the table of contents sidebar that the other version includes; is this intentional?

Also, I'll hold off on submitting my review until `dev` gets pulled in. This includes changes from #570 (which touches the same code). That way, I can review the final code more diligently. Overall, looks great!

davidmakoyo:
Thanks for your comments Luca, the sidebar isn't included in this version of the minimal mode just because I initially took the title "minimal mode" a bit literally and also just wanted to differentiate it from the regular UI. Once we merge this code with chigame and it has chigame branding on it i will add the sidebar back and the only function of minimal mode will be to remove chigame branding.

src/sandbox/markdown-demo/md_app/views.py
```diff
@@ -102,23 +106,23 @@ def markdown_content_view(request):
...
-    html_content = md.convert(MARKDOWN_STRING)
+    html_content = md.convert(content)
+
+    template = "md_app/minimal_content.html" if is_minimal else "md_app/markdown_content.html"
 
-    # 4) Pass both the HTML and our manual TOC into the template
     return render(
         request,
-        "md_app/markdown_content.html",
+        template,
```
Luca-W-D, line 126: This is a cool idea for conditional rendering; I think we should keep this in mind for when we are implementing other knowledge base pages.

Luca-W-D (approved):
That makes sense– thanks for the response. I pulled the code locally again, and everything still works as expected. Overall, this just looks super clean and is well implemented. Thanks David!

jacqsmei:
nice job! this looks like a cool feature to have

on the topic of the sidebar:
looks like being in minimal mode gets rid of the idea of a table of contents sidebar completely, putting all sections together into one doc? is that the expected/desired result?
also, doing something like `http://127.0.0.1:8000/?section=combat-basics?minimal=true` would just get rid of all the (necessary) text on the page, but doing `http://127.0.0.1:8000/?section=combat-basics&minimal=true` just brings back the full view, with no regard for which section was selected.

What do you intend to be the expected result?

davidmakoyo:
1. Yes, for now. The sidebar isn't included in this version of minimal mode because I initially interpreted "minimal" literally and wanted to clearly distinguish it from the regular UI. Once we merge this with chigame and the branding is in place, I’ll reintroduce the sidebar—at that point, the only purpose of minimal mode will be to remove the chigame branding.

2. That should never happen unless a Chigame user manually appends ?minimal=true to the URL. Minimal mode is intended for developers using Chigame to host game guides without any Chigame branding. In this setup, users would only access the guide via a direct link like http://127.0.0.1:8000/?minimal=true, and since there’s no table of contents in minimal mode, they wouldn’t be able to navigate to any section URLs from there.

Per issue #640: While most users will interact with the ChiGame knowledge base through its primary interface, some game developers may eventually want to use the platform as a general wiki for their game. A small game studio, for example, may not be able to self-host its wiki elsewhere. To support this use case, we allow contributors to generate a URL to their guide that, when opened, contains minimal ChiGame UI. This URL can then be distributed for broader use outside the ChiGame ecosystem. While there may still be some links or references to ChiGame, the interface should present almost exclusively their game guide.

That said, I have changed it so that any url with minimal mode query parameter just displays the full guide (section parameter is ignored when in minimal mode) thank you for pointing this out!

jacqsmei (approved):
sounds good to me

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! This was maybe a unique but very helpful idea to implement

Issue #687, opened 2025-05-04
Implement Knowledge Base Contributor Manage Uploaded Guides Page


mrollinswaterman:
Similar to Issue #656, this issue should write HTML/JS that matches with the prototyped frontend designs created in Issue #508. Refer to PR #654 for the template we need to follow in order to sync up with backend development.

davidmakoyo:
Added Create New Guide button

davidmakoyo:
Fixed Linter errors

mrollinswaterman:
Thanks for your help with the linter errors David. I'll close this issue up once the PR is approved.

yizou275:
We decide to expand this issue now, to include both frontend and backend. Here is the new description:

As described by knowledge base user stories, a contributor accesses a "manage upload guide" page that lists all their previous uploaded guides and the status. This issue implements this page.

There are two sub-tasks:
1. Creating the HTML template file and any associated static assets (CSS, etc.)
2. Creating the backend interactions (mainly the view() function) so that the frontend pulls data from ChiGame

yizou275:
I'm now planning to wait for PR #754 to get merged first.
After that, I could refine the backend logic of the page. That might include:
- Removing the placeholder guides in the HTML page
- Associating uploaded guides with request.user
- Adding a "resubmit" button for guides with "requested change" status and link to the reupload page
- Possibly allowing contributors to download their previously uploaded guides (if time allows)

yizou275:
Update on backend by PR #819. Completed:
- connect HTML with backend data on authored guides
- reorganize previous HTML/CSS by `<table>`
- add "view detail" link for all guides NOT in "pending" status and link to its recentest feedback; add "resubmit" link for all guides in "requested change" status (the page access is only for users who author the guide)
- can download all guides (only allow for users who author the guide)
- create a FeedbackDetail page that shows info of the respective piece of feedback

Future improvement:
- Currently a contributor only sees links to the recentest update of an uploaded guide. Maybe future we can support history of feedbacks for one single guide.
- Actually not sure whether the details of feedback should be displayed on a separate page (as implemented now), or maybe a pop-up window or some other prettier options.

jacqsmei:
Issue Score: Excellent

Comments:
Awesome teamwork on this! Cool to see your page ideas come to life

PR #754, opened 2025-05-06
[Knowledge Base] Created an HTML page for the contributors to managed uploaded guides

davidmakoyo:
As per Issue https://github.com/uchicago-cs/chigame/issues/687, we created an HTML page for the "Manage Uploaded Guides for Contributors" UI Prototype view, designed here: [https://www.figma.com/design/A36Yb2ijf1hfco1kcOKa8a/Chigame-Knowldege-Base?node-id=1-87&t=iu864mHEEQcp6PU1-0](https://github.com/uchicago-cs/chigame/pull/url)

- **`src/templates/knowledge-base/contributor.html`**
  - Created and refined the contributor page template.
  - Moved inline styles to CSS classes for cleaner design and reusability.

- **`src/chigame/knowledge_base/urls.py`**
  - Added or updated the URL route to support the contributor view.

- **`src/chigame/knowledge_base/views.py`**
  - Implemented or modified view logic to render the contributor page.

- **`src/templates/knowledge-base/base.html`**
  - Adjusted layout and navigation to support the contributor UI.
  
  Test by pulling knowledge-base/contributor-view-frontend, running server in root of chigame directory, and running server. URL: http://127.0.0.1:8000/knowledge-base/contribution. It should look like this:
  
  
image

Buttons are not functional yet as this is only the frontend implementation


src/templates/knowledge-base/base.html
```diff
@@ -19,9 +19,15 @@
...
         </a>
-        <a href="{% url 'knowledge-base-contributor' %}">
-          <p class="mb-0">Contributor View</p>
-        </a>
+        {% if user.is_authenicated %}
+          <a href="{% url 'knowledge-base-contributor' %}">
+            <p class="mb-0">Contributor View</p>
+          </a>
+        {% else %}
+          <a href="{% url 'knowledge-base-contributor' %}">
```
Luca-W-D, line 27: Is this supposed to be the same for both conditions (i.e. to be implemented later in subtask 2)? Or are there supposed to be differences in rendering?

Luca-W-D:
I'll wait for your response to the comment above (edit: comment referenced here was moved to a subcomment of my review) before submitting my review. In the meantime, I have a few of stylistic notes. Of course, this part is subjective, so I won't request changes based on the following comments:

1. The header is a little difficult to read accessibility wise. I'd recommend lightening the background, making the text's weight a bit heavier, and maybe using shadows to signify element's hierarchy than color alone
2. The left and right borders are a bit heavier than those at the top and bottom of the graph
3. I think that the use of a `<table>` element might make the DOM slightly more parseable

I like the addition of the 404, and – of course – thank you for separating CSS into a separate file! :)

davidmakoyo:
Agreed, these changes will be implemented in subtask 2

Luca-W-D (approved):
In that case, sounds good! Thanks for the work, David & Miles! :)

src/templates/knowledge-base/base.html
Luca-W-D: ^ Resolved by David's comment.

jacqsmei (approved):
looking good! some responsiveness issues (table looks funky on smaller screens) but that can be an issue you tackle after getting functionality down

jacqsmei:
also 110 BCE is a fun touch

jacqsmei:
Issue Score: Excellent

Comments:
Nice teamwork on this! For future ref it would also be helpful to include a screenshot of the important Figma contents as links can die

PR #756, opened 2025-05-06
[knowledge-base] Create Fixture with Dummy Data

### LINKED ISSUE #752: Create a Fixture with Dummy Data for Knowledge Base
Before starting the frontend and backend tasks on knowledge base, a fixture file containing dummy data is necessary for us to test our code. I'll create a JSON fixture that contains several guides (by different users, for different games, and in different status), and several reviewfeedback objects (for different guides, in different status).

yizou275:
This PR is mostly creating fixtures for testing purposes and also modifies models.py when encountering problems in creating the fixture.

File Changes:
- Created `kb_user_game_data.json`, `kb_guides_data.json` and `kb_feedbacks_data.json` in `knowledge_base/fixtures` folder (according to issue comment at https://github.com/uchicago-cs/chigame/issues/752#issuecomment-2853202709). They contain `User` + `Game` objects, `Guide` objects, `ReviewFeedback` objects respectively.
- Modified `knowledge_base/models.py` to make `like`/`favorite` field in `Guide` object to be optional. Otherwise, the Admin interface forces me to choose users that likes/favorites a guide. (included the related migration)
- Modified `knowledge_base/models.py` to limit the choices of `reviewer` field in `ReviewFeedback` object to those users with `moderator=True`. Avoids confusion or mistake when creating a `ReviewFeedback` object from Admin interface.

To test it, in a fresh repository (with no prior data), do 
```python manage.py loaddata src/chigame/knowledge_base/fixtures/kb_user_game_data.json \ src/chigame/knowledge_base/fixtures/kb_guides_data.json \ src/chigame/knowledge_base/fixtures/kb_feedbacks_data.json```.
Try if you can log into these users with given password.
Try going to the admin interface, and check whether the objects are there. (Seems that the list of "email addresses" in Admin interface is not the same as "users" list, so make sure to look at the "users" list.)

Also appreciate any comment if the test cases need to be modified, or if my objects differ from my documentation:)

### CODE REVIEW

File: src/chigame/knowledge_base/fixtures/kb_user_game_data.json [hunk 1]
```diff
@@ -0,0 +1,144 @@
...
+        "email": "alice@example.com",
+        "username": null,
+        "tokens": 1,
+        "moderator": false,
+        "groups": [],
+        "user_permissions": []
+    }
+},
+{
+    "model": "games.game",
```
Luca-W-D, line 79: I'm a bit hesitant to have fixtures for other tables combined into one file; will elaborate in review.

Luca-W-D (changes requested):
Looks great! I noticed that, besides the main functionality of this issue, you also cleaned up our models– I especially like how you enforce reviewers to only be moderators.

I ran the tests as described, and everything works flawlessly. I had a few minor recommendations, though, regarding organization:
1. You include `games` in the knowledge-base fixtures. Considering there are already-defined game fixture files, could we use that data instead? This may help the codebase be more readable.
2. Could you add a README to the knowledge base with the commands included above? I think this will be helpful for future developers who will use these fixtures while developing.
3. As an addendum to the previous point, I had to remove the backslashes to get the `loaddata` command to work locally; maybe include both versions?

Overall, great work– thanks Yi! Super excited to use this for our coming tasks! :)

yizou275:
Thanks Luca for this quick and helpful feedback! Regarding 1, I think it's a good idea to load games-fixture-5.json; though one issue is that the current game fixture hasn't included the new `published_guide_id` field yet. Any insights on how to deal with this? Maybe I'll try out a way to load the game fixture data, and then send a post_migration signal so that certain games are linked to certain guides afterwards? 
Also, I think we have similar problems regarding User object (it's also from another app). Do we need to do anything about this? This may be harder to change, as I don't know their passwords. I did notice that the game fixtures contain user objects as well.
And regarding 2 and 3, sure! I'll add a README.md file. Thank you so much!

Luca-W-D:
> I did notice that the game fixtures contain user objects as well.

Given the challenges you identify and the fact that there's precedent for this decision, I think that we can keep the fixtures as-is. Thanks for the insight!

yizou275:
Regarding 3, would this work on your end?
```bash
python manage.py loaddata \
src/chigame/knowledge_base/fixtures/kb_user_game_data.json \
src/chigame/knowledge_base/fixtures/kb_guides_data.json \
src/chigame/knowledge_base/fixtures/kb_feedbacks_data.json

Luca-W-D:
Sadly not– something weird happens with the backslashes. It doesn't recognize `kg_user_game_data.json` in that form. I used...
```python manage.py loaddata src/chigame/knowledge_base/fixtures/kb_user_game_data.json src/chigame/knowledge_base/fixtures/kb_guides_data.json src/chigame/knowledge_base/fixtures/kb_feedbacks_data.json```
...(one line) instead.

Luca-W-D (approved):
Looks great, thank you for the in-depth responses! Also, I love the table rendering on the README.

jacqsmei (approved):
nice job! i also love the table rendering

jacqsmei:
Issue Score: Excellent

Comments:
Great job! And good work trying to make sure testing instructions works for everyone

PR #574, opened 2025-04-28
[Frontend] 'Trending Games' Carousel

bluto0:
Creating a "Trending Games" section on the homepage for [Issue #567](https://github.com/uchicago-cs/chigame/issues/567).

PR #757, opened 2025-05-06
[frontend] Fix carousel alignment issues

### LINKED ISSUE #567: Code Game Carousel [Homepage]
Creating a "Trending Games" section on the homepage

clairezzhang:
After merging #567 into dev, I noticed that there were alignment issues with how the carousel fit into the homepage. I fixed these alignment issues and made a new css file (static/carousel.css) for the carousel code so that it would be easier to find than in its previous location (static/machina/custom-styling.css).

image

Fixes: #567 and #574


pjacobsen1 (approved):
The site ran smoothly, and the carousel looks as expected, so seems all set to me!

src/static/css/homepage.css
```diff
@@ -104,7 +104,7 @@
   background-color: #F9F9F9;
   border: 1px solid #D9D9D9;
   border-radius: 5px;
-  padding: 0.5rem;
+  padding: 0.5rem 0.5rem 0rem 0.5rem;
```
jacqsmei: this is a nitpick, but i don't think this change is needed

jacqsmei (changes requested):
the rest looks good to me! nice catch

src/static/css/homepage.css
clairezzhang: I took a look and I think you're right! I think I changed it when I was trying to figure out why the alignment was off, but it wasn't the reason.

jacqsmei (approved):
nice work!

jacqsmei:
Issue Score: Excellent

Comments:
Nice work catching this bug! Make sure everyone's commits follow the developer guideline standards!

PR #766, opened 2025-05-07
[Inbox + Notifications] Creating notification labelling system

### LINKED ISSUE #599: Create system for labelling notifications
There is currently no system to label notifications. These notifications could relate to spam, creating labels for a specific person, etc. The plan is to look at common notification labels for other sites or a potential library, and implement a system where users can label notifications. 

When starting to work on this issue, we decided to not use an existing library but instead we want to create some general labels so that users can more easily categorize their notifications.

hossainmd-uc:
This PR references issue #599 .

Currently, there is no notification labelling/filtering system. This enhancement adds a notification label to each notification in the inbox.

To create this, a category field was added to `Notification` model. Then a corresponding view was created to support frontend actions. Finally, changes were made to the template file `user_inbox.html` to allow users to categorize notifications.


src/templates/users/user_inbox.html
```diff
@@ -52,6 +54,60 @@
...
+
+    .dropdown-submenu li {
+      padding: 8px;
+      cursor: pointer;
+    }
+
+    .dropdown-submenu li:hover {
+      background-color: #f0f0f0;
+    }
+  </style>
```
jonathanbacon22, line 116: great job also taking care of the frontend!

src/chigame/users/views.py
```diff
@@ -634,3 +636,18 @@ def bulk_inbox(request):
                 notification = Notification.objects.get(pk=pk)
                 notification.mark_as_read()
     return redirect(reverse("users:user-inbox", kwargs={"pk": request.user.pk}))
+
+
+@csrf_protect
+@require_POST
+def move_notification(request, pk):
```
jonathanbacon22, line 643: Great!

src/chigame/users/models.py
```diff
@@ -241,6 +247,15 @@ class Notification(models.Model):
     Handles visibility, read/unread status, and timestamping of events.
     """
 
+    CATEGORY_CHOICES = [
+        ("inbox", "Inbox"),
```
jonathanbacon22, line 251: Nice choices for the categories

jonathanbacon22 (approved):
Changes look great to me! I think I recall Chris mentioning we could also group notifications based on who they are from, like grouping by friends or groups. Perhaps this is something we could also talk about adding. For now, I think this is a great labelling system. Nice work!

src/chigame/users/models.py
cmarshall282: At any point do we let users make their own categories?

cmarshall282 (changes requested):
The actual mechanism for moving a notification seems a little unintuitive to me. Could we also add some sort of move button? Generally, we should make it easy for a user to figure out how to use our site.

But otherwise great job! Can we set up a way to filter our inbox by label next?

src/chigame/users/models.py
hossainmd-uc: We are planning to have that be a different custom labelling system for which we will open another issue.

cmarshall282 (approved):
Looks great! Thanks for adding that button

cmarshall282:
Issue Score: Excellent

Comments:
Great job on this! Here is some feedback:

PR #767, opened 2025-05-07
[Knowledge Base] Sanitize incoming markdown content

### LINKED ISSUE #641: Secure markdown rendering against code injection
Given that we allow arbitrary users to generate interfaces that are then rendered to moderators (and eventually other users), we want to ensure they are not able to include malicious scripts. Therefore, we want to look into extensions/configurations that prevent exploiting of `iframes`, `scripts`, and `links`. Although we eventually aim to support external links in some conditions (#639), we will initially be blocking all external references.

Luca-W-D:
By default, the markdown Python library assumes that all content included in a markdown file is developer-generated; therefore, it is safe to render to users. It also explicitly preserves HTML tags embedded in the content to render. This causes a security vulnerability, as contributors can inject code to run when moderators review their content (or, in the worst case, to users visiting the malicious published guide).

This PR makes the following changes:
 * `src/sandbox/markdown-demo/md_app/markdown_extensions.py`: Add two new processors to replace links and scripts with clearly-visible plaintext warnings.
 * `src/sandbox/markdown-demo/md_app/static/css/index.css`: Add new styling so that moderators can quickly identify the problematic sections when reviewing guides
 * `src/sandbox/markdown-demo/md_app/templates/base.html` and `...markdown_content`: Repairing some changes that were lost in a previous commit
 * `src/sandbox/markdown-demo/md_app/views.py`: adding examples of attempted code injection

To test, checkout this branch and navigate to `src/sandbox/markdown-demo`. Then, assuming you already enabled the venv, run `python3 manage.py runserver`. You should see the rendered markdown with clearly-delineated warnings of injected code and hard-coded links. Specifically, it should look like:

image


src/sandbox/markdown-demo/md_app/views.py
```diff
@@ -1,127 +1,46 @@
...
+More content and <a href="...">link</a>
+"""
 
 
 def markdown_content_view(request):
-    # 1) What section to show/highlight
+    # Get the requested section from query parameters
     requested_section = request.GET.get("section", "introduction")
+    # Check if minimal UI is requested
+    is_minimal = request.GET.get("minimal", "false").lower() == "true"
```
yizou275: seems that this belongs to the white-label pr?

src/sandbox/markdown-demo/md_app/views.py
```diff
@@ -1,127 +1,46 @@
...
-        },
-    )
+    # Create context with the rendered content
+    context = {
+        "html_content": html_content,
+        "requested_section": requested_section,
+    }
+
+    # Choose template based on UI mode
+    template = "md_app/minimal_content.html" if is_minimal else "md_app/markdown_content.html"
```
yizou275: similarly, the `minimal_content.html` doesn't seem to belong to this branch?

yizou275 (changes requested):
Thanks Luca and David for this great work! This code generally looks good to me. I test it as you suggested and they all work as expected. I only have two minor points/questions:
- As I commented, there are some codes that belong to the white-label feature, and seem to involve in a `minimal_content.html`, a non-existing file? Maybe does this belong to the white-label PR?
- Question: It seems that all links are currently replaced by text "Hard-Coded Link", and the original links are lost (rather than saving a record somewhere). Is this an intentional decision?

Everything else looks good to me! Thank you so much!

Luca-W-D:
Yep, you're totally right! There was a commit along the way that got added to the wrong branch; I thought I had caught all the changes. I'll remove the minimal-content changes.

In response to your other question – "...Is this an intentional decision?" – yes! I envisioned that moderators reviewing markdown submissions would see side-by-side (1) the raw markdown content and (2) the previewed content. The bold visuals are intended to attract their attention so that they can examine what the markdown file contains. This hopefully should also prevent Unicode attacks (with small/invisible chars)

Luca-W-D:
Removed the relevant sections, checked that it ran locally, and the `diff` looks more accurate now. Now, `&minimal=true` has no effect. Thanks for the feedback!

yizou275 (approved):
Tested it again and everything looks good to me now! Thank you so much!

Luca-W-D:
Great– thanks!

jacqsmei (approved):
cool idea on making it bold to stand out! 
nice work

jacqsmei:
Issue Score: Excellent

Comments:
Great work on this yall! Awesome that y'all thought of security concerns before it could become a problem

Issue #704, opened 2025-05-05
Add leaderboards settings


bacBenChaplin:
Working on leaderboard settings. These are about privacy settings where people can opt out of the globally viewed leaderboards. Their stats are hidden from the rest of the world.

bacBenChaplin:
Brainstorming ways to break this task into different chunks:
1. Change models/ERD to accommodate for leaderboard privacy settings
- - involves updating wiki with visual and text - writing the django models and a PR
2. Create View(s), urls, and templates for interacting with settings/changing settings- - could be split into more tasks
3. integrate privacy settings with existing leaderboard views 
4. Personal stats page for users who have opted out
- - This would be a new view for indivudal leaderboards that track all stats from a single user. includes urls and templates
- - could be multiple tasks
5. Testing

bacBenChaplin:
Idea: I think it makes sense to make some tests for the previous issue #765 as another task.

bacBenChaplin:
Update
So far I have completed the following tasks regarding this issue:
- Design and implement models for leaderboard privacy settings
- Write tests for the model
- implement views, urls, and basic templates to interact with the settings. 
- - views include a setting base page to see all settings for the logged in user
- - a way to modify existing settings / set new ones. basically a management view
- - a way to delete a setting

To do:
- make nicer templates
- make sure that the settings are properly interacting with leaderboard views. i.e., if i set that i am anonymous, then i am actually anonymous on the leaderboards
- investigate how to make sure you cannot be both anonymous and opted out at the same time
- general view tests

bacBenChaplin:
Closing this as the 4 sub issues have been completed, marking the completion of the implementation of the leaderboard privacy settings.

The 4 completed sub issues:
1. Design and implement models to accommodate leaderboard privacy settings.
2. testing of the model implementation.
3. building of the views, forms, templates to interact with the privacy settings.
4. making sure that one cannot be both opted out and anonymous.

Future potential work:
1. testing of the views
2. putting links to the settings on different parts of the site

cmarshall282:
Issue Score: Excellent

Comments:
Great job!

PR #773, opened 2025-05-07
[Leaderboards] Add privacy settings model

### LINKED ISSUE #765: Change models/ERD to accommodate for leaderboard privacy settings
As part of issue #704 I need to change the current ERD to allow for leaderboard privacy settings, and I have to change the current django models as well

bacBenChaplin:
In order to address issue #765 as part of issue #704. We are creating the model that will track the privacy setting. There is a hierarchy to these settings:

1. Specific leaderboard Setting (highest priority)
2. Game-level setting
3. Global setting (lowest priority)

In addition to the model, there is a method that will return the user's setting given a user and optionally a game and leaderboard.


karunkundrum (approved):
This looks good! Maybe try to fix the linter tests by running the pre-commit but I think it looks good.

src/chigame/leaderboards/models.py
```diff
@@ -72,3 +74,64 @@ class MetricScore(models.Model):
...
+    """
+    Stores user Settings for leaderboard visibility.
+
+    Setting hierarchy:
+    1. Specific leaderboard Setting (highest priority)
+    2. Game-level setting
+    3. Global setting (lowest priority)
+    """
+
+    complete_opt_out = models.BooleanField(default=False)
```
cmarshall282, line 65: Would this be better represented as a choice? It doesn't make sense for anonymous to even be an option if we are opting out.

cmarshall282 (approved):
Looks good. I have one question, but in the end it's your decision on what to do with it

cmarshall282:
Issue Score: Excellent

Comments:
Great job on this PR!

PR #777, opened 2025-05-08
[Leaderboards] Updating Django Models and Addressing Comments

### LINKED ISSUE #776: Updating Django Models and Addressing Comments
As part of PR #696 the models designed in issue #510 were implemented, but there were a few things that needed to be changed. These have been noted in the comments of PR #696, but generally, there were a few models that have already been created by other teams, so they need to be properly referenced in this file and the models themselves.

bacBenChaplin:
In our models, we were created models that already existed. We are now correcting it to reference the already created models. Also fixed the references to certain attributes of the models. This is to get proper functionality.


bacBenChaplin:
Just a quick thing to note: in deleting the User model that we had, I believe that we lost the relationship with the region. I think this should be addressed when we add filtering by region, or deal with regions at all. The goal of this PR was to get the functionality of the models correct. To address this, I think we would have to either make a User model that inherits from the UserProfile model or make a change to the UserProfile model directly.

karunkundrum (approved):
This looks good. I think to address the region issue we can look into editing the User model or extending it in following sprints.

cmarshall282 (approved):
Thanks for getting to this!

cmarshall282:
Issue Score: Satisfactory

Comments:

Issue #508, opened 2025-04-22
Prototype Knowledge Base Frontend


Juliamarin2023:
We would like a prototype of what the knowledge base will look like in order to help guide our future progress/tasks. This prototype will be completed in Figma, and we will rely on the user stories created in the spring #1 as a guidline.

Juliamarin2023:
Started working on a very rough and rudimentary version of the wireframe based on the user stories provided. I did my best to implement all the features mentioned in the user stories. I attached an image of this "outline" of the wireframe looks like. Work on the actual prototype will be done on Figma at this [link](https://www.figma.com/design/A36Yb2ijf1hfco1kcOKa8a/Chigame-Knowldege-Base?node-id=1-87&t=gS0lRw9aYVp6P7ZY-1). 

![Image](https://github.com/user-attachments/assets/bcee3831-7a95-4c66-8e5f-a7bebaed7370)

Juliamarin2023:
Initial prototype of the "viewer" perspective of frontend is complete

<img width="835" alt="Image" src="https://github.com/user-attachments/assets/363d7cb2-6507-4914-a30c-9567d394a9c9" />

yizou275:
Added my (partial) alternative ver for viewer to the figma link during weekend.

mrollinswaterman:
Removed myself as an assignee since I didn't do any work on this issue.

yizou275:
Finished the contributor side, regarding the "upload guide" and "manage guide" page.

Juliamarin2023:
Given the number of pages, I think the wireframe has enough content to efficiently guide our work for the next sprint therefore, I am closing this issue.

jacqsmei:
Issue Score: Excellent

Comments:
Nice job! It would be nice to have updated wireframe imgs attached in the issue in the closing statement to people can have a final view of the work done.

PR #778, opened 2025-05-08
[knowledge-base] Add Viewer Landing Page Backend

### LINKED ISSUE #747: Implement knowledge base landing page for normal users
As described by our user stories, a normal user interacts with the knowledge base through one of two pages. First, when they land on the knowledge base, there is a list of game guides visible to them. This should be true for both authenticated and unauthenticated users.

This task consists of two sub-tasks:
1. Creating the HTML template file and any associated static assets (CSS, etc.)
2. Creating the backend interactions (mainly the `view()` method) so that the frontend pulls data from ChiGame

yizou275:
This PR implements the backend of knowledge base's landing page. Note that while there should be links to other pages on the landing page, for now I keep `href = '#'` for all links, and let people who implement other pages add their url later.

## Description
The implemented feature of the page includes the following (all the following links are empty now):
1. Display a list of published guides on the page
2. Include a link to "Manage My Guide" page for all users that are logged in and have previously uploaded any guide
3. Include a link for "Upload a Guide" page; Hide ContributorView link on the top of the page (I just comment the related lines out in html)
4. Show "ModeratorView" link on the top of the page, only when a user is logged in and is a moderator (`moderator=true`)
5. Support searching, including searching by title like "Guide for Chess" (there is no `title` field in `Guide` object, the title display here is always in the format "Guide for [game name]", and by content
6. Support filtering by category, though single-choice only
7. Support sorting
(Note that 6-7 only works after explicitly hit the "Search" button on the page. Frontend could work on JS to make the page reload automatically when the user select a value.)

## File Changes
Specifically, the files I change are:
- `views.py`: modify DefaultView to include displaying Guides that are published, searching, filtering and sorting
- `landing.html`: modify according to `views.py` to give an unstylished view of the landing page
- `src/static/css/knowledge-base-styles.css`: a placeholder box to contain each guide in basic grid view (should be removed after frontend work is done)
- `base.html`: let only the moderators see the "moderator view" link; comment out the "contributor view" link (I actually don't think we need this link in the base html)
- `urls.py`: modify the path of DefaultView accordingly due to generic view

## Testing
To test it, first pull this branch, load the fixture according to `knowledge-base/fixtures/README.md`, and then runserver.
1. First as an unlogged-in user. Go to knowledge-base page. Verify that it displays (1) the list of published guides, (2) "upload a guide" link, (3) no "moderatorview" link, (4) no "manage my guide" link, (5) search, filter, and sorting boxes.
Under the current fixture, it should be:
Screenshot 2025-05-08 at 6 11 39 PM

2. Then log in as user Alex (a moderator). Notice if both "Moderator View" and "Manage My Guide" links show up. The "Moderator View" link actually works (the only working link on this page). (could also test for a non-moderator user Sam, or a user that has never uploaded a guide (forgot to include in fixture, can use a random email to create one yourself)
Screenshot 2025-05-08 at 6 16 49 PM

4. Test the search box. Search "Go Game", "Guide for Chess", and any random content within the guide (can access Guide content by logging into admin interface via `admin@example.com`).
5. Test sorting. Choose one entry from sorting dropdown, click search, and see whether this changes the page.
Screenshot 2025-05-08 at 6 18 50 PM

6. Test filtering. This one is slightly complicated because the fixture forgets to include "Category" for Games. Go to admin interface to add the three Games to different categories (they already have Category example objects), and then test the page.

> AI Reference: consulted ChatGPT on how to do search/query/sort in Django generic Listview. Understood and then modified its example code to accommodate our need.


src/templates/knowledge-base/base.html
```diff
@@ -16,18 +16,15 @@
...
-        {% else %}
-          <a href="{% url 'knowledge-base-contributor' %}">
-            <p class="mb-0">Contributor View</p>
+        {% if request.user.moderator %}
+          <a href="{% url 'knowledge-base-moderator' %}">
+            <p class="mb-0">Moderator View</p>
           </a>
         {% endif %}
+        <!--
+        <a href="{% url 'knowledge-base-contributor' %}">
```
Luca-W-D, line 25: I think that our navigation buttons should stay in the header so that they're consistent across pages. I think that your change of adding distinct "manage guides" and "upload a guide" options makes sense, but could we maybe have them here & replace this comment? Open to your thoughts!

src/templates/knowledge-base/landing.html
```diff
@@ -1,5 +1,43 @@
...
+    <select name="category">
+      <option value="">All Categories</option>
+      {% for cat in categories %}
+        <option value="{{ cat }}"
+                {% if request.GET.category == cat %}selected{% endif %}>{{ cat }}</option>
+      {% endfor %}
+    </select>
+    <select name="sort">
+      <option value="">Sort by: Most Recent Update</option>
+      <option value="az" {% if request.GET.sort == "az" %}selected{% endif %}>Game Name A-Z</option>
```
Luca-W-D, line 19: I think this will be a good spot for people working on the frontend to jump in. I think this could probably be cleaned up with JS. (Not an issue here, just flagging)

Luca-W-D (changes requested):
Wow! This has way more features than I was expecting for a first pass; having both sorting and filtering in the first pass is impressive.

I followed your testing steps:
1.  As a logged-out user, the 5 checks you listed functioned as expected
2. When logged in as Alex\*, the checks worked as you describe. I notice, however, that the `Note: if you have contributed...` still renders even when logged in. Is this intentional?
3. The search worked for the two games and (correctly) showed no results when a game didn't exist. Amazing!
4. I couldn't quite get the filtering to work, though: to recreate, (1) go to the admin site, (2) give Go the abstract strategy category, and (3) refresh the page. There is now a new filter option "abstract strategy;" (4) select this option. Then after you hit "search", the page displays "no guides found" and resets the filter.

Also, when performing step 2, I noticed one issue with the fixture. While the fixture works as expected, it looks like we didn't add email confirmations for the static users. This'll be a hassle for future debugging. So I'll open up an issue to add this verification & categories to the games.

Overall, really cool! If you can easily fix filtering, then I think it's worth pushing through with this PR. However, if it ends up being a headache, I'd also be fine with you deferring it for later implementation.

src/templates/knowledge-base/base.html
yizou275: Yes great point! I think I initially did this because I'm unsure how frontend prototype would look like (previous [figma link](https://www.figma.com/design/A36Yb2ijf1hfco1kcOKa8a/Chigame-Knowldege-Base?node-id=0-1&p=f&t=5lpDMtaTaK0Jp4PQ-0)) in issue #508. Seems that the previous prototype doesn't include a header, though I also agree that including a header can be helpful, to make them able moderate/contribute on each page. I'm personally fine with both ways. Your thoughts on this?

Luca-W-D: Fair– this does seem to fall under a frontend focus. We can boot it to that issue!

Luca-W-D:
Yi and I debugged offline synchronously in response to point 4 from above. This behavior was downstream of the fact that the "Go" game object had no published game guide, but it had an "Abstract Strategy" category. Therefore, when I was filtering by "Abstract Strategy", the frontend correctly displayed that there were no _guides_ (even though there was a _game_) of the same category. The result is that the frontend looks slightly finicky, since it also resets the "category" once this request resolves. While the page appears to glitch, this seems to be explainable behavior.

yizou275:
> Wow! This has way more features than I was expecting for a first pass; having both sorting and filtering in the first pass is impressive.
> 
> I followed your testing steps:
> 
> 1. As a logged-out user, the 5 checks you listed functioned as expected
> 2. When logged in as Alex*, the checks worked as you describe. I notice, however, that the `Note: if you have contributed...` still renders even when logged in. Is this intentional?
> 3. The search worked for the two games and (correctly) showed no results when a game didn't exist. Amazing!
> 4. I couldn't quite get the filtering to work, though: to recreate, (1) go to the admin site, (2) give Go the abstract strategy category, and (3) refresh the page. There is now a new filter option "abstract strategy;" (4) select this option. Then after you hit "search", the page displays "no guides found" and resets the filter.
> 
> Also, when performing step 2, I noticed one issue with the fixture. While the fixture works as expected, it looks like we didn't add email confirmations for the static users. This'll be a hassle for future debugging. So I'll open up an issue to add this verification & categories to the games.
> 
> Overall, really cool! If you can easily fix filtering, then I think it's worth pushing through with this PR. However, if it ends up being a headache, I'd also be fine with you deferring it for later implementation.

Thanks Luca for your help!
- Great catch on 2! I fix this and hide the line for the logged in users.
- Regarding 4, I fix the filter's visually reset bug. I still keep the filter entries to contain categories of games with no guides for now (a helpful edge case for testing whether the filter is working), but we could change it if necessary in future.
- Agree that the choice of header buttons seems to be a frontend focus. I'll just leave the code as is for now. We could bring this up to the next team discussion.

Luca-W-D (approved):
Pulled the updated changes, and everything still works as expected. I can also see the changes you implemented in response to my feedback. Looks great– thank you!

yizou275:
Great! Thank you so much for your helpful feedback as always!

jacqsmei (approved):
nice work! lots of cool features you included for this

jacqsmei:
Issue Score: Excellent

Comments:
Very thoughtful work on this one! And great job with the collaboration to get it to work

PR #779, opened 2025-05-09
[knowledge-base] Add Backend Moderator Pending Guide List Page

### LINKED ISSUE #749: Implement Knowledge Base Moderator's Submission Overview Page
As described by our user stories, moderators are provided with two special views that allow them to manage the knowledge base game guides. The first is the most general: they are provided with a view of all pending game guides. They can then select a submission from this list to review. This task concerns the high-level overview.

There are two sub-tasks:
1. Creating the HTML template file and any associated static assets (CSS, etc.)
2. Creating the backend interactions (mainly the `view()` function) so that the frontend pulls data from ChiGame

yizou275:
This PR implements the backend of knowledge base moderator's main page that displays a list of pending guides, waiting to be reviewed.

### Description
The implemented feature of the page includes the following:
1. Page Access: If user NOT logged in, redirect to log in page; if user is NOT moderator, show 403 forbidden page. 
2. Display a list of pending guides on the page
3. Include a link to "Review" page for each pending guide (empty link now)
4. Can be sorted by Recentest/Newest Upload time

### File Changes
Specifically, the files I change are:
- `views.py`: delete the original ModeratorView (the name is unclear) and add ModeratorGuidesPending view to include displaying pending guides and sorting
- `moderator_pending_guides.html`: modify according to `views.py` to give an unstylished view of the moderator pending-guide-list page
- `moderator.html`: deleted due to unclear name
- `urls.py`: modify the path of DefaultView accordingly due to new generic view name

### Testing
To test it, first pull this branch, load the fixture according to `knowledge-base/fixtures/README.md`, and then runserver.
1. First check user access. Go to the page as unlogged-in user, check whether they ask you to log in. If so, log in first as Sam (`moderator=false`), check whether there is a 403 page. Then log in as Alex (`moderator=true`), this time the page should display.
2. Then check whether the pending lists are displayed. In this fixture, it should be Alex's "Guide for Go Game" and Alice's "Guide for Game A". Check whether it contains the title, the author, the upload date and an empty link "review". (now the author automatically shows the email field, which I think is because in our current system, email is the only mandatory field rather than name)
3. Test sorting. Choose one entry from sorting dropdown, click search, and see whether this changes the page.

Ideally the page (in default sorting) should look like this:
Screenshot 2025-05-08 at 9 25 28 PM


davidmakoyo:
The backend logic and access control are solid, the page works as described. I'm not sure if there is supposed to be a confirm email page for already existing accounts (or if that is possibly a task for a future issue), but after i follow the confirm email link in the terminal and confirm, i see what is intended for each user. The sorting feature is also a very nice touch!

davidmakoyo (approved):

yizou275:
Thank you David! Yes the email confirmation thing is a bug of the fixture -- the data is already there, but you just need to do the verification when you log in. Luca seems to open an issue about this yesterday #782

jacqsmei (approved):
looking good! email thing was a nice catch

jacqsmei:
Issue Score: Excellent

Comments:
Nice job! Really thorough and instructive description as usual

Issue #594, opened 2025-04-28
Adding reactions to live chat messages


nathanL534:
Add the ability to add reactions to messages. This depends on #658 to be completed and (potentially) be merged first.

pmdepaulo:
Pushed new model to branch `live-chat/message-reaction-models`

pmdepaulo:
Created the model LiveChatMessageReaction that includes the user and the message that it maps to, including the content of the actual reaction. This should be a **single emoji**, which is enforced using a validator in a new file validators.py. This will be handled (and the error will be caught) once further implementation of the messages is finished.

pmdepaulo:
Finished, demo in PR #962 ! should get merged once checks pass

abargher:
Issue Score: Excellent

Comments:
Great work!

PR #783, opened 2025-05-09
[Live Chat] Adding message reaction model to support icon-based reactions

pmdepaulo:
Added a LiveChatMessageReaction that maps to a user + message, and is represented by an icon (CharField)


pmdepaulo:
WIP for Issue #594

PR #784, opened 2025-05-09
[Live Chat] Created models to implement reactions to messages.

pmdepaulo:
Created a LiveChatMessageReaction model that maps users and livechat models to each reaction, along with an icon to represent it.

PR #788, opened 2025-05-09
[Live Chat] Added validator for reaction to validate (single emoji)

pmdepaulo:
Added `validators.py` that checks to see if a passed in attribute to a model (in this case, `content`) is a single emoji character. This will throw a ValueError that will later be caught and sent as a message to the User.
Implementation into `models.py` has not been done yet in order to avoid merge conflicts. However, in order to complete the validation process, `validators=[validate_emoji]` must be added as an argument to the `CharField` of the `content` attribute.


quincyleung:
Hi! I’m only seeing the validator code changes included in PR #784 — this PR (#788) doesn’t appear to introduce any new changes on top of that. I think this might have been created from the same branch?

quincyleung (approved):
This looks great—code is clean and validation error message is helpful. Approved!

abargher (approved):
Looks good! I like this idea, good use of exceptions.

abargher:
Issue Score: Excellent

Comments:
Great work! Excellent title and description as well.

Issue #813, opened 2025-05-11
Create achievements factories for testing


bensim0305:
Setting up the achievements factories for future testing with any functions. This will also help the API teams and frontend teams in creating tests.
PR: #814

bensim0305:
Created factories.

abargher:
Issue Score: Excellent

Comments:
Nice work!

PR #814, opened 2025-05-11
[achievements] Created achievement model factories for testing

bensim0305:
created fixtures for the achievements models to setup testing for any functions we create or API's that need testing from the API's team or any other team. Addresses issue #813 .


mhenry111 (commented):
It looks mostly right to me, but I think that the UserAchievement progress attribute needs to be somehow based on the threshold attribute of the corresponding Achievement. Maybe using factory.LazyAttribute functionality could be useful here?

bensim0305:
From my understanding, I thought the achievement progress was gonna be a fixed range (0-1 or 0-100), and basedo n the achievement conditions the progress would be mapped accordingly. For example, if an achievement is "Collect 7 coins", then if the user has 2 out of 7, we can just do quick arithmetic to scale whatever 2/7 is to on a 0-100 scale, so in this example it would be 29/100 progress for that achievement.

mhenry111:
ok. i didn't realize that we can actually overwrite or set factory attribute values in the tests, so i guess random default values are fine. though for the 7 coins example, i thought Achievement threshold would be set to 7 and then the game would report that a user earned 2 coins, and so we would update UserAchievement progress to 2, and then officially award the achievement once progress >= threshold. if that makes any sense. i don't think it really relates to this issue though

mhenry111 (approved):
looks great! might need to be updated later if we add to the model though

abargher (changes requested):
Please resolve the merge conflict-- hopefully shouldn't be too tricky.

abargher (approved):
Looks good!

abargher:
Issue Score: Excellent

Comments:
Great work, and nice discussion!

Issue #734, opened 2025-05-05
Build imported twine file storage


zaynacheema:
With the twine importation feature implemented, we want to see if we can send an imported twine game to the storage of games in Chigame. This will allow for imported games to be accessible on the site and then hopefully, searchable on the page.

Goals:
- Store the twine imports in the game base

Open Question:
- How do I make the imported twine games are "searchable" in the site?

zaynacheema:
I looked for where games may be stored and went into src/chigame/games/models.py and added the code to twine_file to the Game class to allow for twine importations, now a twine file can be uploaded


I was able to go onto the server, go to admin and the models and add a game where the option to upload a file was available for twine:

<img width="906" alt="Image" src="https://github.com/user-attachments/assets/0aece5ec-c1c8-4c0e-8d76-f40cd2e1ebad" />

<img width="953" alt="Image" src="https://github.com/user-attachments/assets/68cc6952-6537-49a3-aa00-3553ac397fcf" />

This actually has made the game searchable by adding it to the central gamebase:

<img width="1305" alt="Image" src="https://github.com/user-attachments/assets/33ca8e36-371b-4c27-bc1c-702e009b47d7" />

zaynacheema:
I will modify the template though so it looks like the one we created in the last issue rather than the normal game_detail template, which I will do by adding a line to Game_detail view to redirect to the IF_view we built if a twine file is uploaded, and modifying the IF view to use the game_detial than a create object:

<img width="984" alt="Image" src="https://github.com/user-attachments/assets/7ffe0d86-4f8c-4a96-96d8-b8c7b56145a3" />

which after testing by opening the server ^ looks to work as expected

I will open a PR for the modifications

zaynacheema:
PR link: #815

zaynacheema:
With goal met to figure out how to store the twine imports and make them searchable from the main page, and with the Pull Request closed, I will now close the issue.

I worked solely on this issue.

Future integration would be to make the page more readable, adding instructions for creators to upload Twine games that anyone can play after opening the link, as well as making the upload connected to the creation page we will make.

jacqsmei:
Issue Score: Excellent

Comments:
Nice job with this! And thanks for the continual updates!

PR #815, opened 2025-05-11
[Interactive fiction] Twine storage importation in game base

zaynacheema:
The twine importation feature now includes being in the central game base. This is allows for twine games to be searchable and addable by admin to models, and display a link to play the game in a separate browser. (Future integration will include letting creators do this when uploading their game, but this is the base code/logic). This is related to issue #734 .

The changes are also highlighted in the issue but here they are:
- I looked for where games may be stored and went into src/chigame/games/models.py and added the code to twine_file to the Game class to allow for twine importations, now a twine file can be uploaded
- I will modify the IF view template in views.py so it looks like the one we created in the last issue rather than the normal game_detail template, which I will do by adding a line to Game_detail view to redirect to the IF_view we built if a twine file is uploaded, and modifying the IF view to use the game_detail than a create object

Also this is what the page should look like after adding the game through admin view:
442532580-7ffe0d86-4f8c-4a96-96d8-b8c7b56145a3

Overall this code is to figure out how to make games searchable and will help us see how to connect uploaded games (if we change the importation method) in the future


zaynacheema:
(The last few commits before PR mention were to pass the checks)

src/chigame/games/urls.py
```diff
@@ -25,16 +25,16 @@
...
     path("create/", views.GameCreateView.as_view(), name="game-create"),
     path("<int:pk>/edit", views.GameEditView.as_view(), name="game-edit"),
-    path("<int:pk>/", views.GameDetailView.as_view(), name="game-detail"),
     path("bgg_search_by_name/", views.bgg_search_by_name, name="bgg_search_by_name"),
     path("search/", views.search_results, name="game-search-results"),
     path("<int:pk>/reviews/", views.ReviewListView.as_view(), name="game-review-list"),
     # interactive fiction
+    path("interactive-fiction/", views.InteractiveFictionView.as_view(), name="interactive-fiction"),
+    path("<int:pk>/upload/", UploadFileView.as_view(), name="upload-file"),
+    path("if-game/<int:pk>/", InteractiveFictionView.as_view(), name="interactive-fiction-detail"),
```
gkharrell01, line 37: 35 and 37 seem to be similar? Very minor adjustment that could be made later to combine IF and interactive-fiction branches.

src/chigame/games/views.py
```diff
@@ -401,38 +408,37 @@ class InteractiveFictionView(TemplateView):
...
-    def post(self, request, pk):
+    def post(self, request, pk=None):
         uploaded_file = request.FILES.get("uploaded_file")
 
         if uploaded_file:
-            upload_path = os.path.join(settings.MEDIA_ROOT, "interactive_uploads")
-            os.makedirs(upload_path, exist_ok=True)
-
-            fs = FileSystemStorage(location=upload_path)
+            # Save the file to twine_games/
```
gkharrell01, line 428: Smart to create a new folder in database!

src/chigame/games/views.py
```diff
@@ -401,38 +408,37 @@ class InteractiveFictionView(TemplateView):
...
-            fs.save(safe_filename, uploaded_file)
-
-            request.session["uploaded_interactive_file"] = f"interactive_uploads/{safe_filename}"
+            filename = fs.save(safe_filename, uploaded_file)
+
+            # Create a basic Game instance
+            game = Game.objects.create(
+                name=uploaded_file.name.replace(".html", ""),
+                description="Uploaded Twine game",
+                min_players=1,
```
gkharrell01, line 437: This is something I was thinking about with the min/max players... Could this be something maybe adjusted where these are assigned explicitly as 1 for all IF games?

src/chigame/games/urls.py
```diff
@@ -44,4 +44,5 @@
     path("tournaments/archived/", views.TournamentArchivedListView.as_view(), name="tournament-archived"),
     # placeholder game
     path("lobby/<int:pk>/coinflip", views.coin_flip_game, name="placeholder-game"),
+    path("<int:pk>/", views.GameDetailView.as_view(), name="game-detail"),
```
gkharrell01, line 47: Should this be using the GameDetailView or one that is specifically for IF games? Not sure if those commits were pushed at this point, so maybe this could be altered so that all IF games use same view.

gkharrell01 (approved):
Looks great! Aside from some url inconsistencies, these are changes seem to be clean and readable for someone not super familiar with backend twine integration. Review approved for merge.

src/1
jacqsmei, line 1: what is this file for?

jacqsmei (commented):
i agree w grace's comments! if there's a way to set player number strictly to 1 that could be neat

jacqsmei (approved):
nice work! will be cool to see this in the demo

jacqsmei:
Issue Score: Satisfactory

Comments:
Nice job! Make sure your commits match the developer guideline standard! Good notes on review

PR #820, opened 2025-05-11
[Interactive-Fiction] Adding Non-Functional IF Create Template

### LINKED ISSUE #614: Create new unstyled template for uploading Interactive Fiction (IF) games
**Objective**
Create a new Django template that allows users to view and interact with an Interactive Fiction (IF) game within Chigame. The template should match our mock-up design and include the appropriate fields and textboxes to display game content, receive user input (e.g., text commands or choices), and provide placeholders for dynamic elements. This task is limited to markup and template structure only—do not add styling or backend functionality.

**Potential Directions**
- Base the structure of the new template closely on the existing game_detail.html, aligning with the mock-up layout.
- Include all required fields and textboxes using Django template syntax, but do not apply any styling or implement functionality.

gkharrell01:
We added a HTML template, as well as a IFGameForm in forms.py and InteractiveFictionGame in models.py to extend on Game model to hopefully help full integration down the line. This template is not fully functional yet, but incorporated elements from precious PR requests into one template. This is referencing issue  #614 , and the template can be viewed at this URL: http://127.0.0.1:8000/games/interactive-fiction/

Below is how it should be displayed. 
[Interactive Fiction Game set up.pdf](https://github.com/user-attachments/files/20151699/Interactive.Fiction.Game.set.up.pdf)

Next steps: adding a way to get to this page from Original Game Creation page as well as incorporate the Categories scroll bar from the Original page as well.


zaynacheema (approved):

zaynacheema (approved):
IFGameCreateView looks good, appreciate the comments -- it makes sense to me.

the urls look good too

The InteractiveFictionForm looks good too! I think those fields make sense for an IF game

src/chigame/games/models.py
zaynacheema, line 1: Perhaps for the models, instead of removing thr set black=True, could do models.PositiveIntegerField(default=1, editable=False)
Since we only have 1 max and min player in IF games, but this is a good start

src/chigame/games/migrations/0029_interactivefictiongame.py
zaynacheema, line 1: Looks good! These changes make sense -> aren't necessary for an IF game.

src/templates/games/interactive-fiction/IF_game_create.html
zaynacheema, line 1: I think a possible issue with this code comes at:
Screenshot 2025-05-11 at 6 44 12 PM
(screenshot because the comment kept removing my text)

as here it is a form tag inside another which I think can cause bugs in HTML, so it might be good to have the inner form outside of the main one

docs/developer/guidelines.md
```diff
@@ -20,20 +20,6 @@ These checks include the following:
   for common style issues in Django templates.
 * Miscellaneous checks for trailing whitespace, missing newlines, etc.
 
-These checks are enforced via GitHub Actions any time you push code to the
```
jacqsmei: I don't think you meant to delete this as part of this PR?

jacqsmei (changes requested):
i agree w zayna's comments!

docs/developer/guidelines.md
gkharrell01: Nope! I just added it back! Thanks for catching this!

alisk8:
Noticed a couple of missing fields (Author name, allow comments and enable ratings checkboxes, genre, and content warnings), which I added, and one unnecessary field (Categories) on the page, which I have removed.

jacqsmei (approved):
looks good now
sorry for the merge conflict hell

gkharrell01:
PR merged and Issue has been closed! See issue #614 . HTML added that matches the mockup given for IF create page. Both Ali and Grace contributed to this PR.

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! Make sure all commits match the developer guideline standards! For future reference the expected behavior in testing instructions would be better as a screenshot rather than a pdf

PR #821, opened 2025-05-11
Live chat/implement models and files needed for profanity filtering

### LINKED ISSUE #800: Implement models and files needed for profanity filtering
To assist in #604, we need to create the models and add the files needed for profanity filtering, as well as determining how we are to implement this:

- We will need to check for users if they want to toggle this on (`profanity` filter in users)
- Also we need to determine if we will store profanity locally or on DB (probably locally for speed's sake)

I will determine these and add them, and figure out what to do with profanity.

aflekkas:
This addresses #800. In this PR, I added the basic features needed for profanity filtering.

image


aflekkas:
@quincyleung @pmdepaulo finally passing all tests!

src/chigame/chat/utils.py
```diff
@@ -0,0 +1,51 @@
+"""
+Utilities for the chat app. Includes profanity filtering.
+"""
+
+import re
+
+
+class ProfanityFilter:
```
quincyleung, line 8: Nice job structuring it - very easy to follow!

quincyleung (approved):
Great work setting up the profanity filter! I tested this locally and it seems to be working well. Just a thought and I know you resolved it, but I saw Copilot's suggestion of censoring only the profane words instead of the entire message and was wondering what you thought of it? I think it could look a bit better if implemented, but this could also be something for a future issue.

aflekkas:
> Great work setting up the profanity filter! I tested this locally and it seems to be working well. Just a thought and I know you resolved it, but I saw Copilot's suggestion of censoring only the profane words instead of the entire message and was wondering what you thought of it? I think it could look a bit better if implemented, but this could also be something for a future issue.

Thank you for your feedback! I think perhaps this could be forwarded to Paulina's issue, if it's needed.

Nevertheless, Ty a lot for your review!

aflekkas:
Will fix the CI tests then ping

aflekkas:
@abargher just fixed the migration merge conflicts & general merge conflicts, this should be able to be merged now!

src/chigame/chat/consumers.py
```diff
@@ -111,14 +116,20 @@ async def receive(self, text_data):
         message = text_data_json["message"]
         user_id = text_data_json["user_id"]
 
-        # Save message to database once when first received from client
-        username = await self.save_message(self.chat_id, user_id, message)
+        # this will need to be made conditional at some point
+        filtered_message = self.profanity_filter.censor_message(message)
```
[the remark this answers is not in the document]
aflekkas: @pmdepaulo check this out I think - you can implement this in your toggle feature

src/chigame/chat/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+
+    # a list of profane words
+    # this list should be updated with new profane words as they are discovered
+    profanity_list = [
+        "uchicago sucks",
+        "uchicago is a bad school",
+        "uchicago is a bad university",
+        "uchicago is a bad college",
+    ]
```
abargher, line 20: This is good for now in this PR, but perhaps this could be read from a server-side text file? That would make it easier to add more filtered words in the future.

abargher (approved):
Looks good overall! The tests are failing again due to a migration issue (I believe), so fix that up and you should be set to merge.

src/chigame/chat/utils.py
aflekkas: That's what I plan to add later! I've opened another issue/PR for this

aflekkas:
@abargher fixed the issues!

abargher (approved):
Great work!

abargher:
Issue Score: Excellent

Comments:
Great job!

PR #852, opened 2025-05-12
commiting base leaderboard files

### LINKED ISSUE #846: Create Basic Leaderboard View

karunkundrum:
This is the base functionality for views on leaderboards. This is created with issue #846.


karunkundrum:
In this PR I have added the base leaderboard functionality to views.py. In order for this to work, I also added the base URL links in urls.py in both leaderboards and the repository base(in configs). I added the leaderboards to the settings/base.py too. I also added a placeholder leaderboard.html template inside of a new leaderboards folder in the templates part of the repository. Finally I added a blank __init__.py to our leaderboards folder. 

Note that empty.html is referenced in the code but has not been implemented. I will create another task to flush out the HTML files correctly. This code also references region, which is an unresolved problem in our code that will be discussed and resolved in the next sprint. 

Again, this code is base code that we will build upon with specifics.

src/chigame/leaderboards/urls.py
```diff
@@ -0,0 +1,7 @@
+from django.urls import path
+
+from .views import leaderboard_view
+
+urlpatterns = [
+    path("leaderboard/<int:game_id>/", leaderboard_view, name="leaderboard_view"),
```
bacBenChaplin: This makes it such that we access game specific leaderboard info with leaderboards/leaderboard/<int:game_id>/ which is confusing. 

Can this be changed to `path("<int:game_id>/", leaderboard_view, name="leaderboard_view")`? This would make it so we can access game leaderboard info with leaderboards/<int:game_id>/.

This would be another issue/PR but it would be nice to have an index page where we see all of the games. I'd be happy to do this.

src/chigame/leaderboards/views.py
```diff
@@ -0,0 +1,22 @@
...
+def leaderboard_view(request, game_id):
+    game = get_object_or_404(Game, id=game_id)
+    leaderboard = game.leaderboards.first()
+
+    if not leaderboard:
+        return render(request, "leaderboards/empty.html", {"game": game})
+
+    entries = (
+        LeaderboardEntry.objects.filter(leaderboard=leaderboard)
+        .select_related("user", "user__region")
```
bacBenChaplin: I think we should comment out or get rid of references to region until it's properly implemented

src/templates/leaderboards/leaderboard.html
```diff
@@ -0,0 +1,20 @@
...
+      <th>Player</th>
+      <th>Region</th>
+    </tr>
+  </thead>
+  <tbody>
+    {% for entry in entries %}
+      <tr>
+        <td>{{ entry.rank }}</td>
+        <td>{{ entry.user.display_name }}</td>
+        <td>{{ entry.user.region }}</td>
```
bacBenChaplin: I think we should comment out or get rid of references to region until it's properly implemented

src/templates/leaderboards/leaderboard.html
```diff
@@ -0,0 +1,20 @@
+<h1>{{ leaderboard.name }}</h1>
+<p>{{ leaderboard.description }}</p>
+<table>
+  <thead>
+    <tr>
+      <th>Rank</th>
+      <th>Player</th>
+      <th>Region</th>
```
bacBenChaplin: I think we should comment out or get rid of references to region until it's properly implemented

bacBenChaplin (changes requested):
I think it mostly looks good, but there are a few comments that I think should be addressed:

- change the url pattern to `path("<int:game_id>/", leaderboard_view, name="leaderboard_view")` so we access it with leaderboards/game instead of leaderboards/leaderboard/game
- remove references to region until it's successfully implemented.

bacBenChaplin (approved):
talked about making the changes in a new issue during the meeting. looks good to me

jgb4 (commented):
LGTM, thanks for this. Going forward we should add game score as one of the Leaderboard html attributes but I'll take care of this in another PR

cmarshall282 (approved):
Looks good!

cmarshall282:
Issue Score: Satisfactory

Comments:

jgb4 (commented):
Sorry I forgot to leave a review. LGTM, and code has been working as I build off it. Thanks

Issue #863, opened 2025-05-12
new UserAchievement attribute


mhenry111:
Add a last_updated attribute to UserAchievement model. The main reason for this is to help track streaks, so ideally the datetime is auto set when the user achievement is created, and can be updated as time goes on

mhenry111:
also updated the date_earned attribute to be able to be null, since progress-based UserAchievements typically won't be earned at the same time they are created

ntagtmeier:
Created the appropriate migrations in the branch

ntagtmeier:
Closed with PR #877

abargher:
Issue Score: Excellent

Comments:
Nice work!

PR #877, opened 2025-05-12
[achievements] updated UserAchievement time fields

mhenry111:
to address issue #863 i added a last_updated attribute to UserAchievements to help with streak tracking, and also updated the date_earned field to be able to be null since not all UserAchievements are earned at the same time they are created.


mhenry111:
seems to be failing a lot of newer tests; need to have the models and factories finalized so that future PRs don't break other teams' code. i think i just have to make migrations to pass these tests though

emilyzen (approved):
Looks good to me

abargher (approved):
Nice!

abargher:
Issue Score: Excellent

Comments:
Great work!

PR #905, opened 2025-05-13
Testing for leaderboards privacy settings model

### LINKED ISSUE #809: Testing for leaderboard privacy setting model and method
I think it makes sense to write some tests for the leaderboard privacy setting model and method from issue #765.  I would have to implement some factories for this. I assume that I need one for User, Game, and Leaderboard, and maybe Leaderboard_entry. The first 2 might already be implemented elsewhere; I would have to implement the last 2 and write tests for my model and method

bacBenChaplin:
This is the testing for the leaderboards privacy setting model. I had to create some factories and 6 tests regarding the method about getting the user settings. I also had to configure our leaderboards app. 

NOTE: I am using the leaderboard privacy setting model in this, and it has not yet been pushed to dev. I have decided to not include the updated models in this PR because that would create unnecessary conflict. The point is that in order to run the tests, you need the updated models which are not yet pushed to dev.


bacBenChaplin:
CI / run-tests is failing because i am referring to models that don't technically exist, but they should soon exist assuming that the models in review go through.

Aufsteyegen (approved):
Looks good. Tests comprehensively cover all privacy‐setting scenarios, factories are well-configured, and the app is properly registered.

cmarshall282 (approved):
Thanks for adding this

cmarshall282:
Issue Score: Excellent

Comments:
Great job!

PR #911, opened 2025-05-13
Live chat/enable public chat rooms and add a list of global chats users can participate in

### LINKED ISSUE #826: Enable public chat rooms and add a list of global chats users can participate in
Since the chat component allows anyone to go to a chat using `/route/{id}` - it would be interesting to see a directory of chats that the user can browse. Perhaps, we can even allow users to create live chats among users and browse them here. For now, we will just create a directory of live chats and improve it in the future.

aflekkas:
This PR addresses #826. I have added the models needed as well as the UI elements to support it.


quincyleung (commented):
This is a solid feature addition—great job! Tested it locally and the `/chat/live-chat-list/` view works as expected and makes global chats discoverable. The chat links are clickable and route correctly.

I do have some minor comments. Do we want to only show public chats in the list view or  allow private chats to show and users can request to join? Do we also want to add a is_authenticated check to prevent unauthenticated users from joining chats?

```
if request.user.is_authenticated and chat.public and not chat.users.filter(id=request.user.id).exists():
    chat.users.add(request.user)
```

Nice work pushing this through!

src/chigame/chat/views.py
```diff
@@ -7,4 +7,13 @@ def chat(request, chat_id):
     chat = get_object_or_404(LiveChat, id=chat_id)
     messages = LiveChatMessage.objects.filter(live_chat=chat).order_by("sent_at")
 
+    # if the chat is public, add the request user to the chat
+    if chat.public and not chat.users.filter(id=request.user.id).exists():
```
quincyleung: Could add an authentication check if request.user.is_authenticated and chat.public and not chat.users.filter(id=request.user.id).exists():

aflekkas:
> This is a solid feature addition—great job! Tested it locally and the `/chat/live-chat-list/` view works as expected and makes global chats discoverable. The chat links are clickable and route correctly.
> 
> 
> 
> I do have some minor comments. Do we want to only show public chats in the list view or  allow private chats to show and users can request to join? Do we also want to add a is_authenticated check to prevent unauthenticated users from joining chats?
> 
> 
> 
> ```
> 
> if request.user.is_authenticated and chat.public and not chat.users.filter(id=request.user.id).exists():
> 
>     chat.users.add(request.user)
> 
> ```
> 
> 
> 
> Nice work pushing this through!

I think private chats should not show up otherwise we'd show like chats between two users. Would be unusual, no? When it comes to the other line, I believe chats are already protected by with but maybe we add it.

src/chigame/chat/views.py
```diff
@@ -7,4 +7,13 @@ def chat(request, chat_id):
...
+        chat.users.add(request.user)
+
     return render(request, "chat/index.html", {"chat": chat, "messages": messages})
+
+
+def live_chat_list(request):
```
quincyleung, line 21: should we change this to `chats = LiveChat.objects.filter(public=True)` so private chats don't show?

aflekkas: Yes! Do you want to join on this PR and make the change?

quincyleung: Sure let me join and make the changes rn!

quincyleung:
Assigned @nathanL534 to review since I made slight changes to the code!

nathanL534 (approved):
Tested it locally, everything seems to be working. Looks good guys!

abargher (changes requested):
Looks good overall, but there is an odd issue with the DB causing test failures. Take a look, ping me if you have any trouble finding a solution, and I can look closer.

abargher (approved):
All set now!

abargher:
Issue Score: Excellent

Comments:
Great work!

PR #914, opened 2025-05-13
[Notifications] Implement notification pop-up interface based on prototype

### LINKED ISSUE #895: Implement the notifications interface based on prototype
We have initiated talks with frontend about redesigning / creating a notifications interface. Currently, notifications live in the inbox and it would be helpful to have some real time alerts for users.

While Frontend team sent an initial prototype, it did not have all the functions that we would ideally want in notifications, including timestamp, color-coding based on label, searching, and performing bulk actions such as deleting.

The goal of this issue is to finalize all the functionalities of this pop-up interface and implement a basic, working version of it.

hossainmd-uc:
- PR closes #895 

This pull request implements a notification interface based on the prototype provided by the frontend team. The goal was to redesign the notification system to be more interactive, styled, and visible across all pages, rather than being confined to the Inbox view.

🔧 Features Implemented
- Added a context processor to globally expose the current user's undismissed notifications.
- Created a Bootstrap-styled dropdown menu in the navbar to display live notifications.
- Added a red badge displaying the current notification count.
- Built a dismissNotification() function using fetch() to support deletion via AJAX, with instant DOM updates.
- Reused backend logic for the deleted field to avoid duplicating deletion behavior.
- Applied custom Bootstrap classes and hover styles for better user experience.

📸 Preview
Screenshots of the interface, including badge and dropdown behavior, were shared in the original issue: #895


src/static/js/notifications.js
```diff
@@ -0,0 +1,29 @@
...
+  const badge = document.querySelector('#notification-btn .badge');
+  const remaining = document.querySelectorAll('.notifications-dropdown .dropdown-item').length;
+  if (badge) {
+    if (remaining === 0) {
+      badge.remove();
+    } else {
+      badge.innerText = remaining;
+    }
+  }
+}
```
davdma, line 29: Awesome! Adding a style sheet for notifications will help standardize the design in the future. We could consider having one for the profile/accounts page too.

davdma (approved):
The notification button on the top right looks good. I think the next step would be making sure clicking on the notification redirects via the handler since right now clicking on the notification dropdown does not do anything (I can help with that).

src/static/js/notifications.js
```diff
@@ -0,0 +1,29 @@
+function dismissNotification(notificationId) {
+  fetch(`/users/act_on_inbox_notification/${notificationId}/delete`, {
```
cmarshall282, line 2: Is this something worth talking to the API team about?

cmarshall282 (approved):
Looks good. I'm wondering if we want to go through the api team for exposing an endpoint like this

src/static/js/notifications.js
hossainmd-uc: Hmm, I think one of us created this url and the associated view to support inbox functionality; this merely references that view. I'm thinking it should be fine. Also, I'm not sure what you mean by exposing an endpoint

cmarshall282: By exposing it as an endpoint, I basically mean provide an API view for it. Usually, when we want to call things like this from the backend, we want an API for it. We can stick with this for now, but it is something we might want to keep in mind.

cmarshall282:
Issue Score: Excellent

Comments:
Great job on this PR!

Issue #604, opened 2025-04-28
Add conditional rendering of profanity filters


aflekkas:
Add the ability for profanity filtering to be toggled on and off and conditional rendering of it. Most services do not _force_ you to implement this.

As it stands with issue #800 - profanity filtering is currently mandatory for all users. The flag has been added, but logic needs to be implemented in `consumers.py` or on the frontend to facilitate the toggling of profanity filtering.

aflekkas:
@pmdepaulo hi! If it has not already been merged, I would recommend working off of #800 when you begin work on this task

nathanL534:
https://github.com/user-attachments/assets/58ea67e2-ea1a-4e24-b27d-2b13456dfcb7

added a button to toggle profanity

abargher:
Issue Score: Excellent

Comments:
Nice work!
Here are some ways this could be improved:
- Even though you can add labels to issues to categorize them, you should avoid titles that could plausibly apply to multiple teams.

PR #918, opened 2025-05-13
Live chat/improve documentation potentially where wordsphrases are stored for profanity filtering features

### LINKED ISSUE #858: Improve documentation & potentially where words/phrases are stored for profanity filtering features
Documentation for the current profanity filtering issue #800. Right now, documentation for this important feature is lacking, so this issue will aim to address that.

Right now, words and phrases for profanity filtering are stored in a constant, another improvement that could be made would be to store them in a text file or something that could be edited more clearly without diving into code. I will also explore that in this issue.

aflekkas:
This addresses #858 - I add a few comments, but the main part of this was separating profane words to be from a file rather than a variable. They are now a lot more manageable, and can be edited more easily.


aflekkas:
In the future, if there are wishes to make profanity filtering common across the app - this feature could provide the framework for or also integrate with a global list of profane words.

src/chigame/chat/consumers.py
```diff
@@ -111,14 +116,20 @@ async def receive(self, text_data):
         message = text_data_json["message"]
         user_id = text_data_json["user_id"]
 
-        # Save message to database once when first received from client
-        username = await self.save_message(self.chat_id, user_id, message)
+        # this will need to be made conditional at some point
+        filtered_message = self.profanity_filter.censor_message(message)
```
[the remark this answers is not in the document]
aflekkas: @pmdepaulo will address this in her issue #604

src/chigame/chat/PROFANITY.txt
```diff
@@ -0,0 +1 @@
+uchicago sucks,uchicago is a bad school,uchicago is a bad university,uchicago is a bad college
```
pmdepaulo, line 1: this is hilarious
for the future, do you plan on manually inputting profanity or maybe finding an existing database (so that we don't have to write a bunch of it in a file haha) 
it also doesn't have to be in the scope of this quarter, since the functionality is there but something to think about if you wanted to move forward!

src/chigame/chat/utils.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        Censor the message from profane words.
+
+        Args:
+            message (str): The message to censor.
+
+        Returns:
+            str: The censored message.
+        """
+        return self.pattern.sub(lambda m: "*" * len(m.group()), message)
```
pmdepaulo: i like this censorship approach

pmdepaulo (approved):
This looks good, left a comment on considering the scope of using a profanity file, but I'm also more than happy to explore that on my own as well. Thank you for this foundation!

src/chigame/chat/PROFANITY.txt
aflekkas: Oh, that's a good idea - might be something we add as an issue in the backlog - do you want to create that?

src/chigame/chat/utils.py
aflekkas: Ty!

src/chigame/chat/utils.py
```diff
@@ -2,29 +2,41 @@
...
-        escaped_words = [re.escape(word) for word in self.profanity_list]  # create the regex patterns
-        self.pattern = re.compile(r"\b(" + "|".join(escaped_words) + r")\b", re.IGNORECASE)
+        # Get the directory where this file is located
+        profanity_file = PATH_TO_PROFANITY_FILE
+
+        # Read profane words from file (comma separated)
+        try:
+            with open(profanity_file) as f:
+                profanity_text = f.read().strip()
+                self.profanity_list = [word.strip() for word in profanity_text.split(",")]
```
abargher, line 28: I think a newline delimited list is more appropriate for something like this, since it's easier to read and edit. Plus, what if a censored phrase contains a comma? This is fine for now, but something worth thinking about.

abargher (approved):
Nice work! Good change for extensibility.

abargher:
Issue Score: Excellent

Comments:
Nice work!

PR #931, opened 2025-05-13
[Adding Games - Checkers] Settings menu redesign

### LINKED ISSUE #919: Improve appearance of settings tab
The settings tab currently overlaps jarringly with the navbar, and also does not have a transition animation at all. The accessibility button is also the default appearance, rather than the buttons present on the rest of the page.

Coco555-cyber:
This PR redesigns the settings menu to be more inline with the rest of the frontend. The settings button and menu were reworked into a slide-out menu at the bottom. Hovering over the settings tab will cause the menu you slide up, revealing the setting selectors. The "Toggle Colorblind Mode" button has also been redesigned to differentiate whether or not the setting is active. Moving the cursor off of the menu or tab should cause the menu to slide back into place, leaving only the tab exposed.

This PR also slightly adjusted the css of the rest of the frontend.


jacobwolf-1 (approved):
Nice work, tested the hover and color blind mode button changes and they work. I like this redesign, LGTM. 
settings-post-hover
settings-pre-hover

abargher (approved):
Nice work! Looks good to make the UI more consistent.

abargher:
Issue Score: Excellent

Comments:
Great work!

PR #932, opened 2025-05-13
[knowledge-base] Normal User Landing Page

davidmakoyo:
### Summary
This PR updates the frontend for the knowledge base landing page. The page now uses Bootstrap for layout and styling. This closes issue #747

---

### File Changes

**`src/templates/knowledge-base/landing.html`**
- Refactored the template to use Bootstrap classes for a clean, responsive layout.  
- Improved the search/filter/sort bar for better usability.  
- Displayed guides in a responsive grid of cards, each showing the game, category, and last updated time.  
- Enhanced empty state and login prompt for clarity.  
- Kept all existing features: search, filter, sort, upload, and manage links.  
- Removed the broken guide detail link to prevent errors.

---

### How to Test

1. **Run the development server:**
   ```bash
   python manage.py runserver
   ```
2. **Navigate to the knowledge base home page** (http://127.0.0.1:8000/knowledge-base/).
3. **Test the following:**
   - Make sure it looks like this 
image.
4. **Check for errors:**
   - There should be no broken links or template errors.


yizou275:
Thanks David for this work! Followed your instructions for testing, and pages and links all function as intended. (have a link break for Admin, which is due to a separate issue and doesn't suggest problem with this PR -- I'll note it in a later comment).

A few minor thoughts or questions:
1. Rather than having "Guide List" as title, I'm thinking maybe change it to "Knowledge Base", to make the page represent the frontpage of the knowledge base feature? (or any other better titles are also welcome)
2. Could we change "moderator view" and "contributor view" to something else? For instance, "contributor view" seems confusing to me -- maybe "manage my guides" would be a better name? "moderator view" can be something like "moderator dashboard"?
3. Do we need to add "upload new guide" to the header as well? If so, will it appear repetitive to make it appear twice on the landing page?
4. Could you add the link to guide detail page, for "Guide for [game name]" grids? At least add an empty link for now?
5. Could we include category as looking similar to tags (like your tags on "accepted"/"rejected")? For those guides with multiple categories, while it's okay to include only the first category if the space is limited, could you add some indicator that there are more categories than this? For instance, tag "Abstract Strategy" and text "and more"?

These are just minor issues, so feel free to ignore them or save those for a future issue. Let me know what you think, and I'll approve it afterwards :)

yizou275:
Note: When I log into Admin and test the "manage my guides" page, it shows a "NoReverseMatch" error. I think this is due to the contributor page itself, and will open a separate issue that debugs and fixes this.

davidmakoyo:
I think all of your suggested changes are good ideas. However, I do think that adding "Upload New Guide" to the header would be repetitive because there is already a "create new guide" button in the "manage my guides" tab, so I implemented them all except for that one. Thank you!

yizou275 (approved):
This looks good to me now! Thanks:)

jacqsmei (approved):
nice job! the page looks really good!

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! Excited to see this at the demo

PR #935, opened 2025-05-13
Adding Dictionary API for Word Validation

Zhejiej:
Implemented Dictionary API validation:
Dictionary API - https://dictionaryapi.dev/

I created the isValidWord function and sent an asynchronous request to the Dictionary API to check if the provided word exists in the dictionary.

The word is passed to the API endpoint, and a fetch request is made. If the API returns a 404 response (meaning the word is not found), it would return False. If the word exists, the API response is parsed as JSON, and the function returns True.

This function is used in the handleSubmitWord function, where the current word entered by the user is checked for validity before proceeding with game logic.

The await keyword ensures the game logic waits for the API call to finish before continuing, providing a smoother flow for word validation.

### DIFF
--- a/src/sandbox/words-game/README.md
+++ b/src/sandbox/words-game/README.md
@@ -1,3 +1,6 @@
 # Words-Game
 Word list based on the 5-letter word list compiled by shmookey.
 https://gist.github.com/shmookey/b28e342e1b1756c4700f42f17102c2ff
+
+Dictionary API for known words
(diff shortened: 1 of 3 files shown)


Zhejiej:
Closed Pull Request. Going back to just merging to dev since then there will only be 1PR instead of 1PR to merge into solo then another PR to merge into dev.

Issue #933, opened 2025-05-13
Create html template, url, and view for demo game


emilyzen:
Create an html template for a fake game for the purpose of demonstrating achievement features. For this issue, the actual demo game won't be implemented yet, and the html template will essentially be blank. I also plan to add this page to urls.py and views.py.

emilyzen:
I've made PR #943 to address this issue. In that PR, I added a demo game HTML template and the url and view. Once that PR is merged, this issue will be resolved.

emilyzen:
PR #943 was merged, so I'm closing this issue.

abargher:
Issue Score: Excellent

Comments:
Nice work!
Here are some ways this could be improved:
- Even though you can add labels to issues to categorize them, you should avoid titles that could plausibly apply to multiple teams.

PR #943, opened 2025-05-13
[achievements] added demo game template, url, and view

emilyzen:
I created a template for the demo game and also added a view and url. This PR addresses Issue #933. I'm experiencing issues when I try to run the server to test these files, so I would appreciate if other people could test my changes.


ntagtmeier (changes requested):
I think you need to add something to the src/config/urls.py file for the URLs to be accessible on the chigame site.

emilyzen:
I added the achievements url to src/config/urls.py.

src/chigame/achievements/urls.py
ntagtmeier, line 1: This is a little nitpicky, but could you change the url to "demo-game" instead of "achievements/demo-game"? There's already an "achievements" in the path because of how the config is set up

ntagtmeier (changes requested):

src/chigame/achievements/urls.py
emilyzen: I changed the url to "demo-game"

ntagtmeier (approved):
Checked the URL and it looks like exactly what we want. Great job!

abargher (approved):
Looks good! This will be very useful to have.

abargher:
Issue Score: Excellent

Comments:
Great job!

PR #945, opened 2025-05-13
[Adding-Games-II] Adding End Screen and Back Button

Zhejiej:
Adding Hidden End Screen so that when game.mode == True, it would trigger the new `showEndScreen` function, which unhides the End Screen with all the words you've guessed. Also added a back button to go back to the homepage to play other game modes.

### DIFF
--- a/src/sandbox/words-game/game.css
+++ b/src/sandbox/words-game/game.css
@@ -179,11 +179,67 @@ h1 {
 
 #title a{
     font-family: inherit;
+    display: inline-block;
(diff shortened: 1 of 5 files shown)


namdabest253 (approved):
This looks good. Good work.

cmarshall282 (approved):
Looks good! Can we try to remove repeated code between game and solo?

cmarshall282:
Issue Score: Satisfactory

Comments:

Issue #908, opened 2025-05-13
Re-add Download Option for Contributor Manage Guides page


mrollinswaterman:
In PR #900, the download option for Contributors to download their guides from their "manage guides page" was unintentionally removed. This Issue is to re-add that functionality.

mrollinswaterman:
PR approved.

jacqsmei:
Issue Score: Satisfactory

Comments:
Nice work! Make sure to include updates on your work before closing

PR #955, opened 2025-05-14
[knowledge-base] Re-added Download option to Contributor guides page

mrollinswaterman:
## Overview
This PR fixes Issue #908. Users are now able to download guides from their contributor page. In addition to adding downloads back in, I also changed the formatting to look a bit cleaner, and included a 3rd party download icon for a more polished look.

## File Changes
**`src/templates/knowledge-base/base.html`**
- Added code to include the 3rd party stylesheet which provides the new download icon

**`src/templates/knowledge-base/contributor_manage_guide.html`**
- Added code that displays the 3rd party icon and makes it a clickable link.
- Added code that connects the icon link to the "download-guide" url/view

## Testing
2. Log in as admin, or another user who is marked as a contributor.
3. Go to your Contributor View page.
4. If you have a guide uploaded already, you should see a download icon next to the guide's name. Select it to download your guide.
  a. If you have no guides, upload one to see the download button.
5. Check that the download worked correctly. Your screen should look like the screenshot below:

Screenshot 2025-05-13 at 7 53 37 PM


Luca-W-D (approved):
Thanks for the fix Miles! I followed your testing steps, and the functionality works as expected (for anyone new to the branch, make sure you load the fixture before replicating this locally).

Stylistically, I think it's an interesting choice that the download icon is to the right of the guide names, since this means that it's alignment varies row-by-row. I think we should move it to be before the text– that way, it's always aligned against the left border. I know that the UI styling is adjacent to this PR, but could be a good enhancement while we're already changing the page!

jacqsmei (approved):
nice work! ui looks great

jacqsmei:
Issue Score: Excellent

Comments:
Great work! Very useful feature to have

PR #958, opened 2025-05-14
Refactor context menu into component and add placeholder actions for all intended features

nathanL534:
This addresses issue #937.
This PR addresses refactoring the drop-down context menu within _messages.html file into another component in another file and adding all the intended features that can be applied on a message. Alongside that, I also made each corresponding dummy function with each respective feature that we intend to add(delete, pin, reply, edit, react).

PR #962, opened 2025-05-14
[Live Chat] Implement Basic Chat Reactions

pmdepaulo:
Using the models/validators created in #784 and #788, I am implementing the basic reaction functionality using POST requests and interactions with the database.


pmdepaulo:
This PR also branches off of live-chat/creating-componenet-for-drop-down-menu-when-right-click-on-message (PR #958).
Contributes to Issue #594

quincyleung:
I'm trying to test it locally but it doesn't seem to work for me - looked in my log and was getting this error: 
```
Not Found: /chat/message/<int:message_id>/react
HTTP POST /chat/message/%3Cint:message_id%3E/react 404
```
Perhaps there's a mistake with the URL?

src/templates/chat/components/_context_menu.html
```diff
@@ -0,0 +1,86 @@
...
+      }
+    });
+  }
+
+  /**
+   * Reacts to a message on the frontend.
+   */
+  function reactToMessage(content = "❤️") {
+    const messageId = replyToMessageId;
+    fetch('/chat/message/<int:message_id>/react', {
```
quincyleung: i believe the issue might be related to this line - could try '/chat/message/${messageId}/react' instead

pmdepaulo: just fixed, thank you! requesting another review if possible

src/templates/chat/components/_context_menu.html
```diff
@@ -69,7 +69,7 @@
    */
   function reactToMessage(content = "❤️") {
     const messageId = replyToMessageId;
-    fetch('/chat/message/<int:message_id>/react', {
+    fetch(`/chat/message/${messageId}/delete`, {
```
quincyleung: I think you meant to change it to `react` and not `delete`?

pmdepaulo:
Finished basic chat reaction with a <3 emoji! Planning on further implementing with an emoji picker, works to add and remove reactions to the message UI.

https://github.com/user-attachments/assets/25f67cae-a25e-4234-8740-f9705adc9b41

quincyleung (approved):
Tested it locally and it works smoothly, great work! Seems like you have a solid plan for how to move forward as well 💯  I did notice that the add reactions option on the menu has a different background color than the rest, but that's super minor and could be addressed in a future PR!

abargher (approved):
Looks great, this is significant work!

abargher:
Issue Score: Excellent

Comments:
Great job! Nice work documenting the process as well, this is a large change.

Issue #832, opened 2025-05-12
Expand upon group functionalities - allow users to join and leave groups properly


hollyfwei:
Create views for users to join and leave groups according to user stories.

PR #981, opened 2025-05-14
[Users Profiles + Friending] Group Join and Leave Views and Paths Added

hollyfwei:
To address issue #832 832 :

- added GroupLeaveView
- added GroupJoinView
- added urls for both

*leveraged standard permissions w loginrequiredmixin (custom for prev function)


davdma (approved):
LGTM. I'm wondering if we can fill in any of the redirects currently based on what the tournaments and achievements teams have implemented.

harriet-fisher (commented):
I'm not sure how this differs from the previous PR, is that because they were both on the same branch or am I missing something? Also wondering if there's a specific reason for the group views to be in API, is that just to ensure functionality with API users?

jonathanbacon22 (approved):
Looks great to me!

Issue #829, opened 2025-05-12
Expand upon group functionalities - only logged in users can create groups


hollyfwei:
Starting to expand upon group functionalities, will adjust so that only users who have logged in can create a group.

PR #983, opened 2025-05-14
[Users Profiles + Friending] Group Creation Permissions Limited to Authenticated Users

hollyfwei:
To address issue #829:

- added a custom permissions class
- added a create function in the proper view
- adjusted serializer


src/chigame/api/views.py
```diff
@@ -107,10 +110,22 @@ class MessageView(generics.CreateAPIView):
...
+        if request.method in SAFE_METHODS:
+            return True
+        return request.user and request.user.is_authenticated
+
+
 class GroupListView(generics.ListCreateAPIView):
     queryset = Group.objects.all()
     serializer_class = GroupSerializer
     pagination_class = PageNumberPagination
+    permissions_classes = [IsAuthenticatedOrReadOnly]
```
hossainmd-uc: Is this the correct syntax (the name of variable)? Just want to double check since I found this on the django documentation:
image

src/chigame/api/serializers.py
```diff
@@ -84,7 +84,8 @@ def create(self, validated_data):
 class GroupSerializer(serializers.ModelSerializer):
     class Meta:
         model = Group
-        fields = "__all__"
+        fields = ["id", "name", "description"]
```
hossainmd-uc: Good adjustment!

hossainmd-uc (changes requested):
Looks good! There's just a minor variable name that might pose an issue.

davdma (approved):
Looks good Holly! It's great that tournaments redirect are now working.

src/chigame/api/views.py
```diff
@@ -107,10 +110,22 @@ class MessageView(generics.CreateAPIView):
     serializer_class = MessageSerializer
 
 
+class IsAuthenticatedOrReadOnly(BasePermission):
+    def has_permission(self, request, view):
+        if request.method in SAFE_METHODS:
+            return True
+        return request.user and request.user.is_authenticated
```
jonathanbacon22: nice work, I like this convention you have been doing

jonathanbacon22 (approved):
LGTM! I noticed Rain had a question, I wonder if your method is different than the one in the documentation. But if it is something you wrote yourself, I think it gets the job done!

src/chigame/api/views.py
hollyfwei: Thank you for flagging this!! The syntax of the right side of the equality is fine but the left side was off - I fixed the permission_classes bit.

hossainmd-uc (approved):
LGTM! Thanks for making needed changes

Issue #992, opened 2025-05-14
Issue with IF game leading to Create Game Template instead of detail page


zaynacheema:
While testing and working on issue #979  I saw that clicking on a Twine game in the search results now leads to a Create Game Template instead of the detail page as it did before (probably the result of push after that code was implemented)

Goal:
- Fix the code so that clicking on the IF game leads to the detail page instead of the creation template

zaynacheema:
Witht the fix, it no longer goes to "Create IF Template" when I click on an IF game but to:

<img width="972" alt="Image" src="https://github.com/user-attachments/assets/8d38330e-69a1-4155-944d-3991b10d0a4b" />

zaynacheema:
When I click on the game:

<img width="1101" alt="Image" src="https://github.com/user-attachments/assets/5fa42165-4800-4ef2-ba60-89a33a8e2dff" />

This shows up
<img width="702" alt="Image" src="https://github.com/user-attachments/assets/ad5e7894-21d5-41ec-b88d-4de55d5bf354" />

Which shouldn't happen, it should be the game detail view instead.

zaynacheema:
With the fix, it no longer goes to "Create IF Template" when I click on an IF game but to:

<img width="972" alt="Image" src="https://github.com/user-attachments/assets/8d38330e-69a1-4155-944d-3991b10d0a4b" />

(I want to note I made the fix before making the PR, I just thought I would put the images here of before an after to highlight the issue and the fix)

zaynacheema:
With the error (mis-redirection) fixed and the PR successfully merged and closed, I will now close the issue.

I worked solely on this issue.

Future integration would include making sure the Create Game Template is correctly directed from the create page, rather than the game page as it was doing so here.

jacqsmei:
Issue Score: Excellent

Comments:
Nice catch on this bug! and nice clear updates and imgs

PR #995, opened 2025-05-15
[Interactive Fiction] Fixing misredirection

zaynacheema:
Clicking on a Twine game in the search results now leads to a Create Game Template instead of the detail page as it did before (probably the result of push after that code was implemented) 

This PR is related to issue #992 

The code was so that clicking on the IF game leads to the detail page instead of the creation template, primarily editing the game_detail html and the game view code to make sure the redirect doesn't happen.

It should look like this when a game is clicked on:
Image

### DIFF
--- a/src/chigame/games/views.py
+++ b/src/chigame/games/views.py
@@ -78,10 +78,8 @@ class GameDetailView(LoginRequiredMixin, FormMixin, DetailView):
         context["popularity"] = self.object.reviews.count()
         context["avg_rating"] = self.object.reviews.filter(is_public=True).aggregate(Avg("rating"))["rating__avg"]
 
-        #FOR IF/twine GAMES
-        context["is_twine_game"] = (
-            self.object.twine_file.name.endswith(".html") if self.object.twine_file else False
-        )
+        # FOR IF/twine GAMES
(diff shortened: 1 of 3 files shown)


src/chigame/games/views.py
```diff
@@ -79,6 +77,9 @@ def get_context_data(self, **kwargs):
         context["reviews"] = Review.objects.filter(game=self.object)
         context["popularity"] = self.object.reviews.count()
         context["avg_rating"] = self.object.reviews.filter(is_public=True).aggregate(Avg("rating"))["rating__avg"]
+
+        # FOR IF/twine GAMES
+        context["is_twine_game"] = self.object.twine_file.name.endswith(".html") if self.object.twine_file else False
```
gkharrell01, line 84: Looks great to me! Is readable, but this could be expanded if others think that the readability is an issue.

gkharrell01 (approved):
Good work, Zayna! See comment on readability, not an issue but could be expanded on if need be.

jacqsmei (approved):
nice catch zayna!

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! Remember to keep all commit messages matching the standard format!

PR #999, opened 2025-05-15
[Word game] creating django models for word game

namdabest253:
In model.py, created a WordStat model that stores users and their streak and last played. Created WordSession model that keeps track of guesses, board state, completion status, start, and end time. Created WordAchievementsUnlock model that keeps track of what achievements each user has unlocked.


namdabest253:
Made migrations to fix not passing the tests.

Zhejiej (approved):
good job, but remember to comment on your work.

PR #1001, opened 2025-05-15
Live chat/add the models needed to create polls for live chat

### LINKED ISSUE #847: Add the models needed to create polls for live chat
An underrated feature of chats which could be very useful for live chats is the ability to do polls.

Image

I was thinking of something like Discord, as shown above.

To do this we first need to create the models. I think polls should be able to be attached to messages, or messages should be able to have a poll attached to them, if a message with a `poll_id` is detected when loading, we also load in the poll.

We also need models for votes, etc. Once this is complete, we can add the integrated polls.

aflekkas:
This addresses #847. I have added the models needed for poll implementation. @sidkama can you review?


sidkama (approved):
looks good Alex - just curious, why is the Meta class a subclass of LiveChatPollVote? wouldn't it make sense to just have unique_together be an accessible list within LiveChatPollVote instead of its own class? aside from that clarification, this looks good to go pre implementing the actual poll feature in the live chat itself.

nathanL534:
Hi @alexandros-lekkas if you don't mind ill hop on and fix the merge conflicts.

aflekkas:
> Hi @alexandros-lekkas if you don't mind ill hop on and fix the merge conflicts.

Thank you so much big dawg - appreciate it

abargher (changes requested):
More merge conflicts here, please take a look.

nathanL534:
> More merge conflicts here, please take a look.

On it right now! -- it looks like the errors are steming from a previous commit so im going to revert back to a working commit

nathanL534:
> > More merge conflicts here, please take a look.
> 
> On it right now! -- it looks like the errors are steming from a previous commit so im going to revert back to a working commit

All fixed

abargher (approved):
These commits were already merged, looks like they got swept up in another PR while fixing conflicts.

aflekkas:
@abargher does that mean we can close this PR?

PR #916, opened 2025-05-13
[Notifications] Organize notifications by label into sections in Inbox

hossainmd-uc:
- Closes #906

This pull request introduces categorized sections in the Inbox view for notifications, allowing users to navigate between different notification types such as Inbox, Spam, Social, Promotions, Updates, and Archived.

🔧 Changes Implemented
- Updated the notifications frontend to render by section based on user-assigned labels.
- Matched section styling with the overall Bootstrap theme for consistency.
- Fixed a bug where sections would disappear after certain navigation events.
- Ensured that clicking "Inbox" in the navbar redirects correctly to the default Inbox section.

See screenshots attached in [#906](https://github.com/uchicago-cs/chigame/issues/906) for examples of section layout and styling.

PR #1016, opened 2025-05-15
[User Profiles + Friending] Support Custom Notifications

### LINKED ISSUE #961: [User Profiles + Friending] Custom Notification Labelling
We currently have a system that users can use to create notifications. However, this may not cover the entire breadth of labels that users want to create, so our plan is to create a system in which users can create labels to categorize their notifications. We plan to change our current model in order to support this feature. Initially, we thought to create a new system, but we realized that this would clash with the current one.

jonathanbacon22:
- Closes #961 

In order to support custom notification labelling, I made changes to the models, views, and urls for the Users section of the codebase. 

I added a NotificationLabel model and labels field to Notification, and ensured that users can create their own custom notifications. 

I also added views to create labels, assign labels, and filter notifications by label, which was a nice additional touch. 

Then, I added routes for these new views.


davdma (approved):
Good thinking with the labels on the notifications! Low hanging fruit that I think adds a lot more user interaction for the notifications

harriet-fisher (commented):
LGTM! One thing we might consider implementing in the future is adding some pre-determined/default labels for users that are most commonly used (i.e ‘Flag as important’ or ‘Flag as Event’… I’m not sure, we can brainstorm haha) Which might make the labels easier to become acquainted with if users decide to implement more personalized, niche labels later!

hossainmd-uc (approved):
The changes look great, nice work! Just needs frontend support. The bookmarking feature could use frontend support too so let's open an issue for those?

cmarshall282 (commented):
What is the difference between this and the categories? I get that this adds some customizability, but my question is more along the lines of why do we have two different systems for labeling notifications. Can we clarify this?

Also how does this integrate into existing work like Rain's PR #916

hossainmd-uc:
This is more like an optional feature that could potentially enhance the user experience. In popular emails like Gmail and Outlook, there is a custom labelling feature. While it's not the most significant feature, it is nevertheless available for the user. 

If we decide to move forward with this, I think we can find a way to make these features integrate well. Perhaps we can have a context menu and give users both option to create custom label/use pre-defined categories.

harriet-fisher (approved):

cmarshall282 (approved):
Looks good!

cmarshall282:
Issue Score: Excellent

Comments:
Great job!

PR #1032, opened 2025-05-15
[Adding Games - Checkers] Colorblind mode pieces fix

### LINKED ISSUE #970: Add primary light and primary dark variables
Currently the color of the pieces is set directly using the COLORS.red and COLORS.black variables, which causes a bug when using the colorblind setting and then resetting the game. Rather than directly setting the pieces to a specific value, a placeholder variable for the lighter and darker colored pieces could be used, allowing the colorblind setting to still function after a reset.

Coco555-cyber:
This PR fixes the bug of movement not working in colorblind mode by adding lightPiece and darkPiece variables to track the current player, rather than the exact color.


vliao1223 (approved):
Nice job! Changing players from COLORS.red and COLORS.black to a separate lightPiece and darkPiece variable seems to have done the magic. I play tested it, and now I can select the pieces in colorblind mode.

abargher (changes requested):
Please fix the merge conflicts from `dev`-- most of them are relatively straightforward, but there are some larger sections I'm not certain how to resolve myself without knowing which sections we want to keep.

abargher (approved):
Nice work!

abargher:
Issue Score: Excellent

Comments:
Great job!

PR #1039, opened 2025-05-16
Adding the ability to pin messages

### LINKED ISSUE #604: Add conditional rendering of profanity filters

### LINKED ISSUE #608: Adding the ability to pin messages in a chat
Pinning messages involves having 1 message for each chat that is pinned to that. This depends on #720 to be merged. Working off this will mean potentially working off of that branch.

nathanL534:
this acts upon issue #608.  This pr addresses adding the ability to pin messages both on the frontend and backend


src/templates/chat/components/_context_menu.html
```diff
@@ -64,6 +65,75 @@
...
+
+  function editMessage() {
+    hideContextMenu();
+
+  }
+
+
+
+
+  function reactToMessage() {
```
pmdepaulo: reactToMessage already exists below, this might cause issues

pmdepaulo (commented):
Only 1 small thing about repeating reactToMessage() function

abargher (changes requested):
Some relatively complex merge conflicts to address, please take a look.

abargher:
Issue Score: Excellent

Comments:
Nice job!

PR #1042, opened 2025-05-16
[knowledge-base] Add QOL features to fixture and clean plaintext README

Luca-W-D:
### Description
As described by #782, we've noticed two issues with the knowledge base fixture that currently exists in the repo. When first loaded, the included users have unverified emails, requiring testers to manually confirm these addresses before accessing the knowledge base. Additionally, we've been prevented from testing some features (such as filtering games by categories) since the fixture doesn't include this data for the sample games. These category-specific features appear across a few knowledge base pages, so they're important to add for ease of future testing.

### Changes
The PR affects the following files:
* `README.md`: First, I updated the file to reflect the new changes to the fixtures, including email verification and game categories. I also noticed that, while the README renders well as markdown, it was hard to read as plaintext. So I changed the formatting to be accessible when viewed in a text editor. I also made some headers slightly more consistent.
* `kb_email_verifications_data.json`: This includes email verification for all four test users, so we no longer have to manually confirm their addresses
* `kb_user_game_data.json`: I added categories to the three games, including cases where: a game has two categories; two games share a category; and a game has one unique category.

### Testing
In order to test, you first need to simulate someone pulling our codebase for the first time. This ensures that you don't have pre-existing email verification for the test users. *Don't run this command if you have important data stored in your local database! This will erase that data.*

Noting the warning above, run `python3 manage.py flush` to clear your local database.

Then, execute this command (also found in the fixture README):

`python manage.py loaddata src/chigame/knowledge_base/fixtures/kb_user_game_data.json src/chigame/knowledge_base/fixtures/kb_guides_data.json src/chigame/knowledge_base/fixtures/kb_feedbacks_data.json src/chigame/knowledge_base/fixtures/kb_email_verifications_data`

If successful, you should see that 275 objects from 1 fixture and 24 objects from 4 fixtures where installed.

Then, run Chigame and log in as any test user (Sam, Alice, Alex, Admin; each has the password "test12345"). Confirm there is no email verification required. Go to the knowledge base page and confirm that the two Guides indicate that the games are assigned categories (Abstract Strategy).

Issue #902, opened 2025-05-13
Implement Knowledge Base Single Guide Page Backend 1 -- before Markdown Parser


yizou275:
As described by our user stories, a normal user interacts with the knowledge base through one of two pages. The first is the landing page in #747. The second page is the page for a single guide, where the markdown file of a single guide is displayed in a wiki format.
This issue specifically works on the first stage of backend of this page, namely pulling data from the database and display the content of markdown file directly on the page. (The second stage of backend is the integration of markdown parser to convert the markdown into html.)

yizou275:
My backend tasks include the following:
- make urls for published guides (hopefully in the form of "/guide-for-[game name]/", rather than pk for guide)
- modify the current GuideDetailView, if necessary
- make basic html that displays all info needed; will display just raw markdown content
- if have time, work on "like" and "favorite" feature

yizou275:
Regarding the first point above, I decide to use pk in urls for now, as the form of "/guide-for-game-name" requires a game slug. Adding this field to the game object will affect the fixture. The games app don't print out their game name in their urls as well. So I wouldn't bother it for now, and maybe changing guide url for pk to game name slug could be a good idea for a future issue.
That being said, I still handle the scenario of unpublished guides by claiming the guide is unpublished and pointing them back to the knowledge base landing page.

yizou275:
Closes by #925.

jacqsmei:
Issue Score: Excellent

Comments:
Great job breaking down the task and giving updates!

PR #1045, opened 2025-05-16
[Knowledge Base] Integrate markdown parsing with guide view

### LINKED ISSUE #903: Implement Knowledge Base Single Guide Page Backend 2 -- Markdown Parser
As described by our knowledge base user stories, a normal user interacts with the page for a single guide, where the markdown file of a single guide is displayed in a wiki format.
This issue specifically works on the second stage of backend of this page, which integrates our sandbox markdown parser such that the markdown content is converted into html. (The pre-req for this issue is backend stage 1, in #902).

Luca-W-D:
### Description
Building on Yi's work in #902, this branch generates a public-facing UI (including styled HTML tags) from the markdown's plaintext contents. Specifically, we now support the primary user story interaction where a curious player is looking for more information about a game: they navigate to the knowledge base, select a game, and then see a "Single Guide Detail" page. With this PR's changes, they can now view the guide as rich text rather than just plain text.

### Files Changes
* `markdown_extension.py`: This sanitizer was iterated on in the sandbox, and – now that the sandbox feature is being integrated into the main codebase – requires a production version. I duplicated the code directly into the knowledge base subapp, so that we don't have a dependency on a non-production file. I also added an exclusion case for `<pre><code>...</code></pre>` when sanitizing tags: HTML will not run any JS inside these tags, and they are necessary for code block rich text.
* `views.py`: I added the logic to pass the plaintext guide to our parser, and forward the rich text it generates to the template's context for rendering
* `markdown.css`: By default, the markdown parsing library has some rather bland styling for some tags (i.e. blockquote). We will want to add our own additional styling. However, we render markdown scripts on multiple pages (both this page
 and the moderator review page, for example). Moving this styling to a stand-alone CSS file will help with importing.
* `single-guide.css`: There are a few page-specific styles needed just to make the basic template readable. Future styling (for example, for handling like / favorite buttons) can also go here.
* `guide_detail.html`: I moved the previous metadata into a separate header and then add a section to render the rich content as well

### Testing
To test these changes, pull this branch and ensure that you've loaded the knowledge base testing features. If you haven't loaded the fixtures yet, you can find instructions in `src/chigame/knowledge_base/fixtures/README.md`. If you're testing this before #1042 has been merged with dev, note that the codeblock will not render as expected due to a malformed codeblock in the dev's current fixture.

Run ChiGame and navigate to the knowledge base landing page. You do not necessarily need to be logged in. Choose either of the guides, which should take you to a new page that shows the (1) properties of the guide, including its author, edit date, categories, and a link to visit the game and (2) guide contents, displayed using rich text.

For example,
image

Based on the fixture version you're using, the code block might render different and/or there may not be a game category. Both of those bugs are fixed by other PRs.


mrollinswaterman (commented):
Everything looks good to me. No notes!

Luca-W-D:
Thanks, Miles!

jacqsmei (approved):
nice work! great to see this come to life in chigame

mrollinswaterman (approved):
Looks good, just approving because I forgot to before!

jacqsmei:
Issue Score: Excellent

Comments:
Nice job! really important feature to have

Issue #1048, opened 2025-05-16
Optimize landing page to be more user friendly


Luca-W-D:
The knowledge base landing page has been updated by several PRs, which leads to some buttons appearing throughout the page. In order to prepare for our final presentation – and, since most people will interact with our feature through this page – it'd be nice to have it appear slightly more consistent. This includes reorganizing and restyling some components of the page.

Luca-W-D:
I think that the most important changes are 
* Making the search box one, combined component on the page (maybe with the input above and options below)
* Moving the action buttons at the bottom of the page to the top
* The user stories describe multiple logged-out users; therefore, the large warning banner seems slightly loud. Maybe we should have a more subtle way of indicating they are logged out.

Also, I think that the gray knowledge-base header / nav bar (implemented for debugging) will eventually need to be phased out. That requires touching multiple pages, so I'll create a new issue for that!

Luca-W-D:
Closing, as #1050 has been merged.

jacqsmei:
Issue Score: Excellent

Comments:
Nice job! Excited to see the improvements in the demo

PR #1050, opened 2025-05-16
[Knowledge Base] Make knowledge base frontend more user friendly

Luca-W-D:
### Description
As described by #1048, there are some stylistic issues with the landing page that make it hard for first time users to understand its functionality. This PR aims to make the landing page more accessible by: (1) making the search UI more consistent, (2) bringing action buttons to the top of the page, so that users can see them more easily, and (3) making the indication that a user is logged in more subtle.

### Files changed
* `knowledge-base-styles.css`: I removed the banner that warns users they are logged out, and replaced it with a disabled "manage my guides" button that prompts them to log in when hovered over. This style sheet holds the styles for this button
* `base.html`: I found it challenging to read the navbar sometimes, given that it was low-contrast with small text. I made it bold so that it's slightly more accessible
* `landing.html`: As described above, I (1) changed how the searchbar was organized and displayed and (2) moved the buttons from the very end of the page to a new location at the top. I also added a variation of the "manage my guides" page for unauthenticated users.

### Testing
Pull this branch, and first navigate to the knowledge base landing page when not logged in. There should be guides displayed (presuming you've loaded the knowledge base fixture; if not, instructions are in `src/chigame/knowledge_base/fixtures/README.md`). It should appear similar to:
image

First, confirm that you can still search for a guide and sort by some condition as well.  Depending on which fixture you're using, there may not be categories to filter by.

Then, log in as Sam and return to this page. You should see the "manage my guides" button no longer disabled. Confirm this link works.
image

Finally, visit some guide from the landing page to ensure these links also work.


Luca-W-D:
Also I realize this is minor, but, while writing the description, I realized that I wanted to switch the order of the buttons in the top right. You might notice this change when testing.

yizou275 (commented):
Thanks Luca for making landing page prettier! The page matches your screenshot and everything works well!

Just a few question/confusion:
- When I'm unlogged in, the grey "manage my guide" button is disabled but doesn't point to a log-in page. Instead, I click on the "manage my guide" link on the header and this redirects me to the log-in page. Is this intentional?
  - I personally get confused, and think maybe the two buttons should have the same behavior? Though also I'm unsure whether we wish to repeat "manage my guides" for two times on the landing page (open to your thoughts!).
  - I also notice the grey button, though displayed as a disabled version for unlogged in users, does not display for a logged in user that contributed no guides. Is this also intentional?

- This is being nitpicky, but the position of "clear search" looks not fitting the page well? Would it help if put it to the left of "search", or right next to the sorting window? Or maybe make it into a button or something? (I'm also not sure what to change, but feel that the vertical gap between the guide entries and the searching windows is a little too  large due to "clear search" on a separate line?)

Luca-W-D:
Thank you for the response!
1. Yes– both of those behaviors were intentional, though mostly because they maintained the spirit of the previous page. After thinking about this for a bit longer, I think a better course of action is just replacing this with a single "Login to Contribute" button. This should be more intuitive.
2. Yeah, I agree it looks weird. I tried a couple of different things (including being to the left of "search" as a button, but I clicked it a few times by accident). However, I think being to the left of "search" as text is a happy middle ground.

Will implement both– thanks again!

yizou275 (approved):
Yes agree with your ideas on "log in to contribute" and the new "clear search" position. This now looks good! Thank you for your work!

Luca-W-D:
Thank you!

jacqsmei (approved):
nice work! my comments were basically the same as Yi's

jacqsmei:
PR Score: Excellent

Comments:
Great job! Really thorough PR description and good teamwork on the review

PR #1065, opened 2025-05-16
[Interactive-Fiction] Changing Cover Image URL Field to a Cover Image Drag and Drop Area

### LINKED ISSUE #849: Make the Cover Image Field a Drag and Drop Instead of a Textbox for a URL
**Objective**
On the IF Upload Game page, we have a field called cover image, which serves as the cover image for the game the user is uploading. Currently, we prompt the user to enter a URL for the cover image. However, the team decided that it would be more user-friendly if we allowed the user to upload an image instead.

**Directions**
- To achieve our objective, we will turn the URL-prompting text box for the cover image into a drag-and-drop zone, where the user can upload an image they have stored locally.

alisk8:
This PR will include code that allows users to upload images that they have stored locally as their game's cover image. Previously, we prompted the user for a URL, however, we deemed that to be less user-friendly. This PR references this the problem described in issue #849


alisk8:
With the current push, the cover image should look as follows, and mark the end of this PR, barring any issues caught by reviewers.

image

zaynacheema (approved):
it looks like the main change was just adding the cover image, which also has drag and drop. The code for that looks good and is readable. It also shows up as expected when I run the server myself and look at the creation page.

jacqsmei (approved):
nice work! i like how the ui looks

PR #1063, opened 2025-05-16
POST request to update game state

enzeeo:
Sends a POST request everytime a piece is move to update the game state. This is currently saved into the django database. The code in here has everything necessary for the code to run including a copy of the checkers models from PR #674. Additionally, there is also code from PR #909. 

I may take those two PRs down if this one works properly seeing as they use the same code. 

To run this please do the following: 
1. `python manage.py makemigrations` 
2. `python manage.py migrate`
3. `python manage.py loaddata src/chigame/games/fixtures/checkers-fixture.json`
4. `python manage.py runserver`
5. Add `games/checkers/1/` to your current URL
6. Inspect and go to Console
7. Every move should say it was successful 
8. To double check you can do one of the following: 
9. Go to the django admin and see whether the board state under `game_checkersboard` updates its 2D array 
10. Or within Visual Studio Code download SQLite Viewer and view the `game_checkersboard` within the db.sqlite3 file (please close and reopen it after every move you make to see if the array has updated).

PR #1071, opened 2025-05-17
[Checkers] Print Checkers Board according to Django database

### LINKED ISSUE #1140: [Checkers] Get Board State from Database
Use the GET function to retrieve the board state from the database. This means that we can have multiple tabs opened with the same game and it should be updated such that making a move on one tab should be translated / seen once you switch to the other tab.

enzeeo:
The checkers game will now retrieve the board from the database which will allow for support for multiplayer. This means that you can have the same game open on multiple tabs, which will update by itself without needing to hit refresh. 

To test this out please run the following: 
Please keep in mind that this also uses code from PR #1063, but I did not change up any of the code so it should merge without issue. 

1.` python manage.py makemigrations`
2. `python manage.py migrate`
3. `python manage.py loaddata src/chigame/games/fixtures/checkers-fixture.json`
4. `python manage.py runserver`
5. Add `games/checkers/1/` to your current URL
6. Now open another tab with the same URL 
7. Change the state of the game in one tab and look at the other tab which should have the same change 

Please also note that when you change tabs, the turns may not have changed and therefore every tab change will cause you to start as a red player. This will be fixed in later PRs.


brianHarder (approved):
I tested out the code using the provided instructions, and everything works as expected. Moving a piece on one tab results in the correct update happening on the other. The only error I encountered was the tabs not keeping track of whose turn it. It seems like each tab only remembers whose turn it is according to the last move made on that tab, so moving red on tab 1 means tab 1 now thinks it is black's turn whereas tab 2 still thinks it is red's turn. However, this error was noted in the PR.

I looked over the Python code for the views as well as the HTML/JS, and everything makes sense. Apart from the noted error, LGTM!

abargher (changes requested):
Looks like the same linter issue as in #1063. Once it's fixed in one of either that PR or this one, we should be set, since I'll merge the fixed one first.

enzeeo:
PR #1098 which was further along in development has been approved and therefore it is not necessary to merge this one as there will just be unnecessary merge conflicts

PR #1083, opened 2025-05-17
[Word game] Fix Tile Color Coding Logic Error

### LINKED ISSUE #1009: [Word Game] Fix color coding logic error
When I tested the game, I typed in the word "KILLS" and got 🟨🟩🟩🟨🟩. With this logic, this means that there are 2 Ls in the word because 1 is already green and there's another yellow one. And will only 2 yellows and the rest being green, this color coding implies that the correct word would be "KILKS" which was not the correct answer. And the correct answer turned out to be "SILKS".
Plan to fix this error so that the game letter checking accounts for the number of times that letter has appeared in the word already. To address this, we will revise the letter-checking algorithm to ensure that tile colors are assigned accurately based on the number of times each letter appears in the target word. The updated logic will correctly distinguish between exact matches and partial matches, and will prevent overcounting duplicate letters. This task will be complete when the game provides consistent and accurate color feedback for guesses containing repeated letters.

namdabest253:
This PR closes #1009 
In the original code the color logic didn't take into account the number of times a letter appears in the word. For example:
Target word: "SILKS"
User guess: "KILLS"
The tiles were colored 🟨🟩🟩🟨🟩 when it should be 🟨🟩🟩⬛🟩
So it's marking the second L as yellow even though there's only 1 L in SILKS.

New algorithm:
Removed the complex and error-prone getTileColor function
Created a clearer calculateTileColors function that processes the entire word at once
Used a two-pass approach:
First pass handles all exact matches (green)
Second pass handles remaining letters based on what's left in the target word
Tracked letter usage accurately:
Used a targetUsed array to track which positions in the target word have been matched
Ensured each position in the target word is only used once for coloring
Fixed duplicate letter handling:
Now correctly handles words with duplicate letters following official Wordle rules
Properly assigns yellow only when there are sufficient instances of the letter in the target

Screenshot 2025-05-17 181240
In the original code, the first L would have been colored yellow when it should not be.


vtcrenshaw (approved):
Looks good, useful fix and creative algorithm.

cmarshall282 (changes requested):
Can we remove all changes that don't relate to this fix?

cmarshall282 (approved):
Thanks for catching that bug

cmarshall282:
Issue Score: Excellent

Comments:

PR #1098, opened 2025-05-19
[Checkers] Prevent non-logged-in users from accessing Checkers game

### LINKED ISSUE #1142: [Checkers] Ensure only logged in users can access the game
Make sure that only logged in users, and users whose emails are associated with the game can access the checkers game.

enzeeo:
This PR ensures that only logged-in players can access the checkers game and that only ones with the correct email can access it. To test this PR run the following: 

1. Sign into Chigame as any account
2. Go to `games/checkers/1`
3. You should be denied access 
4. Now sign out 
5. Sign into Chigame with email `red@example.com`
6. And reset the password
7. You may also need to confirm the email as well
8. Go to `games/checkers/1`
9. You should be granted access and can play normally


nneidlinger (approved):
As outlined in the PR, I tried visiting games/checkers/1 as not red@example.com and was denied access to see the game. When signed in as red@example.com, I was able to see and play the game. Switching accounts also gave the correct results. I also tried to visit games/checkers/1 without logging in and was sent to a login page, very good!
However, when playing the game, I encountered a minor bug where moving too fast generated an Internal server error that forced me to redo my turn. I think this is really minor, though, and may even be resolved just by implementing multiplayer as we hope to do, as players will be locked from moving until their opponent's turn is over and sent to them, preventing them from moving too fast. I've attached the full log message for future debugging.
Because this bug doesn't seem serious or like it will be very common, I think the PR is still valid. Ship it!
[error.txt](https://github.com/user-attachments/files/20276910/error.txt)

abargher (changes requested):
Nice work! Quick linter issue to clean up, but otherwise looks great.

abargher (approved):
Great!

abargher:
Issue Score: Excellent

Comments:
Great job! Thorough testing description and review.

PR #654, opened 2025-05-01
[knowledge-base] Implementing routing and basic page layouts

Luca-W-D:
As discussed in Issue #653, we plan to start implementing knowledge base components. However, the team plansto create to create these views in parallel (i.e. the content contributor page will be made while the moderator page is developed) as described by our user stories. Thus, we need to ensure that we're working from a common code base.  That way, all of our individual views can communicate with each other. Therefore, this issue:
1. Creates a URL manager for knowledge base pages
1. Creates a frontend for the Knowledge Base app that integrates with the larger interface (routes URLs to our manager, includes our feature in the navbar)
1. Creates template HTML, CSS files

PR #1105, opened 2025-05-19
[Knowledge Base] Implement user-facing subapp navbar

Luca-W-D:
PR #654, which implemented the core pages and routing for the knowledge base subapp, included a debug navbar for testing different pages. As we prepare for our final presentation (and for users to hopefully use the knowledge base!), it's important to implement to a more permanent solution.

This PR moves the debug navbar to a new, stacked layout that integrates with the Chigame navbar. We hope that by extending the Chigame navbar, other features could eventually integrate similar sub-navigation within their own features. Additionally, as a minor UI improvement, I switched the "moderator" and "contributor" links to be standalone buttons, included in the navbar.

## Files Changed
* `knowledge-base-styles.css`: aesthetic styles for the navbar presentation and structural styles for the page as a whole
* `templates/base.html`: adds a new block to the Chigame navbar, allowing nested pages to add their own nav elements
* `templates/knowledge-base/base.html`: adds a navbar specific to our sub-app and removes the old debug version

## Testing
Pull this branch and navigate to the knowledge base (unauthenticated or authenticated are both okay). You should notice that the previous debug header (a standalone gray-background div above the "Knowledge Base" header) no longer appears. Instead, there is a new navigation row below the main Chigame navigation UI.
image

One other change is that the header should now scroll with the page. Open the devtools console, and run the following command.
```
document.querySelector(".knowledge-base.container").style.minHeight = "1000px";
```
This simulates the behavior of when the knowledge base is fully populated, with many guides causing the page to grow downwards. If the command executes successfully, the footer should "disappear" as it moves to the bottom of the now-longer page. As you scroll down, the knowledge base row of the navbar should move with the page

image

Also, assuming you have the knowledge base dummy fixture loaded, open one of the guides. The navbar changes should propagate to this page as well.


mrollinswaterman (approved):
Looks good to me. Well done Luca!

jacqsmei (approved):
great work! stickiness of the bar is nice

jacqsmei:
Issue Score: Excellent

Comments:
Very thorough and helpful PR description!

Issue #1123, opened 2025-05-19
Implement Achievement Pinning Client-Side Functionality


jiayuew1:
To follow up with issue #1096, This issue addresses the need to enhance our achievement system with client-side functionality for pinning/unpinning achievements. Currently, the achievement display uses static HTML, and we need to implement JavaScript functionality to make the pin buttons interactive.

- Allow users to pin/unpin achievements via client-side JavaScript
- Provide visual feedback when achievements are pinned/unpinned
- Dynamically update the UI to move achievements between the "All Games" and "Pinned Achievements" tabs
- Add animations for a smoother user experience

jiayuew1:
Addressed in PR #1124

abargher:
Issue Score: Excellent

Comments:
Great job! Well done breaking down the tasks that need to be done.

Issue #1040, opened 2025-05-16
Discuss with user profile about how to udpate username when changed after logging in


jiayuew1:
After testing with our fixtures, I have noticed that, even if a user has changed the username after logging in, the achievement page will still display the original name in the view. I want to fix this in order to display properly

jiayuew1:
Talked with their team and addressed in PR #1124

abargher:
Issue Score: Excellent

Comments:
Nice job!
Here are some ways this could be improved:
- Status updates on an issue should have something tangible to share. "Discuss" is not always a tangible goal, so it would be better to include a brief update describing what that discussion entailed. Fixing it with a PR is a good update, but generally you want to have a more concrete description of how the discussion was resolved.

PR #1097, opened 2025-05-18
[achievements] Adding Pinning functionality in views.py

jiayuew1:
To address issue #1096, this PR implements the backend functionality for pinning achievements to user profiles. Two Django view functions have been added to the views.py file:

1. toggle_pin_achievement: An AJAX view that allows users to pin or unpin an achievement. It uses get_or_create to handle both initial pinning and toggling of existing achievements.
2. get_pinned_achievements: A view that retrieves all pinned achievements for the current user, including relevant details about each achievement and its associated game.
3. Added URL pattern in achievements/urls.py to expose the endpoint

PR #1124, opened 2025-05-19
[achievements] Pinning Functionality UI and Tiny Bug Fix

jiayuew1:
To address issue #1123,
Split the codebase into separate HTML, CSS, and JS files for better maintainability
In the src/static/js/achievements.js, I have added the following functionalities.
1. Adding animation to the pin icon when clicked (e.g., rotation and color change)
2. Providing immediate visual feedback in the UI when an achievement is pinned/unpinned
3. Visually updating the achievements list in the pinned section without page reload

I've also fixed the issue where the HTML was not displaying the correct username (Issue #1040). It appears that the User Profile team updated the database to store the name instead of the username. I've updated the HTML to display the username accordingly. Additionally, I changed the URL to use the user ID instead of the username, since the username can be nullable and is less reliable.


mhenry111:
I pulled the latest commit into my test branch and ran chigame. The new organization of the code looks good, but none of the new pinning functionality mentioned in your comment worked for me. Am I supposed to be able to pin/unpin achievements?

jiayuew1:
> I pulled the latest commit into my test branch and ran chigame. The new organization of the code looks good, but none of the new pinning functionality mentioned in your comment worked for me. Am I supposed to be able to pin/unpin achievements?

I've added the pinning functionality to views.py in PR #1097. Ben has approved it, so I expect it will be merged soon. Due to time constraints, I went ahead and created the PR—sorry for any extra work this may cause on your end. Once you update the views and URLs, everything should work as expected.

mhenry111:
> 
> I've added the pinning functionality to views.py in PR #1097. Ben has approved it, so I expect it will be merged soon. Due to time constraints, I went ahead and created the PR—sorry for any extra work this may cause on your end. Once you update the views and URLs, everything should work as expected.

Thanks for the clarification! The pinning/unpinning functionality looks great. I only have a couple recommendations, which can be addressed in future issues:

1. update code to not allow pinning for achievements that haven't been earned yet, because currently when they appear on the pinned achievements page, it says 'Unlocked on Not Unlocked' for the date. 
2. adjust achievements cards so that the longer achievement names aren't split in the middle of a word like this:

lexicograph
er

mhenry111 (approved):
nicely organized and works with the pinning functionality of #1097; looks good to me!

jiayuew1:
> > I've added the pinning functionality to views.py in PR #1097. Ben has approved it, so I expect it will be merged soon. Due to time constraints, I went ahead and created the PR—sorry for any extra work this may cause on your end. Once you update the views and URLs, everything should work as expected.
> 
> Thanks for the clarification! The pinning/unpinning functionality looks great. I only have a couple recommendations, which can be addressed in future issues:
> 
> 1. update code to not allow pinning for achievements that haven't been earned yet, because currently when they appear on the pinned achievements page, it says 'Unlocked on Not Unlocked' for the date.
> 2. adjust achievements cards so that the longer achievement names aren't split in the middle of a word like this:
> 
> lexicograph er

I've fixed the display error where "Unlocked on not unlocked" was shown—it now correctly displays "Not yet unlocked." I don't think we need to add a constraint on the achievement status, as users may want to pin achievements they haven't earned yet as goals in the pinned achievements panel. As for the second point, I'll convert it into an issue and address it in the future.

mhenry111:
> 
> I've fixed the display error where "Unlocked on not unlocked" was shown—it now correctly displays "Not yet unlocked." I don't think we need to add a constraint on the achievement status, as users may want to pin achievements they haven't earned yet as goals in the pinned achievements panel. 

Oh I hadn't thought about it that way. That sounds good to me!

abargher (approved):
Looks great! Nice to have things split up into separate files.

abargher:
Issue Score: Excellent

Comments:
Great work! This is very well documented.

PR #1126, opened 2025-05-19
[Achievements] Added advance function to achievement class

### LINKED ISSUE #625: Look into developing separate achievements API for adding new achievements into the game
Ideally achievements should support an API call where the game developers can add and make their own achievements. The API should also support a function that should award users with said achievement. We should also hold the data for the achievements and so whenever a user get's an achievement, the game should communicate with the users api to make a call to the achievements to then point an achievement from the user to the achievement.

ntagtmeier:
This PR is related to issue #625. It adds an advance function to the achievement class that can be used to update a user's progress towards an achievement. It also adds a test to make sure that function is working as intended.


emilyzen (approved):
Looks good to me!

abargher (approved):
Nice work, and thanks for including a test!

abargher:
Issue Score: Excellent

Comments:
Nice work!

PR #1135, opened 2025-05-19
[Checkers] Added possibility to double capture

### LINKED ISSUE #928: Add consecutive jumps/captures
Objective: the goal of this issue is to allow players to do two consecutive captures in the same turn if a piece is in position to do so after the first capture

enzeeo:
In the sandbox, I've added two new functions `getJumpPaths` and `executeJumpChain` in addition to editing the `highlightValidMoves`.

Now when a double capture is possible, this option will be highlighted alongside the single capture. The player can then choose between making a double or single capture. This addresses issue #928. Doing this will prevent the need to track turns in the case where a player captures one piece and has the option to skip or capture another piece.


brianHarder (approved):
I did double and triple captures for red and black, and everything seems to work as expected. The correct squares get highlighted on the board, and the correct pieces get captured. The score tallies update correctly as well.

The getJumpPaths and executeJumpChain functions make sense. LGTM!

abargher (approved):
Looks great, nice work! This is a tricky part of the game logic.

abargher:
Issue Score: Excellent

Comments:
Great work!

PR #1177, opened 2025-05-20
[users] Removed uniqueness constraint on FriendInvitation model

### LINKED ISSUE #1175: [users] Remove Uniqueness Constraint from FriendInvitation Model
A recent PR seems to have merged the uniqueness constraint of 'sender' and 'receiver' in the FriendInvitation model back into `model.py`. This is problematic for the codebase, as the FriendInvitation is designed to allow for past cancelled invitations in the database as record. "Deleting" invitations simply work by setting a `is_deleted` field to `True`. For future reference, and for future people looking into the codebase, the absence of a uniqueness constraint on sender and receiver is part of the design. Only one active `is_deleted=False` invitation is allowed in the database at one time, and it is enforced through `views.py` and the backend functions, but not through the django model constraints.

For this issue I will remove the constraint + add important note to the model to ensure this confusion gets cleared up for future developers of the codebase.

davdma:
Fixed: #1175

Removed uniqueness constraint on `FriendInvitation` recently merged into dev branch. Since I anticipate this might happen again, I added some documentation to highlight the need for soft deletes, and to remind developers that the lack of uniqueness constraint is part of the design, and that enforcement of one active invitation happens in `views.py` functions.


src/chigame/users/models.py
```diff
@@ -106,6 +106,14 @@ class FriendInvitation(models.Model):
     """
     An invitation from a User to another User, requesting that they become
     friends.
+
+    IMPORTANT NOTE TO DEVELOPERS BEFORE MODIFYING THIS MODEL:
```
hollyfwei, line 110: Great thought to add this note!

hollyfwei (approved):
Looks great David - way to keep up with the thorough documentation!

src/chigame/users/models.py
jonathanbacon22: Agreed!

jonathanbacon22 (approved):
Great clarification. When we made this change I did not think to add any documentation. Thank you! LGTM!

src/chigame/users/models.py
hossainmd-uc: Can't believe it was removed. Thanks for correcting it David

hossainmd-uc (approved):
LGTM

cmarshall282 (approved):
Good catch!

cmarshall282:
Issue Score: Excellent

Comments:
Great job!

PR #789, opened 2025-05-09
[achievements] Added fixture

ntagtmeier:
This pull request would add a fixture to the achievements folder that includes all of the Achievements and UserAchievements necessary to rewriting the current achievements frontend to be data-generated rather than hardcoded. This pull request closes issue #781 .

PR #1195, opened 2025-05-20
[achievements] Added URLs for achievements to users/

### LINKED ISSUE #889: Add urls to chigame
add achievement urls so it can be seen from the main chigame page

ntagtmeier:
This PR closes issue #789. It adds URLs to the users urls.py file that allow people to access their own achievements and those of others. There are also URLs that specify filters for a specific game or a specific status. As stands, the filtering has not been implemented, but the view function now has the parameters to account for it.


jiayuew1:
I agree with you that we should use the user ID as the primary key instead of the username. Also, could you clarify the difference between “No Progress” and “Unlocked”? I’m not entirely sure how they’re meant to be distinguished.

ntagtmeier:
> I agree with you that we should use the user ID as the primary key instead of the username. Also, could you clarify the difference between “No Progress” and “Unlocked”? I’m not entirely sure how they’re meant to be distinguished.

"Unlocked" means you actually have the achievement. "No Progress" means that you do not have the achievement and have never made any progress towards it.

jiayuew1 (approved):
Looks good!

abargher (changes requested):
There are a couple merge conflicts, please take a look-- I'm not sure which sections you'd like to keep.

ntagtmeier:
Resolved merge conflicts. The one in views.py was trivial and seemed to result from the confusion about the change to the signature of the user_achievements function. For the html, the changes that had been made to url references in dev didn't account for my added URLs, which should be definitive.

abargher (approved):
Looks good!

abargher:
Issue Score: Excellent

Comments:
Great job!

PR #1199, opened 2025-05-20
[Checkers] Only allow player to move their piece during their turn

### LINKED ISSUE #1190: [Checkers] Ensure that only associated player can move pieces
Right now player 1 can still move player 2's pieces. Make sure that player 1 can only move their pieces. This should be done by making sure that the email matches the user who can move.

enzeeo:
To address issue #1190, I added a current player tracker to the CheckersBoard model. Every time a player makes a move, the move now gets tracked and the pieces are now associated with a player. This means that players can only move their own pieces, and can only move them when it is their turn to play. 

To test, please run the following: 
1. `python manage.py makemigrations`
2. `python manage.py migrate`
3. `python manage.py loaddata checkers-fixture.json`
4. `python manage.py runserver`
5. Log in as `red@example.com`
6. Reset password and confirm email address as necessary 
7. Go to `games/checkers/1` and make a move. Now try making another move and you should be locked from clicking your own piece or the black piece 
8. Log in as `black@example.com`
9. Reset password and confirm email address as necessary 
10.  Go to `games/checkers/1` and make a move. Now try making another move and you should be locked from clicking your own piece or the red piece 
11. Closing the server and opening back up should not affect gameplay e.g. if it was black's turn when you closed the server, it should still be black's turn when you resume.


brianHarder (changes requested):
I followed all the instructions and was able to play as both red and black. The restrictions on pieces work properly: I can only move red pieces when logged in as red on red's turn, and I can only move black pieces when logged in as black on black's turn.

However, the forfeit and draw buttons didn't work for me. Forfeit should be pretty easy (only current player can forfeit), but draw requires the other player to be able to click accept which could get trickier. See if you can update the views or JS code to get these working.

brianHarder (approved):
Confirmed with Enzo that forfeit + draw functionality will be implemented in a later PR. With that aside, the player restriction works as expected and the code makes sense. LGTM!

abargher (changes requested):
There are some merge conflicts plus some issues with the migrations, please take a look.

PR #1203, opened 2025-05-20
[users] Fixed notification attribute error that causes page to crash

### LINKED ISSUE #1201: [users] Fix FriendInvitation notification breaking bug
After creating two users, and searching up one of the users from the other, sending an friend invitation leads to a website breaking bug for the user receiving the invitation. The base page throws an AttributeError: 'Notification' object has no attribute 'MATCH_PROPOSAL'. This is because the Notification model has a `MATCH_PROPOSAL` attribute that was renamed to `MATCH_INVITATION`, but there are still remaining references to `MATCH_PROPOSAL`. The easiest fix here is to update all these bad references. This should be done asap in order to facilitate further invitation testing and friend UI improvements.

davdma:
Fixed: #1201

While this is a small PR, this is an essential bug fix needed to get the dev working with notifications. Currently there are problematic references to `MATCH_PROPOSAL` which is an attribute renamed to `MATCH_INVITATION`. This results in AttributeError on the base page, rendering any notification a website breaking event. Thus, this PR ensures that everyone can test and configure the notifications and friending UI without running into issues.

What was done: changed all references to `MATCH_PROPOSAL` in Notification type to `MATCH_INVITATION`.


src/chigame/users/models.py
```diff
@@ -343,7 +343,7 @@ def get_style_key(self):
             self.FRIEND_REQUEST: "friend",
             self.REMINDER: "system",
             self.UPCOMING_MATCH: "match",
-            self.MATCH_PROPOSAL: "match",
+            self.MATCH_INVITATION: "match",
```
hollyfwei, line 353: Good catch - this has been a confusing term for a while now :)

hollyfwei (approved):
LGTM - thanks for fixing this!

src/chigame/users/models.py
jonathanbacon22: agreed!

jonathanbacon22 (approved):
Thank you for this change! Small fix that helps a lot!! LGTM.

cmarshall282 (approved):

cmarshall282:
Issue Score: Excellent

Comments:
Great job!

Issue #904, opened 2025-05-13
Implement Knowledge Base Single Game Guide Moderator View


yizou275:
As described by our knowledge base user stories, a moderator is able to manage the accepted guides for a single ChiGame game.

This issue specifically works on the frontend of the page, where has the pre-req of #794.

Luca-W-D:
Edited Yi's original description, since this issue was slightly misnamed; this was actually for the single-game moderator view!

Luca-W-D:
The original title may have been slightly confusing for people looking at this feature for the first time ^

Luca-W-D:
There are two very similar pages– one lists the game guides, and the second displays the guides for one of the games you select in the first guide. I'm going to open a PR which updates both, and therefore will complete the functionality of this PR.

Luca-W-D:
Closed by PR #1226, which implements all described functionality.

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! An improvement could be to describe what functionality the frontend needs so the issue is more clear

Issue #794, opened 2025-05-09
Implement Moderator Single Game Guide Management Backend


yizou275:
As described by our user stories, moderators manage the guides on the knowledge base site. They are able to see a list of all previous approved guides for a given game, and then choose one to publish/unpublish. This issue implements this page.

There are two sub-tasks:
1. Creating the HTML template file and any associated static assets (CSS, etc.)
2. Creating the backend interactions (mainly the view() function) so that the frontend pulls data from ChiGame

Luca-W-D:
Editing since this task is slightly too broad– we will add a separate issue for making the frontend for this feature.

yizou275:
This issue may include these subtasks:
-  first a page, "organize guides by game"
- for each game, the page display a list of guides in "approved" status and associated with the game
- can publish, unpublish, or switch for each entry (by switch, I mean automatically unpublishing the current guide and publish the selected guide)
- include a link to download/preview of the respective guide (to be decided after seeing the moderator preview page merged into dev)
Will add more features to this list if something else comes to mind.

yizou275:
Roughly complete the page "organize guides by game".

yizou275:
Added the single game approved list page, and allow publish/unpublish/switch on the backend.
Also modified the DownloadView so that moderators can also download an uploaded guide (besides the author of the guide).
Added no approved guide/no game for empty list in ListView.

yizou275:
Closes by #1043.

jacqsmei:
Issue Score: Excellent

Comments:
Good work breaking down the task into manageable parts!

PR #1226, opened 2025-05-20
[knowledge-base] Implement frontend for moderator single-game view

Luca-W-D:
As described by #904, our user stories define a use case where moderators review all the accepted guides for one game. Therefore, they can switch between the current version and a better, previous version. Or, they can publish a guide for a game for the first time.

This PR extends the barebones UI of #794 (which helped test the backend implementation) into a user-facing version.

## Files Changed
The moderator uses two pages to manage the guides for a single game. First, there's a list of games that they can choose to moderate. This includes some high-level information about the published guide state for each game. Then, after selecting a game, they are brought to a page that shows all previously accepted guide options.
* `moderator_game_list.html` is the first page that lists all possible games. I made changes to render the games as individual cards, similar to the behavior of the landing page. I also modified this page to display how many guides are available when a game has no selected guide.
* `moderator_single_game.html` is the second page that lists all previously published guides for an individual game. I changed this table to have all actions on the rightmost column, and I removed some debug styling.
* `views.py` required serverside filtering to check the number of *accepted* guides for a game versus all guides.

## Testing
Pull this branch and ensure the knowledge base fixture is loaded. If not, the instructions for loading our test data are described in `src/chigame/knowledge_base/fixtures/README.md`

Login as admin (default pswd `test12345`) and navigate to the knowledge base.

In the top right of the navbar, click Moderator Dashboard. At the bottom of the page, there is a link to manage published guides (this page probably should be reorganized in a future issue). Click this link.

You should be brought to a list of games that contains information about their published guide status. For example,
image
Ensure these links work as expected.

Then choose one game with a non-zero amount of guides. Ensure that you can visit it, and that you can switch the active guide.

image


mrollinswaterman (approved):
Everything looks good to me. One thing to consider is potentially making it so that if a guide is approved and there are no published guides for that game, we could automatically publish the approved guide? I think this would save some time on the moderators part because they wouldn't have to manually publish every new game guide, but I can also see how automating publishing could be dangerous / produce unintended consequences. Just a thought.

Luca-W-D:
That's a good point, and I think that likely should be contained in a new issue as a backend enhancement. I'll bring it up in our final presentation meeting. Thanks Miles!

jacqsmei (approved):
lgtm! page looks nice and clean

jacqsmei:
Issue Score: Excellent

Comments:
Great job! And future considerations in the comment thread

PR #1227, opened 2025-05-20
[achievements] test user_achievements template view

### LINKED ISSUE #1202: Create tests for the user_achievements template and view functions
To ensure coverage for the functions implemented in PR #1037 , we need to possibly create additional factories and unit tests for the functions so that other/future teams can use the functions in their implementations. These tests should ensure that the User-Specific Achievement Display and Urls work properly. Note that currently, only the "user_achievments" is in use in the template, and so tests do NOT cover the "user_achievements_by_username" url until that url is implemented into the html template.

bensim0305:
To address issue #1202 , I wrote unit tests testing the url pattern "user_achievements" and the views function user_achievements. This also tested the user_achievements html template, ensuring that cases where there are achievements unlocked, in progress, or none at all were behaving properly.


jiayuew1 (approved):
Tested it. Good to go!

abargher (changes requested):
More merge conflicts, similar to the previous PR with test changes. I expect those to help fix these conflicts.

bensim0305:
Fixed merge conflicts!

abargher (approved):
Looks like these commits were pulled into another PR, so this should be safe to close (it will still be graded). Generally we want to avoid this, but this was more about the order in which I merged things.

Issue #1210, opened 2025-05-20
Groups Front End - Implement GroupCreateView & GroupDeleteView


hollyfwei:
Add GroupCreateView and GroupDeleteView along with associated url paths and templates. [basic implementation]

PR #1237, opened 2025-05-20
[User Profiles + Friending] GroupCreateView and GroupDeleteView

hollyfwei:
Addressing issue #1210 

- added GroupCreateView
- added GroupDeleteView
- added group create url
- added group delete url
- added group_create.html
- added group_delete_confirm.html


davdma (approved):
LGTM Holly! Thanks for continually adding to the groups - hopefully we can get the implementation in good shape very soon.

cmarshall282 (approved):
Looks good. Please resolve the linter before I can merge.

src/chigame/users/models.py
```diff
@@ -155,7 +155,7 @@ class Group(models.Model):
 
     name = models.TextField()
     description = models.TextField(blank=True)
-    members = models.ManyToManyField(User)
+    members = models.ManyToManyField(User, blank=True)
```
hossainmd-uc, line 158: Subtle, but an important change !

src/chigame/users/models.py
```diff
@@ -189,6 +189,9 @@ def delete(self):
         self.is_deleted = True
         self.save()
 
+    def __str__(self):
+        return f"Invitation from {self.sender} to {self.receiver} for {self.friend_group}"
+
```
hossainmd-uc, line 194: What is this used for? I'm wondering if it is used for notifications. If so, how does it integrate with the current model (since, for example, I don't think friend invitations has such a mechanism)

src/chigame/users/views.py
```diff
@@ -817,3 +820,43 @@ class GroupListView(SingleTableView):
 class GroupDetailView(DetailView):
     model = Group
     template_name = "users/group_detail.html"
+
+
+class GroupCreateView(LoginRequiredMixin, CreateView):
```
hossainmd-uc, line 864: I'm wondering if there's a particular reason you used class based views, but I think it gets the job done!

hossainmd-uc (approved):
LGTM!

cmarshall282:
Issue Score: Satisfactory

Comments:

Issue #1209, opened 2025-05-20
Groups Front End - Implement GroupJoinView & GroupLeaveView


hollyfwei:
Add GroupJoinView and GroupLeaveView along with associated url paths and templates. [basic implementation]

PR #1241, opened 2025-05-20
[User Profiles + Friending] GroupJoinView and GroupLeaveView

hollyfwei:
Addressing issue #1209 

- added GroupJoinView
- added GroupLeaveView
- added group join url
- added group leave url
- added group_join.html
- added group_leave.html


davdma (approved):
LGTM Holly! Thanks for keeping the fixtures up to date also as you continue building out the groups implementation!

src/chigame/users/fixtures/users_fixture.json
```diff
@@ -116,6 +116,31 @@
     "date_joined": "2023-10-30T03:31:56.761Z"
   }
 },
+{
+  "model": "users.user",
+  "pk": 8,
+  "fields": {
+    "password": "foo",
```
cmarshall282: The password won't work as intended here

cmarshall282 (changes requested):
I was unable to load your fixute, but otherwise this looks good. Please take a look at that.

src/chigame/users/fixtures/users_fixture.json
harriet-fisher: This hash is a Django hash for "test" if you want to use this instead @hollyfwei 
argon2$argon2id$v=19$m=102400,t=2,p=8$Zzk2ZjFCTXl4RlJmOW1VU1JHa3QxZA$tiaFDXObRmwzYoVytjyCGFVfJnoqp1uYUOvCVUsUHpY

src/chigame/users/models.py
```diff
@@ -152,7 +152,7 @@ class Group(models.Model):
 
     name = models.TextField()
     description = models.TextField(blank=True)
-    members = models.ManyToManyField(User)
+    members = models.ManyToManyField(User, blank=True)
```
harriet-fisher: good catch for validation, I hadn't thought about the potential for error with 0 members upon creation. Another solution could be that any group "creator" is always a de facto "member", but that only works as validation if we implement the logic for that simultaneously, so maybe we can leave that as a possibility for future dev.

src/chigame/users/urls.py
```diff
@@ -62,4 +62,11 @@
     path("labels/create/", views.create_notification_label, name="create-notification-label"),
     path("labels/<int:label_id>/notifications/", views.notifications_by_label, name="notifications-by-label"),
     path("inbox/<int:pk>/<str:category>/", views.user_inbox_view, name="user-inbox-category"),
+    # Group paths
+    path("groups/", views.GroupListView.as_view(), name="group-list"),
+    path("groups/<int:pk>/", views.GroupDetailView.as_view(), name="group-detail"),
+    path("groups/create/", views.GroupCreateView.as_view(), name="group-create"),
+    path("groups/<int:pk>/delete/", views.GroupDeleteView.as_view(), name="group-delete"),
+    path("groups/<int:pk>/join/", views.GroupJoinView.as_view(), name="group-join"),
+    path("groups/<int:pk>/leave/", views.GroupLeaveView.as_view(), name="group-leave"),
```
harriet-fisher, line 105: Thank you for leaving the header comment, definitely keeps the urls cleaner when they're grouped like that

src/chigame/users/views.py
```diff
@@ -792,3 +797,94 @@ def notifications_by_label(request, label_id):
...
+    model = Group
+    fields = ["name", "description", "members"]
+    template_name = "users/group_create.html"
+    success_url = reverse_lazy("group-list")
+
+    def form_valid(self, form):
+        form.instance.created_by = self.request.user
+        response = super().form_valid(form)
+
+        self.object.members.add(self.request.user)
```
harriet-fisher: Looks like you implemented the admin/member adding as well! We may not want to allow blank=True in this case, though I'm not a data validation expert, so I'm not sure if it's good practice to keep the blank=True during creation since the admin is technically "added" and isn't initially there

src/templates/users/group_detail.html
```diff
@@ -0,0 +1,28 @@
...
+  <div class="container">
+    <div class="row">
+      <div class="col-sm-12">
+        <h2>Group Details:</h2>
+      </div>
+    </div>
+    <div class="row">
+      <div class="col-sm-12">
+        <h3>{{ object.name }}</h3>
+        <p>Group ID: {{ object.id }}</p>
```
harriet-fisher: Also not a necessary change-- more for future dev, but we may not need the Group ID, since the id of Group models is an incremental integer, so not super relevant from a UI/UX perspective.

src/templates/users/group_join.html
```diff
@@ -0,0 +1,12 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <h1>Join Group</h1>
+  <form method="post" action="{% url 'users:group-join' group.id %}">
+    {% csrf_token %}
+    <p>Are you sure you want to join the group "{{ group.name }}"?</p>
```
harriet-fisher, line 7: Also a super nit-picky comment, but maybe we could rephrase this to "You're about to join {{  group.name }}". No logical reason, it just feels like the tone matches it a little better ("are you sure" feels a little bit like a warning against joining the group)

harriet-fisher (approved):
Looks great! Thanks for all your work on the group stuff, Holly! It adds a whole other dimension to the project. I left some nit-picky UI comments, but no qualms with your implementation. I also replied to Chris's comment with the pw hash I generated for "test" as a password, so if you want to use that for some of your user fixtures, it should work!

cmarshall282 (approved):
I had to regenerate lots of migrations. Looks good now

cmarshall282:
Issue Score: Satisfactory

Comments:

PR #1245, opened 2025-05-20
[Interactive Fiction] Fix game deletion issue

### LINKED ISSUE #1196: Fix issue with being unable to delete games
A problem we have run into while testing is that once a game is created, games are unable to be deleted off the page, even from both the admin side and the user side.

Goal:
- Fix behavior so that if a game is deleted from anywhere, it deletes and is removed properly without crashing the site

zaynacheema:
A problem we have run into while testing is that once a game is created, games are unable to be deleted off the page, even from both the admin side and the user side.

Goal:

Fix behavior so that if a game is deleted from anywhere, it deletes and is removed properly without crashing the site

What we did to fix this issue:
- Added functionality and a template for game deletion, a page that asks if the user is sure they want to delete the selected files
- Made a migration file called leaderboards, scans for changes on terminal, putting the changes in the file so it is not destructive.

Tested:
- Tested by making games to delete off of the user page and check if they were successfully removed from local storage
- Checked to see if the same applied on the admin side


gkharrell01 (changes requested):
image
This is the error I am getting when I confirm that I want to delete!

zaynacheema:
After pulling the branch and running python manage.py migrate , it should work now, the file I created was not added properly

gkharrell01 (changes requested):
Hi! I figured out that I was able to pull this file, but the __init__.py was not in the leaderboards/migrations/ with the 0001 migration, which doesn't register this migration when running manage.py migrate. Once I added an empty __init__.py to the folder, it worked great! Maybe add that and it should be good to go!

zaynacheema:
Thank you so much for that catch! I've added the file and hopefully it should work now.

src/chigame/leaderboards/migrations/__init__.py
```diff
@@ -0,0 +1 @@
+
```
gkharrell01, line 1: Migration works for me!

src/chigame/games/views.py
```diff
@@ -567,6 +567,19 @@ def get_context_data(self, **kwargs):
...
 
 
+class GameDeleteView(DeleteView):
+    model = Game
+    template_name = "games/game_confirm_delete.html"
+    success_url = reverse_lazy("game-list")
+
+    def dispatch(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        # Only staff members can delete games
```
gkharrell01, line 611: Appreciate the comments - very readable.

src/templates/games/game_confirm_delete.html
```diff
@@ -0,0 +1,13 @@
+{% extends "base.html" %}
+
+{% block content %}
+  <div class="container mt-4">
+    <h2>Delete Game</h2>
+    <p>Are you sure you want to delete "{{ game.name }}"?</p>
```
gkharrell01, line 6: Cool feature!

gkharrell01 (approved):
Requested changes work great and fixed functionality issue. Good work!

jacqsmei:
is there context on why is a leaderboards migration?

zaynacheema:
> is there context on why is a leaderboards migration?

It came up in the delete button fix but it looks like leaderboards made a table but there was no migrations to make for some reason, but this should fix the issue and it was messing with our ability to delete the games.

jacqsmei (approved):
good work!

jacqsmei:
Issue Score: Excellent

Comments:
Nice teamwork on getting this to work!

PR #1255, opened 2025-05-20
[knowledge-base] Homoegenize Moderator Dashboard and Manage My Guides styling and formatting

### LINKED ISSUE #1213: Homogenize Moderator and Contributor page frontend stylings
As it stands, the Moderator and Contributor Dashboards have different frontend styling when it comes to buttons and columns. This difference can create a jarring user experience, so this Issue aims to unify them. Specifically, it will bring the Contributor page in line with the Moderator page which is a closer match to the rest of the Knowledge Base and Chigame in general.

mrollinswaterman:
## Overview
This PR fixes Issue #1213. CSS styling has been removed and altered in order to align the styling for the Moderator and Contributor pages.

## File Changes
**`src/static/css/knowledge-base-styles.css`**
- Removed some CSS that made the Contributor guide table look darker than its Moderator Dashboard counterpart.

**`src/templates/knowledge-base/contributor_manage_guide.`**
- Added some CSS and HTML code to make the buttons on the Contributor page be the same as those in other parts of the Knowledge Base.
- Made the table my concise by removing the separate column for Guide IDs and the Review Feedback action. These have been collapsed into other columns.
- Changed the class of the download guide Icon for a cleaner look that aligns better with the rest of the page/site.

## Testing
2. Log in as any user with Guides uploaded.
3. Go to your Manage My Guides Page
4. Your page look like the screenshot below:

Screenshot 2025-05-20 at 4 56 44 PM


mrollinswaterman:
I discovered a bug where the status of guides with no review or pending review was incorrectly displayed. It has now been fixed, so that if your guide has no review yet, the status will be listed as "Pending..." and it will not be clickable, since there would be no feedback to view. Page should look like the screenshot below.


Screenshot 2025-05-21 at 2 26 56 PM

mrollinswaterman:
There was a mysterious merge conflict on nearly all my PRs with a migration file in src/chigame/chat/migrations. Removing the file and pushing to the branch seems to have solved the issue.

mrollinswaterman:
There was a mysterious merge conflict on nearly all my PRs with a migration file in src/chigame/chat/migrations. Removing the file and pushing to the branch seems to have solved the issue.

yizou275 (changes requested):
Wow thanks Miles! You definitely did a lot of helpful work in this PR! I pulled the branch and the page looks as expected. But a few observations:

- The sorting doesn't seem to work? Though the "oldest update" does order them correctly, the default is not by "recentest update" (I guess maybe they order it default in increasing ID). I think to make the sorting work, maybe you also need to add some codes to `views.py`. It may be helpful to take a look at the codes for other sorting pages.
- Regarding status, the black text "accepted" on green button or "rejected" on red button seem difficult to see? Would it help to change the text color to white? (I am aware that white text may not work for "change requested"'s yellow button, we may need to test how we deal with colors here)
- Lastly, this version doesn't include the resubmit link. Like what we discussed, we could put it into a future PR. I just wish to note that maybe we need to consider whether we want to include this in the demo (it might be helpful in explaining how "request change" process in knowledge base works); if we do want to do that, maybe be aware that we need to open that PR by tomorrow (the soft deadline).

Feel free to share your thoughts! Thanks again! :)

mrollinswaterman:
Thanks for you comments Yi! I noted the sorting bug a while ago and I opened a different PR/Issue since it seemed a little outside the scope of this one. 

I agree about request changes, and I'm happy to open that issue today or tomorrow, but I still think it's not something this PR needs to deal with. 

As for the button text, I kind of agree with your assessment, but I tried making the buttons light text instead of dark, and it looked... weird to say the least. The dark text looked much cleaner in my opinion, and I think the button color code system is robust enough that users don't necessarily need to read the text to be able to tell the status of the their guide.

yizou275 (approved):
Ok this sounds like a good plan! Thank you Miles! I'll approve this PR now :)

jacqsmei (approved):
nice work! this was a good idea

jacqsmei:
Issue Score: Excellent

Comments:
Great job! And nice teamwork in the review as usual

PR #876, opened 2025-05-12
Game management/create join match

pauline196:
User is able to initiate a new match by creating a Lobby for a specific game. The user is able to invite players by selecting their emails in the match creation form. Once submitted, the system automatically generates a new Lobby and assigns it a unique join code. All invited members should be able to join the lobby using the provided code.

PR #1264, opened 2025-05-21
Implemented join match functionality

### LINKED ISSUE #645: Add "create a match" and "join a match" functionality
User should be able to initiate a new match by creating a Lobby. In a Lobby, user should be able to: 
- select a game they want to play 
- provide a lobby name 
- provide min_players and max_players

When created, user should be automatically added to the lobby's members. A match should not be shown as created immediately: it is only started once the number of users who have joined lobby meets specified min_players

pauline196:
This PR allows all invited members to join a match lobby using a unique join code. Once all invited users have joined, the lobby is automatically closed to prevent additional entries. The logic ensures that each user can only join once—if a user attempts to join again, they are recognized as already entered and not added again. The implementation also handles edge cases such as invalid codes, full lobbies, and unauthorized join attempts with clear user feedback.

Originally, we had another PR for this issue (https://github.com/uchicago-cs/chigame/pull/876), but it got very out-of-date, so we created a new PR.


pauline196:
**Testing instructions**

1. Log in as a user invited to a match
2. Go to the match join page and enter the correct join code
3. Confirm you are added to the lobby and see a success message
4. Try joining again and check that you’re not added twice and see an info message
5. Repeat with all invited members; once the lobby is full, verify it is closed
6. Try joining with an invalid code or as an uninvited user and verify tgat error messages are shown

Abebes01 (changes requested):
Hey - I noticed that when all invited players join a game like 'Farmland Empire' - the lobby never closes even if all invited players have joined. However - when I join a game like Chess and my invited player joins - the lobby closes. I believe the reason why this implementation closes the lobby in the game of Chess but not in Farmland Empire is because you are checking against the games upper limit of players - and not the number of invited players.  This should be fixed since the description for this PR states that the lobby should close once all invited players have joined - not when it reaches the games player limit "Once all invited users have joined, the lobby is automatically closed to prevent additional entries" 

More specifically, in your join_match function, you're checking against lobby.max_players (which is set to the game's max_players if Im not mistaken):
pythonif lobby.members.all().count() == lobby.max_players:  
    lobby.match_status = 2
But I believe what you want is to close the lobby when all invited members have joined, not when it reaches the game's maximum capacity.
So I would change this line in join_match:
Current:
pythonif lobby.members.all().count() == lobby.max_players:
Should be:
pythonif lobby.members.all().count() == lobby.invited_members.all().count():

Abebes01 (approved):
LGTM

MuyanXie (approved):
LGTM. Approved!

MuyanXie:
Issue Score: excellent

Comments:
The review is very thorough and the implementation is very thoughtful -- the edge case handling is very smooth!

Issue #1267, opened 2025-05-21
Allow Tournament Owners to View Their Own Tournaments


carriesunluo05:
Currently, tournament visibility is limited only to assigned players. This creates a bug where tournament owners cannot view their own tournaments unless they explicitly assign themselves as players. This behavior is unintuitive and inconsistent with ownership privileges.
We need to fix this by modifying the logic so that tournament owners always have visibility of their tournaments, regardless of their player assignment status.

Here are some steps:
TournamentListView Update: Update the backend query/filter logic to include tournaments owned by the user, even if they are not listed as a player.
Testing: Ensure that owners can view their tournaments under "Tournaments".

carriesunluo05:
Issue closed as PR has been merged.

MuyanXie:
Issue Score: excellent

Comments:
Excellent job implementing this issue! It's very informative and guides the PR well.

PR #1268, opened 2025-05-21
[Tournaments] Allow Tournament Owners to View Their Own Tournaments

carriesunluo05:
We updated the tournament visibility logic so that tournament owners can now see tournaments they own, even if they are not assigned as players. Previously, owners had to manually assign themselves as players to access their own tournaments, which was not intuitive or aligned with expected behavior.

With this fix, the dashboard and relevant views now include tournaments where the logged-in user is the owner, in addition to those where they are a player.

Issue: #1267


carriesunluo05:
How to test (the instruction might be slightly long, but the testing should not take that much time):

1. make the migrations

2. Run: python3 manage.py loaddata src/chigame/games/fixtures/feedback-fixture.json

3. Then: python3 manage.py runserver

4. Create superuser: python3 manage.py createsuperuser

5. Runserver again

6. Go to admin panel: http://127.0.0.1:8000/admin/ → Find Users under the Users section → Click on player 1 → fix the password based on your choice (note: We have to do this as I can’t tell what the password is in English); also click on rec_user1@example.com --> fix the password based on your choice

7. Go back to http://127.0.0.1:8000/; (you might have to verify your email, which will be the link in the console log); log in as player 1

8. Go to tournamnets page. If you see Past Tournament by User 101, the test passes. This is because player 1 created this tournament but isn't a player, so before this bug, player1 wouldn't be able to see this tournament, but now they can.

9. Log in as rec_user1@example.com --> go to tournaments page --> make sure you see Past Tournament by User 201

kwanahita (approved):
Hi Carrie! Everything works as expected. Only problem is I just merged dev and now there's one formatting error. Should be a very quick fix.

MuyanXie (approved):
LGTM. Approved! Please fix the merge conflict

MuyanXie:
Issue Score: excellent

Comments:
Great job implementing the visibility rule! This is very intuitive and greatly help the owner of the tournaments

Issue #1234, opened 2025-05-20
[Checkers] Merge sandbox highlight


enzeeo:
This issue addresses merging the ability to highlight all valid moves. This code is currently in the sandbox and should be moved over to the actual game.

abargher:
Issue Score: Excellent

Comments:
Nice work!

PR #1274, opened 2025-05-21
[Checkers] Implement highlight

enzeeo:
To address issue #1234, I've merged the highlight of potential moves the player can make. This also required editing some of the code to be compatible with the mirrored pieces of player 2's POV. 

To test this out run the following: 
1. `python manage.py runserver`
2. Log in as `red@example.com` and reset password and confirm email as necessary
3. Go to `games/checkers/1` and click on a red piece, it should highlight potential moves 
4. Log in as `black@example.com` and reset password and confirm email as necessary
5. Go to `games/checkers/1` and click on a black piece, it should highlight potential moves


enzeeo:
Made a new PR #1276 instead

PR #1002, opened 2025-05-15
[Leaderboards] Add privacy settings views, urls, and templates

bacBenChaplin:
This is to add 3 new views for the leaderboard privacy settings. One to list them, one to mangage them, and one to delete them. There are new urls to support this. There are also some new templates

Note: The templates were generated by AI. I did this so I could merely interact with this stuff on the web app. I think these should be made nice in another issue.

Note: This should be added after the model about the leaderboard privacy setting has been pushed (#773).

Steps:
1. Make sure that migrations are up to date. Probably need to make migrations for the leaderboard models.
2. run the server with `python3 manage.py runserver`
3. open the web app shell with `python3 manage.py shell`
4. run the following to generate some data in the database (Full disclosure: this was AI generated):
```from django.contrib.auth import get_user_model
from chigame.users.models import UserProfile
from chigame.games.models import Game
from chigame.leaderboards.models import Leaderboard, Metric

# Get User model
User = get_user_model()

# Create a user with no settings
try:
    user = User.objects.create_user(
        username="casey",
        email="casey@example.com",
        password="password123",
        name="Casey"
    )
    profile = UserProfile.objects.create(
        user=user,
        bio="A user with no privacy settings"
    )
    print(f"Created user: {user.username}")
except:
    user = User.objects.get(username="casey")
    profile = UserProfile.objects.get(user=user)
    print(f"Found existing user: {user.username}")

# Create two games
games = []
for name in ["Chess", "Checkers"]:
    game, created = Game.objects.get_or_create(
        name=name,
        defaults={
            'description': f'A game of {name}',
            'min_players': 2,
            'max_players': 4,
            'complexity': 2.5
        }
    )
    games.append(game)
    status = "Created" if created else "Found existing"
    print(f"{status} game: {name}")

# Create metrics and leaderboards
for game in games:
    # Create two metrics per game
    for metric_name in ["Points", "Time Played"]:
        metric, created = Metric.objects.get_or_create(
            name=f"{game.name} {metric_name}",
            game=game,
            defaults={
                'description': f'{metric_name} for {game.name}',
                'unit': 'points' if 'Points' in metric_name else 'minutes'
            }
        )
        status = "Created" if created else "Found existing"
        print(f"{status} metric: {metric.name}")
        
        # Create a leaderboard for each metric
        leaderboard, created = Leaderboard.objects.get_or_create(
            name=f"{game.name} {metric_name} Leaderboard",
            game=game,
            defaults={
[...]

PR #1288, opened 2025-05-21
[leaderboards] Making sure that one cannot be BOTH opted out AND anonymous on leaderboard privacy setting.

### LINKED ISSUE #1187: Make sure that leaderboard privacy setting does not allow both anonymous and opt out
This is to make sure that one cannot set a privacy setting where they are BOTH anonymous AND opted-out, as that does not make sense

bacBenChaplin:
This is the make sure that no one can be opted out of a leaderboard privacy setting and anonymous at the same time.

NOTE: This should be pushed after #1002 is pushed

This makes sure that it's not possible from 3 angles: 
1. the form submission
2. the model itself / database itself
3. the frontend

Steps to review:

1. make sure database is properly migrated
4. make sure fixture is loaded
5. run server with data
7. go to `http://127.0.0.1:8000/leaderboards/privacy/` (log-in if necessary)
8. click to create any privacy setting
9. You should NOT be able to select opt out and anonymous - one box should grey out when you click the other box.
10. save the setting
11. go to edit it the same setting
12. it should be that the box is still greyed out


Aufsteyegen (approved):
Looks great. Excellent job on implementing this validation at all three levels (frontend, form, and model).

cmarshall282 (approved):
Thanks for adding this clarity!

cmarshall282:
Issue Score: Excellent

Comments:
Great job!

Issue #1197, opened 2025-05-20
Implement a way for created IF games to be stored in the database


zaynacheema:
An issue we've run into is that uploaded IF games are stored locally and only visible to different people. A possible workaround in order to implement a way for created IF games to be stored in the database is to do so that when someone runs migrations they can see the game on their site no matter who they are. 

Since the website is not deployed, this is kind of not really feasible in a normal way but perhaps this can be done by editing the game model.

zaynacheema:
I will create a new branch called interactive-fiction/database-storing for this issue.

zaynacheema:
The main step is to store the twine file in binary and is accessible at a link _resembling_ http://127.0.0.1:8000/games/twine-db/23/ (this is an example, like it would have the "twine-db" part)

After uploading "test-db" game I checked:

<img width="316" alt="Image" src="https://github.com/user-attachments/assets/636217f3-c894-4764-9ee5-fc58e9334817" />

And confirmed it was saved in binary. The next step is to push the code in a PR and see if others can access the game but otherwise it has successfully been store in bytes.

If the user pulls the branch and runs migrate: python manage.py migrate
they should see a game called "Midterm Project"

zaynacheema:
The PR is linked above

zaynacheema:
Due to merge conflict issues and under Jacqs guidance I have made a new branch with the commits from the previous branch but actually fixing the conflicts.

This branch is: interactive-fiction/clean-database-upload

And the new PR is linked below

PR #1291, opened 2025-05-21
[Interactive Fiction] Database storing twine files in byte type

zaynacheema:
An issue we've run into is that uploaded IF games are stored locally and only visible to different people. A possible workaround in order to implement a way for created IF games to be stored in the database is to do so that when someone runs migrations they can see the game on their site no matter who they are.

This is related to issue #1197 

Since the website is not deployed, this is kind of not really feasible in a normal way but perhaps this can be done by editing the game model.

The main step is to store the twine file in binary and is accessible at a link resembling http://127.0.0.1:8000/games/twine-db/23/ (this is an example, like it would have the "twine-db" part)

After uploading "test-db" game I checked:

Image

And confirmed it was saved in binary. 

**The next step is to see if others can access the game but otherwise it has successfully been store in bytes which was the goal** and achieved by altering the game model to store the file as a binary field rather than file field and updating all other files referencing it accordingly.

In order to see if others can access the same game, I will put the game in a fixture file 
Which you can get by running python manage.py migrate
After pulling the branch.
You should see a game called "Midterm Project"


gkharrell01 (changes requested):
When looking for a twine game, I recieve this error. Looks to be merge conflicts that need to be resolved! Let me know if I can help!
image

zaynacheema:
> When looking for a twine game, I recieve this error. Looks to be merge conflicts that need to be resolved! Let me know if I can help! image

It looks like the issue is coming from another group's "friendinvitation" not being in our database so I made a migration and pushed it, it should be okay now.

Note, this functionality should also be tested on "Create Game" not create IF game as that was only recently pushed during this PR.

And reminder if you run migrations there should be a game Midterm project visible.

src/chigame/games/migrations/0039_game_twine_file_content_game_twine_file_name.py
```diff
@@ -0,0 +1,23 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("games", "0038_merge_20250519_0011"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="game",
+            name="twine_file_content",
```
gkharrell01: Smart to use original game model!

src/chigame/games/migrations/0044_game_twine_file_name.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 4.2.20 on 2025-05-21 05:06
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
gkharrell01: There are a lot of migrations, I wonder if there is an ability in a further PR to make these cumulative.

src/chigame/games/views.py
```diff
@@ -152,14 +152,18 @@ def get_context_data(self, **kwargs):
     # Ensure the uploaded Twine .html file is saved to the Game model
     def form_valid(self, form):
         self.object = form.save(commit=False)
-        # ✅ Manually assign uploaded file
-        if self.request.FILES.get("twine_file"):
-            self.object.twine_file = self.request.FILES["twine_file"]
+        # save both the file and its contents
```
gkharrell01, line 160: Nice, readable comments

src/templates/games/game_form.html
```diff
@@ -28,6 +28,11 @@ <h1>
       </div>
       <br />
     {% endfor %}
+    <!-- Twine upload (manual field) -->
+    <label for="id_twine_file">Upload Twine Game (.html):</label>
```
gkharrell01, line 32: Thinking this will be updated once we merge other templates in other branches into dev

gkharrell01 (approved):
Looks good to me! Exciting that a game can be stored! Exciting to see more tangible work.

zaynacheema:
Test seems to be failed now from some tournament and lobby objects after merging with dev... not sure what to do as that's unrelated to us.

jacqsmei:
following up on this -- it would also be helpful if this game had a different name to match its purpose

zaynacheema:
> following up on this -- it would also be helpful if this game had a different name to match its purpose

Changed! It is should now be named "Twine Test Game"

zaynacheema:
i've run into like a million merge conflicts and errors particularly from this repeating error:
ls: src/chigame/forum_conversation/models.py: No such file or directory

i'm not even sure how this is related and it does not even exist on my thing so i'm going to just try again to make a placeholder for it...

zaynacheema:
Ran into even more merge conflicts and have no idea how to resolve this since its not really related to my code? maybe something from working with the database but I have no idea

PR #1304, opened 2025-05-21
[achievements] Added achievement percentage method + tests

### LINKED ISSUE #1279: Create helper function for achievement percentage
The end product of this issue is a helper function in the achievement model that calculates what percentage of an achievement's game's players have gotten that achievement.

ntagtmeier:
This PR includes functionality to calculate the percent of a game's users who have obtained a given achievement, which is information we will eventually want to display in achievement notifications. It also includes a helper function to get a user achievement without raising an error if no such user achievement exists as well as tests for this funcitonality.


bensim0305 (approved):
looks good to me! We should note that we should add more test cases.

abargher (changes requested):
Looks like some more merge conflicts.

ntagtmeier:
The conflict in models.py was trivial, but the renaming of tests.py is problematic. I added tests.py back in, because I don't think this is an appropriate function to put in test_views.

abargher (approved):
Looks good!

abargher:
Issue Score: Excellent

Comments:
Great job! Good documentation of reasons for changes.

PR #1305, opened 2025-05-21
[achievements] Altered fixture

### LINKED ISSUE #1186: Improve fixture for demo
Ideas from Jiayue -
Include long descriptions of an achievement (possibly exceeding 20 words).
Add pre-loaded, pinned achievements (ideally different ones for different users)
Consider a scenario where one user has no progress in a game while another user does. In such cases, we might not need to display the game’s achievement panel. (I’m not entirely sure about this—should we show the achievement panel for a game if a user has played it but hasn’t earned any achievements, with all achievements displayed as locked?)

ntagtmeier:
Changes made were in line with the guidelines set out in issue #1186. Despite this fixture being based on sample-achievements, I made a new file called demo-achievements in case people want to test on earlier versions of the models. The specific changes I made were:

- Ensuring that both pre-loaded users had pinned achievements
- Making sure that the last_updated field in the new UserAchievement model was specified when appropriate.
- Adding a long description to one of the achievements for collapse testing
- Specified users for the games in case we want to test functionality where games players haven't played don't show up on their achievements profile page
- Fixing an error where the "Completionist" achievement was awarded when progress was 1 despite the threshold being 3


jiayuew1 (approved):
Tested it. Good to go!

jiayuew1:
Actually I think you could use both name and username, they are both valid, and the current template display the `name`

ntagtmeier:
> Actually I think you could use both name and username, they are both valid, and the current template display the `name`

Oh, you're right. For the record, the issue came up because "name" does NOT work if you try to base the URL on the username, as we have done in the URLs in dev right now. However, PR #1195 should make this irrelevant since it uses the PK instead.

jiayuew1:
> > Actually I think you could use both name and username, they are both valid, and the current template display the `name`
> 
> Oh, you're right. For the record, the issue came up because "name" does NOT work if you try to base the URL on the username, as we have done in the URLs in dev right now. However, PR #1195 should make this irrelevant since it uses the PK instead.

Yes, we did use the pk in the url. I wasn’t clear in my explanation earlier. What I meant is that on the achievement page, the title is displayed as “XXX's Achievements,” where “XXX” is derived from user.name. So I believe we should ensure that the name attribute has some value.

ntagtmeier:
> > > Actually I think you could use both name and username, they are both valid, and the current template display the `name`
> > 
> > 
> > Oh, you're right. For the record, the issue came up because "name" does NOT work if you try to base the URL on the username, as we have done in the URLs in dev right now. However, PR #1195 should make this irrelevant since it uses the PK instead.
> 
> Yes, we did use the pk in the url. I wasn’t clear in my explanation earlier. What I meant is that on the achievement page, the title is displayed as “XXX's Achievements,” where “XXX” is derived from user.name. So I believe we should ensure that the name attribute has some value.

I see. I've made it so both users in the fixture have identical values for "username" and "name".

jiayuew1 (approved):
Looks good!

abargher (approved):
Looks good!

abargher:
Issue Score: Excellent

Comments:
Great work! Nice job working through requested changes.

PR #1299, opened 2025-05-21
Adding migrations and region field

karunkundrum:
This PR was originally to add all the base views for region and game but that was too large of a scope given the tasks that need to be completed before. So in this I migrated all of the leaderboard models and added the region field to the LeaderboardEntry.

PR #1316, opened 2025-05-21
[Leaderboards] Added missing migrations

cmarshall282:
The leaderboards team is missing migrations (they were not automatically detected). This PR adds the migrations for the current leaderboards models and was tested to make sure any subsequent changes would automatically be detected when running `makemigrations`. The root of the problem seems to be that without a migrations folder setup (and without containing \_\_init\_\_.py), no migrations are detected.

Thanks to the APIs team for discovering this!


MuyanXie (approved):
LGTM. Thanks for fixing this!

cmarshall282:
Resolved by PR #1299

PR #1322, opened 2025-05-21
[Word Game] Change Daily Word Algo

### LINKED ISSUE #1270: Daily Word Selection Improvements
Currently it seems that the daily word selection only takes words in alphabetical order, leading every daily word this month to start with the letter A. This is a bit of a problem, as we would prefer good breadth and unpredictability in our daily word selection process. The issue seems to stem from issues in the algorithm that selects the daily word, and perhaps a different and more complex algorithm should be used.

vtcrenshaw:
This pull request changes the algorithm for selecting the daily word. This needed updating as the old one was too predictable. We now have a hashing function that generates a difficult-to-predict word based on the date in string format. Solves issue #1270


schoi12-uchicago (approved):
This makes a lot of sense. Thanks for making these changes!

cmarshall282 (approved):
Great idea. This is much more straightforward

cmarshall282:
Issue Score: Excellent

Comments:
Great job!

PR #1364, opened 2025-05-22
Added cleaning capabilities and factory logic to UserAchievement

### LINKED ISSUE #868: update factories
TENTATIVE ISSUE: update factories to account for new model attributes

### LINKED ISSUE #1323: Create verification for agreement of UserAchievement attributes
Right now, it is possible for a UserAchievement to have a last_updated date that differs from its date_earned date. If date_earned is null, this is clearly appropriate. If not, its not clear why this should be allowed. This issue is related to #868 because it will give us a way of ensuring that the UserAchievementFactory does not produce UserAchievements with unrealistic attributes.

ntagtmeier:
This PR first added a cleaning function to UserAchievement that first made sure that the relationship between date_earned and last_updated was well-defined and second made sure that the "progress" attribute was between 0 and the threshold of the achievement. Next, the factories were updated so that there is one factory that creates completed user achievements and one factory that creates uncompleted user achievements. The former is the one that is currently being used in tests. For that factory, last_updated was implemented as a lazy attribute dependent on date_earned.


mhenry111:
For the clean function, I wonder if we don't want to assert progress < threshold. This is maybe very hypothetical but say there is a basketball game, and there is an achievement like score 10 points, and a user has earned 8 points so far, and then they hit a three-pointer in a match to push them to 11 points overall, then their progress would be > threshold, but that would be fine and they would earn that achievement. Maybe if progress > threshold after userachievement progress has been updated then we can just reset it to be equal to threshold?

ntagtmeier:
> For the clean function, I wonder if we don't want to assert progress < threshold. This is maybe very hypothetical but say there is a basketball game, and there is an achievement like score 10 points, and a user has earned 8 points so far, and then they hit a three-pointer in a match to push them to 11 points overall, then their progress would be > threshold, but that would be fine and they would earn that achievement. Maybe if progress > threshold after userachievement progress has been updated then we can just reset it to be equal to threshold?

I agree with this and have altered the code accordingly. Note that this makes it important that we ensure in our methods that set progress that we do not allow advancement of UserAchievments that have already been awarded. Thanks for the catch!

mhenry111:
> 
>Note that this makes it important that we ensure in our methods that set progress that we do not allow advancement of UserAchievments that have already been awarded

Agreed. Probably once date_earned is set, we shouldn't allow any more updates

mhenry111 (approved):
great updates! lgtm

abargher (changes requested):
Looks like the linter and tests are failing due to a missing definition for UserAchievementFactory.

ntagtmeier:
I had to replace all of the old tests' usage of UserAchievementFactory with either CompletedUserAchievementFactory or UncompletedUserAchievementFactory, depending on the specific goals of the test. I also had to stop using auto_now_add for last_updated because it was producing improper updates in some of the tests (i.e. setting last_updated even though the achievement had already been awarded and wasn't being un-awarded).

abrodriguezalman (approved):
Reviewing on behalf of the API team: The changes made to the tests under api/tests/test_user_achievements.py correctly address the changes to the UserAchievement model. The tests still work as intended and still address the correct cases. LGTM!

ntagtmeier:
It looks like I screwed up and made the default date for last_updated static. Please hold off on review until I can fix it.

ntagtmeier:
OK, that issue is solved, and the logic should be more clear now.

abargher (approved):
Looks great, thanks for addressing these changes!

abargher:
Issue Score: Excellent

Comments:
Excellent work! Really great discussion and documentation of changes.

PR #1372, opened 2025-05-22
[knowledge-base] Added highlighting to Guide entries with unseen feedback on the Manage My Guides page

### LINKED ISSUE #1303: Add highlighting for guide items with unseen changes to the Contributor Dashboard guides table
As discussed in PR #1183, only a single notification displays on a user's landing page to inform them of unseen status changes on a guide or guides they uploaded. This issue was created to add some kind of highlight or indicator to guides on a user's Contributor Dashboard that have changes the user hasn't seen yet. So, when the notification from #1183 redirects them to their Contributor Dashboard, they will immediately be able to tell which guides have new updates.

mrollinswaterman:
## Overview
This PR fixes Issue #1303. Guides with unseen feedback are now highlighted when the user visits their 'Manage My Guides' page. The table's default sorting now displays guides with unseen feedback first. To partially address Yi's comments on PR #1255, the text on status buttons has been bolded for better visibility. 

## File Changes
**`src/chigame\knowledge-base/views.py`**
- Added code to fetch unseen guides as context when using the ContributorManageGuide view. 
- Added sorting code to fetch unseen guides and list them first when displaying the table with no sorting criteria (ie. default behavior)

**`src/templates/knowledge-base/contributor_manage_guide.html`**
- Added code to check the unseen_feedback context and adjust a table entry's styling accordingly.
- This highlights any guides who's PK appears in the list of guides with unseen feedback.
- Added code to bold the text on status buttons.

## Testing
2. Log in as any user with Guides uploaded, and unseen feedback.
- Should be any user with a fresh load of the fixture.
3. Go to your Manage My Guides Page.
4. Your page look like the screenshot below
5. Click on the Status button for a highlighted guide to view the feedback.
6. Return to your Contributor Dashboard to ensure that the guide who's feedback you viewed is no longer highlighted.

Screenshot 2025-05-22 at 1 26 36 PM


Luca-W-D:
Looks cool, thanks Miles!

Pulled the branch, runs as expected. In case anyone else in the future is following along, `sam@example.com / test12345` is a good test user. Alex didn't have any feedback pending :(

The highlighting does go away if I navigate via the "manage my guides" button after viewing feedback, but doesn't disappear if – after viewing the new feedback – I hit the back button. Is this intentional?

Also, as a small stylistic note, I think it could be good to make the text color white for green / red buttons. They're a little low-contrast and hard to read at the moment. That's just personal preference, though!

mrollinswaterman:
Thanks for taking a look Luca!

To the back button thing, I think that's because the context object for the contributor page doesn't get reloaded, it just uses the old one, so there's no room to register that you've now seen the feedback fro that guide. Not sure if that's something we can fix, but I don't know enough about web browsers to say.

In terms of the text color, I had the same feedback from Yi, which is why I bolded the button text in this PR, but I've tried white text and it looks weird in my opinion. Since the consensus is that it would be better, I'll see if I have time to slip it into this PR.

Luca-W-D:
Re:text color, maybe something like `text-light` which is slightly off-white?

Also that's a good point that the previously loaded context may be a technical limitation of Django. Maybe future iterations of the project could integrate an endpoint that the frontend checks using JS instead. Regardless, it's definitely out of scope for this PR.

Great work!

Luca-W-D (approved):

mrollinswaterman:
Fixed merge conflicts and implemented Luca's requested text color change.

jacqsmei (approved):
great work! nice discussion in the review

jacqsmei:
Issue Score: Excellent

Comments:
Well done PR & great teamwork in review!

PR #1380, opened 2025-05-22
[achievements] Achievement helper functions and tests

### LINKED ISSUE #680: write helper functions for models
This task involves adding helper functions and additional logic to the models.py, which currently just includes entities and attributes.

ntagtmeier:
Added an Achievement.set_progress method that sets a user's progress on an achievement to a given value. Altered the advance method to make use of that functionality. Added an is_earned method to check if a user has an achievement. Note that I made a small change to the UserAchievement attributes because it is useful to be able to check that last_updated and date_earned are the same when date_earned is not null, which is not possible if last_updated is set to auto_add.

I also added a test to make sure that set_progress is working correctly. Testing is_earned directly would require more up-to-date factories, such as those in PR #1364.


mhenry111:
the helpers look great! just to make sure I am understanding, for the advance function, does it increment progress by 1 if no amount is specified?

ntagtmeier:
> the helpers look great! just to make sure I am understanding, for the advance function, does it increment progress by 1 if no amount is specified?

Yes, exactly. The logic looks a little more complicated than it really is because the default progress is 1, but the basic idea is that in most use cases, developers of games will just want to increment progress by 1 and let the achievements system handle granting the achievement if necessary.

mhenry111:
> 
> Yes, exactly. The logic looks a little more complicated than it really is because the default progress is 1, but the basic idea is that in most use cases, developers of games will just want to increment progress by 1 and let the achievements system handle granting the achievement if necessary.

Makes sense, especially for the non-progress-based achievements.

mhenry111 (approved):
helpers look good to me, and the new tests look solid as well. great work!

abargher (changes requested):
Some merge conflicts in tests, please take a look

ntagtmeier:
Fixed issues with merging. For some reason, it seems like the tests.py file was deleted and replaced with test_views.py, but because my tests are not front-end, it seemed like keeping them in the old tests.py file made the most sense.

abargher (approved):
Looks great!

abargher:
Issue Score: Excellent

Comments:
Great work, very well documented!

PR #1398, opened 2025-05-23
[User Profile] Improved Style for Friend Request Accept Deny on Profile

### LINKED ISSUE #1390: [users] Prettify Friend Request Accept Deny Button
Currently the friend request accept or deny buttons on the user profile is too clunky (very wide and takes up too much space). I think a clean improvement to the UI can be to place a box within the profile with 'Pending Friend Request' as title and 'Accept' and 'Deny' buttons inside. This way we have smaller buttons that are less distracting. Especially since this would be involved in the demo as part of the interaction in adding a friend, making this small style change I think can be a big plus.

davdma:
Closes: #1390

Original accept and decline friend request buttons were placed next to stats below the user profile photo, but the buttons were slightly too wide and distracting. To improve the UI, I made some style alterations to viewing another users profile, providing a separate card with accept and decline buttons. Same goes for sending a request, the request will have a separate card with cancel button on the user who you have sent the request to.

Added:
- Changes to the userprofile_detail.html style and friend request displays

I also realized for best practice not to override `btn` with my own styles, and instead rename it to `btn-maroon`. A `btn-base` can also be defined to standardize button font sizes, border radius etc.

image


harriet-fisher (approved):
Nice work on this update! The new layout for the friend request buttons looks so much cleaner. Pulling them out into separate cards makes the whole profile page feel less crowded and more intentional so visually, it’s a huge improvement.
Also really appreciated the way you handled the CSS with the renaming .btn to .btn-maroon It keeps things modular and avoids overriding global styles, which will definitely help with maintainability for future students working on this.
I like that you added structure around both sending and receiving requests. It makes the different states super clear.
Only small suggestion would be to maybe leave a quick comment or note somewhere about how to use the new button classes (btn-base, btn-maroon, etc.) so others can follow that pattern. But overall, solid update!

hollyfwei (approved):
LGTM - thank you for making these changes!

cmarshall282 (approved):
Thanks for the UI update!

cmarshall282:
Issue Score: Excellent

Comments:
Great job!

PR #1420, opened 2025-05-23
[frontend] Added sign up functionality

### LINKED ISSUE #730: Implement Functionality for Sign Up Page
The sign up page is currently part of the website, but does not have any functionality, so we need to implement that. Ideally there are more features that we would like the signup page to have, such as the email verification in the Django version, but right now it is completely nonfunctional so we are just hoping to be able to hook up an api for signing up users.

clairezzhang:
Added Signup functionality to the Signup page in our Vue frontend. This checkes if the passwords are the same and ensures that a user can only be created when unique credentials for the email are used. Currently, this redirects to the login page when successful. Navigate to the `frontend/vue` folder and run `npm run dev` to run. In another terminal window, run the Django backend using  `python3 manage.py runserver` from the root of the repository.

image
image

Closes: #730


pjacobsen1 (approved):
Looks great! Note for anyone else - password has some requirements (can't just be password).

jacqsmei (approved):
nice job! another feature for this could be to add a tooltip describing password requirements

jacqsmei:
Issue Score: Excellent

Comments:
Great work on getting the sign up flow working!

PR #1421, opened 2025-05-23
[achievements] Added a threshold achievement to demo game

### LINKED ISSUE #1406: Add support for threshold achievements in demo game
Currently, the demo game only awards an achievement for clicking a button once and nothing else for subsequent clicks. I'd like to add support for achievements with thresholds, e.g., clicking a button 5 times or clicking a button 10 times awards another achievement.

emilyzen:
This PR adds a new achievement to the demo game that is awarded when 5 button clicks are registered. The implementation of this achievement doesn't quite use the threshold feature in the way intended--clicking a button doesn't increase the progress attribute of the user achievement. Instead, a number of clicks are registered in the Javascript code and then the achievement is granted when the number of clicks reaches 5 without increasing the progress attribute after each click. Nevertheless, I think my current implementation is good enough for demo purposes.


bensim0305 (approved):
For demo purposes I think this is pretty good! We should add in the documentation that this can be a lasting issue, or a follow up. Could be something we note in the documentation regarding this functionality. Otherwise lgtm!

abargher (approved):
Looks good!

abargher:
Issue Score: Excellent

Comments:
Great work! Nice descriptive comments.

Issue #1423, opened 2025-05-23
Display error: display all games instead of user-specific games


jiayuew1:
Currently, the template displays all games for the requested user, including games the user hasn’t played.

The tasks are:

- Fix the processing logic to include only the games the user has actually played.
- Correct the calculation of total achievement progress: it should be the number of achievements the user has earned divided by the total number of achievements available in ***the games the user has played***, not across all games.

jiayuew1:
Addressed in PR #1426

abargher:
Issue Score: Excellent

Comments:
Great work! Very clear description.

PR #1426, opened 2025-05-23
[achievements] Display Error Fix - Display User-Specific Games, Not ALL Games

jiayuew1:
To address issue #1423 
There is a minor update to `views.py`:

- Fetch only the games associated with the requested user, rather than all games in the database.
- Adjust the logic for calculating total achievements based on the games the user has played, and update the display accordingly.

### How to test:

1. Load fixtures and log in as `player1` and `player2`.
2. `player1` should see two game panels, while `player2` should see only one.
3. The summary line `Overall: a achievements unlocked out of b` should reflect different values of `b` for each player, based on the total achievements available in the games they've played.


ntagtmeier (approved):
Works great. Out of curiosity, though, why do the changes include pin-related stuff?

jiayuew1:
> Works great. Out of curiosity, though, why do the changes include pin-related stuff?

Oh, I see, I have used the views from other branch not dev. The pinning function is still in PRs, I should delete it to avoid duplicate.

ntagtmeier (approved):
Player 2 still shows up as having only one game, which is the desired behavior, so it seems taking out the pin logic didn't hurt your functionality. Great job!

bensim0305 (approved):
looks solid! Sorry for the late response

abargher (approved):
Looks good!

abargher:
Issue Score: Excellent

Comments:
Great job! Thanks for including clear testing instructions.

PR #1441, opened 2025-05-23
Game History backend

### LINKED ISSUE #1261: Game history implementation
Implement game history back-end which includes storing all complete and incomplete matches and their results. Game history should not include matches that have not started but already got lobby prepared for a game.

pauline196:
PR implements GameHistory model, which is designed to track individual users’ participation in game matches. Each entry  links a user to a specific match and states whether the match has been completed and the result (win, loss, draw or incomplete).

To ensure that the game history is recorded automatically, the logic for creating and updating its entries is implemented inside the Lobby.save() method. This approach is preferred over signals after reading the slack channel discussion about signals used for the similar issue. Specifically, when a lobby’s match_status changes from "Lobbied" (status 1) to "Viewable" (status 2), the application creates a GameHistory entry for every player in that match. When the status later changes from "Viewable" to "Finished", the corresponding GameHistory entries are updated to say that the match is completed and to store the outcome of the match for each player.

To test this feature, create a Lobby, add members, and link it to a Match. Set match_status to 2 (Viewable) and save — this should create GameHistory entries for each player. Then assign outcomes to players, change match_status to 3 (Finished), and save again — the entries should update with results and mark the match as completed.

**Testing:** 
You can load test data with:
- python manage.py loaddata game-history-simulation-fixture.json

Then verify entries using admin interface


Abebes01 (approved):

MuyanXie (approved):
1441 should be merged before 1525 but 1525 is already merged. Thus, we will approve this PR now and grant points as judged by the commit this PR is still meaningful

MuyanXie:
Issue Score: excellent

Comments:
Great job implementing the backend and providing the order or merge. Thanks!

Issue #1128, opened 2025-05-19
Fix easy bot errors in Checkers


brianHarder:
After several PRs were merged on 5/18, the easy bot looks to have been disrupted. The button for it is no longer in the navbar, and even when adding it the bot code does not run. The objective of this issue is to re-add the bot button to the navbar and fix the bot's functionality so that it can continue running like normal.

brianHarder:
Confirmed with teammates that they get the same error. Created working branch: https://github.com/uchicago-cs/chigame/tree/adding-games/bot-fix

brianHarder:
Removed duplicate functions and fixed endTurn functionality. Created [PR](https://github.com/uchicago-cs/chigame/pull/1152).

jacobwolf-1:
Not sure if this has been dealt with yet but my PR #1447 fixed it so the easy and medium bots are functional now when testing offline. I readded a button to the nav bar with  OFF->EASY->MEDIUM functionality

brianHarder:
Bot button returned to navbar. Easy bot functionality restored, [PR](https://github.com/uchicago-cs/chigame/pull/1152) merged.

abargher:
Issue Score: Excellent

Comments:
Great work, nice status updates!

Issue #612, opened 2025-04-28
Prototype for bot that is better than random bot


jacobwolf-1:
Have the bot do the max available jump, move closer to the center and maybe predict user moves

jacobwolf-1:
Feature done sending PR later today

jacobwolf-1:
The medium bot has been implemented in PR #1447 The medium bot prioritizes captures then central moves and can be turned on with the OFF->EASY->MEDIUMM button in nav

abargher:
Issue Score: Excellent

Comments:
Great work!
Here are some ways this could be improved:
- Even though you can add labels to issues to categorize them, you should avoid titles that could plausibly apply to multiple teams.

PR #1447, opened 2025-05-23
[Team Checkers] Adding medium bot and fix bot bug

jacobwolf-1:
This code addresses #1128 and #612 fixing the bug that was preventing the bot from playing, and adding the medium bot which prioritizes first captures and then central moves. **You can run play against the bot by running vs codes open live server on the index.html file.** This PR is pretty big and has a bunch of additions/changes to the previous code. I added two helper functions (Lines 712, 716) to compute distance from center, and if a move is a capture. Then I tweaked the getLegalMoves function (and easyBot to match) so it returns a list of [move, bool, int] where bool is if it is a capture, and the int is the distance from the center of the board. I also made a uniform x, y value in getLegalMoves, instead of row, col, jump_row, jump_col, to fix a bug where when the bot would try and capture a piece, it would disappear because the bot functions would index into nonexistent values. I also changed the bot button to a OFF ->EASY -> MEDIUM cycle and moved the scheduling of the got from the get hint function so it is called properly.

https://github.com/user-attachments/assets/fc778b73-a606-4618-b815-bac1b73e7f28

### DIFF
--- a/src/sandbox/checkers-prototype/game.js
+++ b/src/sandbox/checkers-prototype/game.js
@@ -49,7 +49,7 @@ let drawOffered = false;
 let drawOfferedBy = null;
 
 //BOT SETTINGS
-let vsEasyBot = true;
+let botMode = 'none';
 
 // ----------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


jacobwolf-1:
Update: fixed linter issues

enzeeo (approved):
I've followed the instructions as noted in the PR and everything works seamlessly. The bot is making more strategic moves where eat would eat my pieces when it can and when able to choose between moves, the bot will choose the move that will prevent its piece from getting captured. Excellent work! The buttons also cycle through Off -> Easy -> Medium properly without issue. LGTM!

abargher (changes requested):
Take a look at the merge conflicts, especially in `game.js`. Otherwise, looks great overall!

abargher:
Issue Score: Excellent

Comments:
Excellent work!

PR #1257, opened 2025-05-20
[achievements] Add UI for recent achievements.

jiayuew1:
To address issue #1256, this PR implements a new UI component that displays the user's most recent achievements on their profile page.
Changes Made

- Added a "Recent Achievements" section to the user profile template
  - Positioned in the right corner of the profile page
  - Styled header with blue background matching the active tab style
  - Implemented scrollable content area with fixed header
  - Used relative timestamps ("2 months ago") instead of absolute dates
- Updated user achievements view to fetch recent achievements
  - Modified user_achievements view to include recent achievements in the context

Screenshot:
image

PR #1496, opened 2025-05-24
[achievements] tests for recent_achievements on the UI frontend side

bensim0305:
This PR ensures coverage of PR #1257 . This PR also caught a bug where the "recent_achievements" function was not being called even on the user's own profile, so made the small fix in the user_achievements view to include the recent achievements. This test covers that the user achievement attributes are populating correctly on the frontend.


jiayuew1:
The test worked out great! I just have a minor question regarding the recent achievements not being called.  
It seems that PR #1257 does include recent achievements in both the `if pk` block and the final `context` dictionary.  
Could you help confirm if there's anything I might be missing?

bensim0305:
It looked like the `if` block is for when the user is viewing another user's profile, which then calls the recent_achievements function. In the `else` statement, it doesn't call the `recent_achievements` for viewing one's own recent achievements in the first "if-else" block. I noticed this error when testing, and the "recent_achivements" was not popping up when viewing their own profile.

jiayuew1 (approved):
Thank you for the explanation !

Issue #1206, opened 2025-05-20
Remove the redirection feature from the main model (after the new template button option is made)


zaynacheema:
Initially, we made it so that the original game_detail template allowed for twin uploads in order to test our twine upload implementation. Now that a new page has been made for creating IF games, the goal of this issue is to first remove the upload feature from the original game template and to remove the redirection to the IF game template that occurs when a twine upload is made on the original game template.

Now that the new template button was successfully made, this issue can be implemented safely.

zaynacheema:
Created a new branch for this issue called interactive-fiction/remove-twine-redirection

zaynacheema:
<img width="787" alt="Image" src="https://github.com/user-attachments/assets/909de91e-60d4-4b0b-90d6-edad644b96e0" />

successfully removed the "Upload Twine file" field from the regular game_creation template

zaynacheema:
Also made steps to remove the redirection url and the twine display from the normal game page view in urls.py and views.py

zaynacheema:
This is confirmed by games I previously made with the twine upload having their "Play Twine game" button removed from the detail page:

<img width="856" alt="Image" src="https://github.com/user-attachments/assets/3ad6a05f-256f-4d84-8a61-6f47bb63330d" />

zaynacheema:
PR #1501 has been linked above for this issue and merging this branch

zaynacheema:
With the  goal of removing the redirection done and removing the traces of twine upload outside of the IFcreation form, and with the PR successfully closed and the branch merged, this issue can also be closed.

At the moment I cannot think of future integration after this issue since this involves removing/updating a feature.

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! Great continual updates to track progress

PR #1501, opened 2025-05-24
[Interactive Fiction] Remove twine upload in regular game page and twine page redirection

zaynacheema:
Initially, we made it so that the original game_detail template allowed for twin uploads in order to test our twine upload implementation. Now that a new page has been made for creating IF games, the goal of this issue is to first remove the upload feature from the original game template and to remove the redirection to the IF game template that occurs when a twine upload is made on the original game template. This is related to issue #1206 

Now that the new template button was successfully made, this issue can be implemented safely.

Steps for code included:
-removed the "Upload Twine file" field from the regular game_creation template
-Also made steps to remove the redirection url and the twine display from the normal game page view in urls.py and views.py
- tested by looking if past twine games had the "play twine game" feature (ones that were made with the normal game template)
and looked at the normal game template to see the upload twine file space is gone:
Image


alisk8 (approved):
I, correctly, do not see an upload Twine game button on regular game creation page 

image

There is no play twine game feature on games created using the regular game template
image

This is great work!

jacqsmei (approved):
nice work!

jacqsmei:
Issue Score: Excellent

Comments:
Good clean up to have! Nice work!

Issue #1313, opened 2025-05-21
Leave a review on an IF game (adds onto #898)


smagdalen:
The goal of this issue is to implement a way to leave a review on an Interactive Fiction game on the new IF game detail template.

Currently, the new page made in issue #898 does not have the option to actually write a review, only to submit which also needs to be checked. This issue is based off the template from #898

smagdalen:
Made new branch interactive-fiction/leave_review_IF_game

zaynacheema:
Because this issue is based off of #898 I will create a branch off of the branch they used (which was interactive-fiction/new-detail-view) called 

interactive-fiction/twine-review

zaynacheema:
<img width="581" alt="Image" src="https://github.com/user-attachments/assets/dd03718c-abb7-4dae-bb3e-2869ae6d1c9b" />

After updating the Interactive detail view and corresponding urls.py and the detail template, the review form is visible.

<img width="716" alt="Image" src="https://github.com/user-attachments/assets/3d1ab1db-894e-48da-b645-3f70bc186bdc" />
I also tested submitting a review, and it shows up after clicking the "See reviews" link at the bottom

zaynacheema:
PR is linked above #1512

zaynacheema:
With the goal of this issue to make the review functionality possible and the PR closed and merged, this issue can be closed.

Future integrations could include making a success message that the review has gone through, and to display an error message if someone tries to input a rating that is not in the range of 1-5.

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! Good updates on the issue & future considerations in the closing.

Issue #898, opened 2025-05-13
Create a different game detail/viewing page for IF games 


zaynacheema:
Create a different game_detail page that will connect to the new Create IF game page. This issue will be split into making the backend part - that the data from the upload goes to the viewing page, and then the frontend part to make sure that data is displayed.

Open questions/tasks
- Figure out how to connect the data from the upload to a viewing page

gkharrell01:
Created branch interactive-fiction/new-detail-view to work on this issue

gkharrell01:
Created new file for viewing: IF_game_detail.html. 
Modified game_list to redirect to if-game/pk when a IF game is clicked on

gkharrell01:
Template is now a full draft @ http://127.0.0.1:8000/games/if-game/<pk>/
This removed "match" buttons, players, and rules and replaced with "play game in new tab" Button.

<img width="1233" alt="Image" src="https://github.com/user-attachments/assets/bda890bb-2ba2-4a1b-96e3-6112246a4563" />

gkharrell01:
Updated so where if an IF game is chosen from the home, it then uses this template above ^^

gkharrell01:
Closing Comment:
Addressed with PR #1418 
This implements a new detail template that mimics the original game_detail template while only keeping inputs pertaining to IF games. The buttons for Creating general and IF games lead to the correct create templates respectively. When choosing an IF game from the list, is uses the correct detail template depending on the game type, using the presence of a twine file as the indicator.

Next steps:
- Making sure the forms work correctly
- Debugging a way to make the migrated IF game use this template correctly, as the one uploaded in the migration is throwing an error when looking to review

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! Good closing comment & next steps

PR #1512, opened 2025-05-24
[Interactive Fiction] Implement Review Form for IF Game Detail

zaynacheema:
The goal of this issue is to implement a way to leave a review on an Interactive Fiction game on the new IF game detail template.

Currently, the new page made in issue #898 does not have the option to actually write a review, only to submit which also needs to be checked. 

This PR is related to issue #1313 

Image

After updating the Interactive detail view and corresponding urls.py and the detail template, the review form is visible.

Image
I also tested submitting a review, and it shows up after clicking the "See reviews" link at the bottom


alisk8 (approved):
I see that, on our new page, I am allowed to leave a review now

image

I see that whole number ratings and ratings like 2.5 are allowed. However, any more digits after the 5 reject the form. No error, however, is displayed when the form fails. This could be implemented in a future issue. 

My reviews do show up when I click on the "see reviews" button.

image

Would also love to see a success message that your review has been registered in a future issue.

However, the functionality you described is on point.

jacqsmei (approved):
nice work!

jacqsmei:
Issue Score: Excellent

Comments:
Good job! And good discussion of future steps on the review

Issue #1462, opened 2025-05-23
Fix bug in IF Game Create Template that autofills "None"


alisk8:
An issue I've come across in testing is that the IF game create template upon opening it autofills some fields with the text "None" and the user has to delete the words in order to type. 

The goal is to remove this issue and make it so the boxes are blank, with no words and no spaces either so the user can type normally

zaynacheema:
I will make a branch called interactive-fiction/fix-creation-fields

zaynacheema:
Before the fix it looked like this:

<img width="608" alt="Image" src="https://github.com/user-attachments/assets/9f8cfa15-1d47-4313-af59-e2dab6c9b123" />

and now,

<img width="577" alt="Image" src="https://github.com/user-attachments/assets/d03c3357-f311-433f-a073-e197902e03fd" />

The fields now no longer have the placeholder string "None" nor do they have placeholder spaces like "      "

zaynacheema:
PR #1518 is linked

zaynacheema:
With the goal of this issue met to fix the autofilling error, and with the PR closed and merged successfully, this issue can be closed. 

At the moment, I cannot think of any future integrations for this problem.

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! Good continual updates

PR #1518, opened 2025-05-24
[Interactive Fiction] Fix IF Creation Template autofilling

zaynacheema:
An issue I've come across in testing is that the IF game create template upon opening it autofills some fields with the text "None" and the user has to delete the words in order to type.
The goal is to remove this issue and make it so the boxes are blank, with no words and no spaces either so the user can type normally

This is related to issue #1462 

Before the fix it looked like this:

Image

and now,

Image

The fields now no longer have the placeholder string "None" nor do they have placeholder spaces like "      "

This was fixed by changing every part in the creation template that had form.xxx.value to default to "" if there was nothing in the form.


WilsonMach30 (commented):
This looks great Zayna! I tested the form and it works as intended.

WilsonMach30 (approved):

jacqsmei (approved):
great job!

jacqsmei:
Issue Score: Excellent

Comments:
Nice work! This is really helpful for the user experience

PR #1526, opened 2025-05-24
[Users] Implemented User Blocking system

### LINKED ISSUE #1515: [Users] User Blocking System
Currently, users can send, accept, deny, and even restore friend invitations. However, we do not have a method of blocking unwanted users. This would allow users to have more robust privacy and help manage unwanted notifications.

harriet-fisher:
To address issue #1515:
- Users can now block unwanted users to prevent friend invitations and manage privacy
- Preserves existing functionality while adding blocking capability
- Provides privacy protection by hiding blocking status from blocked users (just says "User Not Found"

Model Changes
- Updated User model with blocked_users ManyToManyField for non-symmetrical blocking relationships
- Automatically handles friendship removal and invitation cleanup upon blocking

Views
- Updated send_friend_invitation view to check for blocking status before allowing invitations
- Added block_user view to handle blocking with automatic cleanup
- Updated user_profile_detail_view to include blocking in context
- Preserved original friend invitation functionality for non-blocked users
- Added privacy-focused error messaging (blocked users see "User not found")

Templates
- Updated userprofile_detail.html to display block option alongside friend actions
- Added conditional display logic for blocked users showing "User not available"
- Maintained existing UI patterns while integrating new blocking functionality

urls
- Added block/<int:pk>/ endpoint for blocking functionality
- Integrated blocking route


src/chigame/users/models.py
hollyfwei, line 1: This feature is great - super cleanly implemented.

hollyfwei (approved):
Looks great Harriet - thanks for the solid work!

davdma (approved):
I like this blocking feature - should we also expose user profiles on the search results page if the user blocked them? Just something to think about. But this currently works well - we can just prevent profile viewing.

harriet-fisher:
> I like this blocking feature - should we also expose user profiles on the search results page if the user blocked them? Just something to think about. But this currently works well - we can just prevent profile viewing.

I also implemented some privacy settings in parallel with this (allows users to hide their profile from search or not allow friend requests from non-mutuals), so we can maybe create a new issue for future dev teams where they can utilize the same privacy logic, but trigger it with blocking

Jon984569 (approved):

cmarshall282 (approved):
Looks great!

cmarshall282:
Issue Score: Satisfactory

Comments:

PR #1530, opened 2025-05-24
[Word Game] Implements streaks for solo and daily mode

### LINKED ISSUE #1005: [Word Game] Implements Streaks For Solo and Daily Mode
We will implement localStorage-based tracking for player streaks in Word Game, with support for both solo and daily modes. Streaks will be saved between sessions and dynamically displayed in the UI. To prevent users from artificially inflating their daily streaks, the system will enforce a rule that only allows the daily streak to increase up to three times per day, once for each word length (4, 5, and 6). Solving the same daily word more than once will not contribute additional streak progress. Additionally, streaks will reset on losses to maintain the integrity of the system. The updated streak display will provide immediate feedback to players about their current performance and encourage continued play. This update will also support further development of achievements and analytics tied to streak history.

namdabest253:
This PR closes #1005 
Made it so that localstorage keeps track of the user's streak and updates the streaks display. Made it so the streaks for solo and daily are separated. And for the daily streak, made sure the user can only increase it 3 times a day (one for each word length), solving the same daily multiple times should not reward the user with an increased daily streak.
Also added streak edits to the Dev Tool ,along with reveal target word to allow better testing for Word Game.

Screenshot 2025-05-23 231857


vtcrenshaw (approved):
Good work and glad you limited daily streak; that's a nice catch.

src/sandbox/words-game/dev-tools.js
```diff
@@ -236,6 +277,117 @@
...
+                alert(`Current Target Word: "${word}"`);
+            } else {
+                alert('Word not available. Game may not be initialized yet.');
+            }
+        });
+
+        // Update streak display in dev panel
+        function updateStreakDisplay() {
+            // Get streaks from localStorage
+            const streaksData = localStorage.getItem("gameStreaks");
```
cmarshall282: Why don't we use getStreaks here?

cmarshall282 (approved):
Looks great. Just had one request

src/sandbox/words-game/dev-tools.js
nhnguyen253: I changed it so it uses getStreaks(), thanks for pointing that out.

cmarshall282:
Issue Score: Excellent

Comments:

- Please make sure that you do a thorough code review and, in the future, make sure to look out for the kind of issues the senior developer pointed out.

PR #1544, opened 2025-05-24
[Team Checkers] Hard Bot

jacobwolf-1:
Added more sophisticated hard bot. In order for the hard bot to "be smart", we are going to  have it "think ahead". In order to do this effeciently, we need to iterate over all possible moves that can be made by either player, and the responses to those moves, to a certain depth (for us, 4). Then once it has done that it finds the best board for black. Then it assumes black will max the scoreBoard and red will min it. So basically we are looking for the move where if red and black play optimally, it is the best move for black. **Changes are made in game.js, code can be tested by running vs code live server and playing the bot.**


jacobwolf-1:
Added simMove function to apply potential move to 2D copy of current board state (from getBoardState)

Added scoreBoard function to quantify how good a given position is for black (higher = better for black, lower = worse), using who is ahead, who controls the board, and who has the most moves. +100/-100 for black/red piece, and prioritizes closeness to center. Also +5/-5 per legal black/red move

Added minimax function which returns a score to quantify a given move. Once we are at the bottom of our recursive tree and depth = 0 it returns the result of the scoreBoard function. Otherwise It generates the legal moves for black if maximizing = true, otherwise it generates them for red. We switch this bool value for each recursive call. Then, it takes the max of those scores if maximizing = true to compute aplha, or the min if it’s false to compute beta, and updates them accordingly. Alpha is max score for black, and beta is lowest score red can enforce black gets. When maximizing is true we get all black moves, recurse on red moves, take and keep the max between the score we have now, or the new score we have found. When maximizing is false, we get all the red moves, recurse on black moves and keep the min between the score we have now, or the new score we have found. Once black’s best score (alpha) ≥ Red’s lowest enforceable score (beta) no move can push alpha higher or beta lower, so we stop looking here. 

Added hardBot function to get the board and for all the black moves, use the minimax function to get a score, then pick the move with the highest score and play it

Also modified the but botton to account for the Hardbot

jacobwolf-1:
Games vs easy/medium/hard bot to demonstrate skill levels:

Easy bot uses no strategy so winning is trivial

Medium bot either captures if available, otherwise it moves as close to the center as possible. Knowing that it is pretty easy to counter

Hard bot is noticeably "smarter" and presents more of a challenge

https://github.com/user-attachments/assets/863eeaae-1bfc-4fa2-8724-9281cac20a7e



https://github.com/user-attachments/assets/6f0cdc52-f18c-4ad5-9cf6-9b6520e84057


https://github.com/user-attachments/assets/50dcd321-301d-4d0e-be8c-a1f7198c2dd4

brianHarder (changes requested):
I tried playing against the medium and hard bots, and they work well! They took pieces that I left exposed and definitely made more strategic moves than the easy random bot. The JavaScript functions make sense too.

Changes look good to me, just some merge conflicts to address.

jacobwolf-1:
Resolved merge conflicts

jacobwolf-1:
Never mind I broke something fixing it now

jacobwolf-1:
Fixed

jacobwolf-1:
There was a missing } on line ~908 that caused everything to fail, and an added brace at the end that was not doing anything, and the var names were off for tile_size and margin in some places, so I fixed those. I also redid the functionality for the settings menu hover bc it was not working in my branch.

vliao1223 (approved):
looks good to me

brianHarder (approved):
NOTE: did not review PR, just approving so my previous change request doesn't stop merge

abargher (approved):
Looks awesome!

abargher:
Issue Score: Excellent

Comments:
Excellent work! The process here is incredibly well documented, nice job!

PR #1560, opened 2025-05-26
[Adding Games - Checkers] Add animation to capturing pieces

### LINKED ISSUE #1561: [Checkers] Adding animation to capturing
Currently, while there is an animation for moving normally, capturing a piece will simply cause the piece to teleport to the new location.

Coco555-cyber:
This PR adds an animation to capturing pieces, causing the capturing piece to slide over to the new position while the captured piece disappears. There is a bug with chain captures where the capturing piece will slide directly to the new position, but that can be fixed in a later PR.


nneidlinger (approved):
LGTM!

abargher (approved):
Looks great!

PR #1562, opened 2025-05-26
[Word Game] Fixed Views URL

namdabest253:
There was a small issue with the views.py URL still being the localhost URL I used for testing.


Zhejiej (approved):
Good fix, this bug would've been bad if it isn't fixed.

cmarshall282 (approved):
Good catch

